PR #1, opened 2021-08-23
Add tokenizer test

drwpow:
You can run this test with:

```
go test -v ./internal
```

Screen Shot 2021-08-23 at 17 52 58

It should pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,33 @@
+name: CI
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+
+      - name: Build
+        run: GOOS=js GOARCH=wasm go build ./cmd/astro
+
+      - name: Test
+        run: GOOS=js GOARCH=wasm go test ./internal
+
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: latest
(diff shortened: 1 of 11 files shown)


internal/token_test.go
```diff
@@ -0,0 +1,63 @@
...
+func TestTokenizer(t *testing.T) {
+	tests := []struct {
+		name string
+		input string
+		want []TokenType
+	}{
+		{
+		  "doctype",
+			`<!DOCTYPE html>`,
+			[] TokenType { DoctypeToken },
```
drwpow, line 18: I was thinking to expand these tests, we could test longer HTML strings that yield slices of tokens. Hence the param here being a slice.

What do you think about the test setup here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+
+      - name: Build
+        run: GOOS=js GOARCH=wasm go build ./cmd/astro
```
drwpow, line 21: This tests that the build can still happen

.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: Test
+        run: GOOS=js GOARCH=wasm go test ./internal
```
drwpow, line 24: This runs the unit tests (currently only in `./internal`; can update as needed)

.github/workflows/ci.yml
natemoo-re: We're actually using [`tinygo`](https://tinygo.org/) to build but `GOARCH=wasm` might be easier for CI? Not sure if there's a setup action for `tinygo`

natemoo-re (approved):
Looks awesome, thanks for adding!

.github/workflows/ci.yml
natemoo-re: Specifically `tinygo build -target=wasm -no-debug -o=astro.wasm ./cmd/astro `

natemoo-re:
@drwpow Got all the checks passing! Going to merge. 😄

PR #6, opened 2021-08-27
Framework react test

drwpow:
OK I lied; I actually did poke at it a little more when I said I wasn’t gonna. So adding the example from our `examples/framework-react` code, it actually does hang in the parser so I guess that makes more sense now!

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -43,7 +43,7 @@ func TestPrinter(t *testing.T) {
 		want   want
 	}{
 		{
-			name: "basic (no frontmatter)",
+			name:   "basic (no frontmatter)",
 			source: `<button>Click</button>`,
 			want: want{
 				imports:     "",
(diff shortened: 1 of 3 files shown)


internal/printer/printer_test.go
```diff
@@ -41,29 +41,29 @@ func TestPrinter(t *testing.T) {
 		want   want
 	}{
 		{
-			"basic (no frontmatter)",
-			`<button>Click</button>`,
-			want{
+			name: "basic (no frontmatter)",
+			source: `<button>Click</button>`,
+			want: want{
```
drwpow, line 46: I was just messing around adding these. Dunno if I like em or not

internal/printer/printer_test.go
```diff
@@ -200,6 +200,71 @@ const name = "world";
...
+        display: grid;
+        place-items: center;
+        margin-bottom: 2em;
+      }
+    </style>
+  </head>
+  <body>
+    <main>
+      <Counter {...someProps} client:visible>
+        <h1>Hello React!</h1>
```
drwpow, line 245: This slot, along with the spread props, may be what’s hanging it up? Anyway this test just never completes, which matches what I was seeing from the npm version. Let’s take a look at it next week!

PR #7, opened 2021-09-01
Fix empty <script> tags

drwpow:
Fixes a bug where we are trying to extract `<script>` tags with no body content.

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -256,7 +256,7 @@ const name = "world";
 
 <body>
   <!-- your content here... -->
-
+  <script src="js/scripts.js"></script>
   </body></html>`,
 			},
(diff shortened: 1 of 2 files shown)


drwpow:
> Is the compiler meant to close the open `<HTML>` tag?

Yes, actually! As a ~accidental side-effect~ feature of using a Go HTML parser, we will start to more closely follow the HTML spec which is extremely relaxed, and allows you to not close tags (of course, in certain instances it’ll guess wrong, so it’s good practice to close tags yourself, but technically speaking you don’t _have_ to close tags).

drwpow:
@natemoo-re could you review when you get a chance?

natemoo-re (approved):
LGTM!

PR #8, opened 2021-09-02
Add scoped CSS feature & tests

drwpow:
Porting this test into the Go compiler: https://github.com/snowpackjs/astro/blob/main/packages/astro/test/astro-scoped-styles.test.js.

This adds support for `:global()` as well as some trickier CSS queries that can easily be messed up.

✅ All tests added are passing (`scope-css_test.go`), but it should be noted that the `main` branch isn’t passing yet so CI will still fail overall.

Screen Shot 2021-09-02 at 19 29 01

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -264,7 +264,7 @@ const someProps = {
 
 // Full Astro Component Syntax:
 // https://docs.astro.build/core-concepts/astro-components/`,
-				styles: []string{":global(:root.astro-HMNNHVCQ ){font-family:system-ui;padding:2em 0;}:global(.counter.astro-HMNNHVCQ ){display:grid;grid-template-columns:repeat(3,minmax(0,1fr));place-items:center;font-size:2em;margin-top:2em;}:global(.children.astro-HMNNHVCQ ){display:grid;place-items:center;margin-bottom:2em;}"},
+				styles: []string{":root{font-family:system-ui;padding:2em 0;}.counter{display:grid;grid-template-columns:repeat(3,minmax(0,1fr));place-items:center;font-size:2em;margin-top:2em;}.children{display:grid;place-items:center;margin-bottom:2em;}"},
 				code: `<html lang="en">
   <head>
     <meta charset="utf-8">
--- a/internal/transform/scope-css.go
+++ b/internal/transform/scope-css.go
@@ -2,6 +2,7 @@ package transform
 
 import (
 	"bytes"
+	"strings"
 
 	// "strings"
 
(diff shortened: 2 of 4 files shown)


internal/transform/scope-css_test.go
```diff
@@ -0,0 +1,184 @@
...
+		},
+		{
+			name:   "import (plain)",
+			source: "@import \"./my-file.css\";",
+			want:   "@import \"./my-file.css\";",
+		},
+		{
+			name:   "import (url)",
+			source: "@import url(\"./my-file.css\");",
+			want:   "@import url(\"./my-file.css\");",
```
drwpow, line 164: All tests were copied over from Astro! 🎉 And then some new ones added (to ensure we aren’t transforming CSS unintentionally)

internal/printer/printer_test.go
```diff
@@ -264,7 +264,7 @@ const someProps = {
 
 // Full Astro Component Syntax:
 // https://docs.astro.build/core-concepts/astro-components/`,
-				styles: []string{":global(:root.astro-HMNNHVCQ ){font-family:system-ui;padding:2em 0;}:global(.counter.astro-HMNNHVCQ ){display:grid;grid-template-columns:repeat(3,minmax(0,1fr));place-items:center;font-size:2em;margin-top:2em;}:global(.children.astro-HMNNHVCQ ){display:grid;place-items:center;margin-bottom:2em;}"},
+				styles: []string{":root{font-family:system-ui;padding:2em 0;}.counter{display:grid;grid-template-columns:repeat(3,minmax(0,1fr));place-items:center;font-size:2em;margin-top:2em;}.children{display:grid;place-items:center;margin-bottom:2em;}"},
```
drwpow, line 267: WOOHOO! 🎉

internal/transform/scope-css.go
```diff
@@ -10,52 +11,200 @@ import (
...
+		n.FirstChild.Data = out
 	}
 }
 
-func scopeRule(rule string, opts TransformOptions) string {
-	scope := ".astro-" + opts.Scope
-	return rule + scope
+// Turn ".foo" into ".foo.astro-XXXXXX"
+func scopeRule(id string, opts TransformOptions) string {
+	return id + ".astro-" + opts.Scope
```
natemoo-re, line 186: Probably not the time for this discussion, but should we use `:where(.astro-xxx)` to avoid increasing the specificity?

internal/transform/scope-css.go
```diff
@@ -10,52 +11,200 @@ import (
...
+}
+
+// Get list of elements that should be scoped
+func shouldScope(id string) bool {
+	switch id {
+	case "base",
+		"body",
```
natemoo-re, line 193: `body` seems valid, no?

internal/transform/scope-css.go
```diff
@@ -10,52 +11,200 @@ import (
...
+		"font",
+		"frame",
+		"frameset",
+		"head",
+		"html",
```
natemoo-re, line 198: `head` and `html` also seem valid?

natemoo-re (commented):
This PR looks good to me! A few questions.

We probably have some more work to figure out our style pre-processing and sourcemap handling, but that's out of scope for this PR.

internal/transform/scope-css.go
drwpow: That would turn `body.theme-dark` into `body.theme-dark.astro-XXXXXX` which is what you don’t want, because `<body>` won’t be scoped within a component (there can only be one). It will still scope `body .localclass` properly.

internal/transform/scope-css.go
drwpow: Same for `<body>`

internal/transform/scope-css.go
drwpow: Maybe this doesn’t answer your question, but as of now, any CSS functions aren’t scoped. So `.my-class:not([data-foo])` would become `.my-class.astro-XXXXXX:not([data-foo])`. Not `:not([data-foo].astro-XXXXXX)`. Same with `:where()`—anything inside any parens will be left as-written.

internal/transform/scope-css_test.go
```diff
@@ -0,0 +1,199 @@
...
+	// note: this must be valid CSS, hence the empty "{}"
+	tests := []struct {
+		name   string
+		source string
+		want   string
+	}{
+		{
+			name:   "class",
+			source: ".class{}",
+			want:   ".class.astro-XXXXXX{}",
```
drwpow, line 23: @jonathantneal No pressure at all, but if you have a moment could you review some of these examples / suggest some of your own? The `source` is the written code, and the `want` is the expected output.

drwpow:
Gonna merge, but happy to take any feedback on this PR post-merge and reincorporate it in future work!

PR #9, opened 2021-09-03
Get tests passing

drwpow:
Updates tests to get CI passing again

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -25,17 +25,18 @@ const Component = %s(async ($$result, $$props, $$slots) => {`, CREATE_COMPONENT)
 var RETURN = fmt.Sprintf("return %s%s", TEMPLATE_TAG, BACKTICK)
 var SUFFIX = fmt.Sprintf("%s;", BACKTICK) + `
 });
-export default Component;
-`
+export default Component;`
+var STYLE_PRELUDE = "const STYLES = [\n"
+var STYLE_SUFFIX = "];\n$$result.styles.add(...STYLES)\n"
 
 func TestPrinter(t *testing.T) {
 	type want struct {
 		imports     string
 		frontmatter string
+		styles      []string
 		code        string
 		// TODO: add scripts & styles for testing later?
 		// scripts []string
-		// styles []string
 	}
 	tests := []struct {
 		name   string
@@ -48,6 +49,7 @@ func TestPrinter(t *testing.T) {
 			want: want{
 				imports:     "",
 				frontmatter: "",
+				styles:      []string{},
 				code:        `<html><head></head><body><button>Click</button></body></html>`,
 			},
 		},
@@ -60,6 +62,7 @@ const href = '/about';
 			want: want{
 				imports:     "",
 				frontmatter: "const href = '/about';",
+				styles:      []string{},
 				code:        `<html><head></head><body><a${` + ADD_ATTRIBUTE + `(href, "href")}>About</a></body></html>`,
 			},
 		},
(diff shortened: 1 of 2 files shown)


internal/printer/printer_test.go
```diff
@@ -198,15 +206,14 @@ const name = "world";
 
 <body>
   <!-- your content here... -->
-  <script src="js/scripts.js"></script>
```
drwpow, line 201: This was missing and I don’t know where it went, but I think it’s part of the script hoisting work and I’m not too worried about it. If this is a bug it won’t be one for long.

internal/printer/printer_test.go
```diff
@@ -250,21 +256,29 @@ const someProps = {
...
-	</html>`,
+				imports: "import Counter from '../components/Counter.jsx'\n",
+				frontmatter: `// Component Imports
+const someProps = {
+  count: 0,
+}
+
+// Full Astro Component Syntax:
+// https://docs.astro.build/core-concepts/astro-components/`,
+				styles: []string{":global(:root.astro-HMNNHVCQ ){font-family:system-ui;padding:2em 0;}:global(.counter.astro-HMNNHVCQ ){display:grid;grid-template-columns:repeat(3,minmax(0,1fr));place-items:center;font-size:2em;margin-top:2em;}:global(.children.astro-HMNNHVCQ ){display:grid;place-items:center;margin-bottom:2em;}"},
```
drwpow, line 267: Added `styles` in our tests

internal/printer/printer_test.go
```diff
@@ -283,20 +297,33 @@ const someProps = {
...
-
 %s
 // ---
 `, strings.TrimSpace(tt.want.frontmatter))
+				// format output
+				formatted := make([]string, 0)
+				for _, part := range strings.Split(output, "// ---") {
+					formatted = append(formatted, strings.TrimSpace(part))
+				}
+				output = strings.Join(formatted, "\n// ---\n")
```
drwpow, line 316: Since the diff library we’re using is very cranky, I put a little more work into normalizing the whitespace within the frontmatter.

Maybe we should do this in the compiler itself?

internal/token.go
```diff
@@ -259,9 +259,9 @@ type Tokenizer struct {
 	// rawTag's contents are lower-cased.
 	rawTag string
 	// stringStartChar is the character that opened the last string: ', ", or `
-	stringStartChar byte
+	// stringStartChar byte
 	// stringIsOpen will be true while in the context of a string
-	stringIsOpen bool
+	// stringIsOpen bool
```
drwpow, line 264: go-lint: unused

internal/token.go
```diff
@@ -798,7 +798,6 @@ func (z *Tokenizer) readString(c byte) {
 	case '`':
 		z.readTemplateLiteralString()
 	}
-	return
```
drwpow, line 801: go-lint: unnecessary `return`

internal/printer/printer_test.go
```diff
@@ -283,20 +297,33 @@ const someProps = {
 				Scope:       "astro-XXXX",
 				InternalURL: "http://localhost:3000/",
 			})
-			output := string(result.Output)
+			output := strings.TrimSpace(test_utils.Dedent(string(result.Output)))
```
drwpow, line 300: Also just ran Dedent on everything because I was tired of tabs vs spaces diffs. Hopefully we won’t see too much of that again

internal/printer/printer_test.go
natemoo-re: Yeah I'd be +1 on doing this in the compiler itself!

natemoo-re (approved):
Awesome, looks great!

PR #12, opened 2021-09-13
use proper polyfills for process and fs

okikio:


drwpow:
Thanks for opening. Can I get a little more detail as to the switch? E.g. how `memfs` is an improvement over what’s there?

okikio:
The current polyfill you are using doesn't have feature parity with normal fs, so, I suggested memfs since it's a better, polyfill. The main benefit is better compatability with node.

okikio:
Also, the tests failing wasn't because of my changes. When I cloned the project I found that the tests still failed

natemoo-re:
Thanks for this! These polyfills are copied from the go `wasm_exec.js` file and slightly adapted. Not sure if we want full polyfills here or to just stick with the "not implemented" version Go has built-in. Our compiler isn't accessing the filesystem or anything, so we're technically never going to hit these polyfills from Go/WASM.

okikio:
Oh,  OK

natemoo-re:
I'm going to close this as these polyfills are just boilerplate from the Go `wasm_exec.js` file.

But honestly, thank you _so much_ for opening a PR @okikio! Exciting to see contributions for our new compiler! 😁

natemoo-re:
I'm going to close this as these polyfills are just boilerplate from the Go `wasm_exec.js` file.

But honestly, thank you _so much_ for opening a PR @okikio! Exciting to see contributions for our new compiler! 😁

PR #13, opened 2021-09-14
Fix multiple expressions bug, add test

drwpow:


internal/token_test.go
```diff
@@ -270,6 +270,11 @@ func TestExpressions(t *testing.T) {
 			"{`${`${`${foo}`}`}`}",
 			[]TokenType{StartExpressionToken, TextToken, TextToken, TextToken, TextToken, TextToken, EndExpressionToken},
 		},
+		{
+			"element with multiple expressions",
+			"<div>Hello {first} {last}</div>",
```
drwpow, line 275: This was failing before. Basically, when entering expression context, we would enter “text” mode but fail to exit.

internal/token.go
```diff
@@ -1602,6 +1602,7 @@ expression_loop:
 				z.tt = TextToken
 				return z.tt
 			}
+			z.openBraceIsExpressionStart = true
```
drwpow, line 1605: Not 100% sure this is in the right place but all tests seem to be passing 🤷🏻. And we can move with other tests.

natemoo-re: Noice! LGTM.

natemoo-re (approved):
LGTM!

PR #18, opened 2021-09-16
Add "as" mode to switch between document/component parsing

natemoo-re:


drwpow (approved):

FredKSchott:
Missing a TS type for "as" when I try to build the `tycho-2` branch. 

Also, I know we'd really tried to avoid different parsing/compilation rules for document vs. component. What's the thing that finally forced us to add this? is there any way around it?

natemoo-re:
@FredKSchott It's not so much "page" versus "component" as it is "document" versus "document fragment" parsing. The HTML spec defines both, so exposing this would mean our components won't automatically get `<html><head /><body/> </html>` wrappers.

FredKSchott:
Makes sense, as long as we don't build too much other functionality into this flag then I'm +1. Thanks for the context.

PR #20, opened 2021-09-17
Do not insert `CommentNode` into slots

natemoo-re:
Previously `<Component><!-- comment --></Component>` would not render a fallback because `<!-- comment -->` was passed in as a slotted node.

We lose a little fidelity here, but I can't think of a good way to preserve comments when passing them to various `<slot>` elements.

### DIFF
--- a/internal/printer/print-to-js.go
+++ b/internal/printer/print-to-js.go
@@ -504,7 +504,9 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 							}
 						}
 					}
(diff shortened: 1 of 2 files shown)


drwpow (approved):

drwpow:
Is there a test or failing syntax that this was breaking? If we wanted to add support for this in the far (far) future, would want to test what this is fixing

natemoo-re:
@drwpow This was breaking slotting behavior because `<!-- comment -->` node should not be passed to slots. I'll add a test.

PR #21, opened 2021-09-21
Bugfix: conditional rendering

drwpow:
Fixes a bug with conditional rendering.

### DIFF
--- a/internal/printer/print-to-js.go
+++ b/internal/printer/print-to-js.go
@@ -303,12 +303,11 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 		}
 
 		for c := n.FirstChild; c != nil; c = c.NextSibling {
+			p.addSourceMapping(c.Loc[0])
 			if c.Type == TextNode {
-				p.addSourceMapping(c.Loc[0])
 				p.print(c.Data)
 			} else {
-				p.addSourceMapping(c.Loc[0])
-				if (c.PrevSibling == nil || (c.PrevSibling != nil && c.PrevSibling.Type == TextNode)) && c.NextSibling != nil {
+				if c.PrevSibling == nil || (c.PrevSibling != nil && c.PrevSibling.Type == TextNode) {
 					// TODO: where is this used?
 					// c.NextSibling.Type != TextNode
 					p.printTemplateLiteralOpen()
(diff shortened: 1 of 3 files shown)


internal/printer/print-to-js.go
```diff
@@ -303,12 +303,11 @@ func render1(p *printer, n *Node, opts RenderOptions) {
...
 
 		for c := n.FirstChild; c != nil; c = c.NextSibling {
+			p.addSourceMapping(c.Loc[0])
 			if c.Type == TextNode {
-				p.addSourceMapping(c.Loc[0])
 				p.print(c.Data)
 			} else {
-				p.addSourceMapping(c.Loc[0])
-				if (c.PrevSibling == nil || (c.PrevSibling != nil && c.PrevSibling.Type == TextNode)) && c.NextSibling != nil {
+				if c.PrevSibling == nil || (c.PrevSibling != nil && c.PrevSibling.Type == TextNode) {
```
drwpow, line 310: This was the expression I removed: `c.NextSibling != nil`. Not sure if that was a logical bug, or if I just introduced a regression. But the test added seemed to pass now.

internal/printer/printer_test.go
```diff
@@ -96,6 +96,16 @@ import VueComponent from '../components/Vue.vue';
...
 			},
 		},
+		{
+			name:   "ternary",
+			source: `<body>{false ? <div>#f</div> : <div>#t</div>}</body>`,
+			want: want{
+				imports:     "",
+				frontmatter: []string{},
+				styles:      []string{},
+				code:        "<html><head></head><body>${false ? $$render`<div>#f</div>` : $$render`<div>#t</div>`}</body></html>",
```
drwpow, line 106: This was what it was rendering before:

```
${false ? $$render`<div>#f</div>` : <div>#t</div>`}
```

Notice the `$$render` and backtick are missing from the right-hand side

internal/printer/print-to-js.go
natemoo-re: I feel like I ran into a condition where that was needed, but I can't remember what it was.

natemoo-re (approved):
Good to merge!

Would love to follow-up with more expression tests... Maybe we can catch some other bugs lurking in this area. (I'm sure there are some)

drwpow:
> Maybe we can catch some other bugs lurking in this area. (I'm sure there are some)

Oh don’t worry! There are plenty more where this came from 😛 (just uncomment some tests in the `next` branch)

PR #22, opened 2021-09-21
Fix dynamic classes

natemoo-re:
This fixes https://github.com/snowpackjs/astro-repl/issues/10 and a few other edge cases. Handles as many scoped class attribute situations as possible.

Unhandled, and unclear _how_ to handle: `{...value}` that contains `class`.

### DIFF
--- a/internal/transform/scope-html.go
+++ b/internal/transform/scope-html.go
@@ -1,6 +1,8 @@
 package transform
 
 import (
+	"strings"
+
 	tycho "github.com/snowpackjs/astro/internal"
 )
 
@@ -33,10 +35,32 @@ var NeverScopedElements map[string]bool = map[string]bool{
 func injectScopedClass(n *tycho.Node, opts TransformOptions) {
 	for i, attr := range n.Attr {
 		// If we find an existing class attribute, append the scoped class
-		if attr.Key == "class" {
-			attr.Val = attr.Val + " astro-" + opts.Scope
-			n.Attr[i] = attr
-			return
+		if attr.Key == "class" || (n.Component && attr.Key == "className") {
+			switch attr.Type {
+			case tycho.ShorthandAttribute:
+				if n.Component {
+					attr.Val = attr.Key + ` + " astro-` + opts.Scope + `"`
+					attr.Type = tycho.ExpressionAttribute
+					n.Attr[i] = attr
+					return
+				}
+			case tycho.EmptyAttribute:
+				// instead of an empty string
+				attr.Type = tycho.QuotedAttribute
+				attr.Val = "astro-" + opts.Scope
+				n.Attr[i] = attr
+				return
+			case tycho.QuotedAttribute, tycho.TemplateLiteralAttribute:
+				// as a plain string
+				attr.Val = strings.TrimSpace(attr.Val) + " astro-" + opts.Scope
+				n.Attr[i] = attr
+				return
+			case tycho.ExpressionAttribute:
+				// as an expression
+				attr.Val = attr.Val + ` + " astro-` + opts.Scope + `"`
+				n.Attr[i] = attr
+				return
+			}
 		}
 	}
 	// If we didn't find an existing class attribute, let's add one
--- a/internal/transform/scope-html.go
+++ b/internal/transform/scope-html.go
@@ -1,8 +1,6 @@
 package transform
 
 import (
-	"strings"
-
 	tycho "github.com/snowpackjs/astro/internal"
 )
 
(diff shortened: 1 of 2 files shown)


internal/transform/scope-html.go
```diff
@@ -33,10 +35,32 @@ var NeverScopedElements map[string]bool = map[string]bool{
 func injectScopedClass(n *tycho.Node, opts TransformOptions) {
 	for i, attr := range n.Attr {
 		// If we find an existing class attribute, append the scoped class
-		if attr.Key == "class" {
-			attr.Val = attr.Val + " astro-" + opts.Scope
-			n.Attr[i] = attr
-			return
+		if attr.Key == "class" || (n.Component && attr.Key == "className") {
```
natemoo-re, line 38: Also handling `className` if it's a Component for React compat

internal/transform/scope-html.go
```diff
@@ -33,10 +35,32 @@ var NeverScopedElements map[string]bool = map[string]bool{
...
+			switch attr.Type {
+			case tycho.ShorthandAttribute:
+				if n.Component {
+					attr.Val = attr.Key + ` + " astro-` + opts.Scope + `"`
+					attr.Type = tycho.ExpressionAttribute
```
natemoo-re, line 43: Switch `<Component {className} />` to `<Component className={className + " astro-XXXXX"} />`

internal/transform/scope-html.go
```diff
@@ -33,10 +35,32 @@ var NeverScopedElements map[string]bool = map[string]bool{
...
+				}
+			case tycho.EmptyAttribute:
+				// instead of an empty string
+				attr.Type = tycho.QuotedAttribute
+				attr.Val = "astro-" + opts.Scope
+				n.Attr[i] = attr
+				return
+			case tycho.QuotedAttribute, tycho.TemplateLiteralAttribute:
+				// as a plain string
+				attr.Val = strings.TrimSpace(attr.Val) + " astro-" + opts.Scope
```
natemoo-re, line 55: Should we `trim` the value or leave it as is?

internal/transform/scope-html.go
```diff
@@ -33,10 +35,32 @@ var NeverScopedElements map[string]bool = map[string]bool{
...
+				n.Attr[i] = attr
+				return
+			case tycho.ExpressionAttribute:
+				// as an expression
+				attr.Val = attr.Val + ` + " astro-` + opts.Scope + `"`
```
natemoo-re, line 60: This is the main fix!

internal/transform/scope-html_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"fmt"
+	"strings"
+	"testing"
+
+	astro "github.com/snowpackjs/astro/internal"
+	tycho "github.com/snowpackjs/astro/internal"
+	"golang.org/x/net/html/atom"
+)
+
+func printToSource(buf bytes.Buffer, node *tycho.Node) string {
```
natemoo-re, line 14: Added a specific printer for this test, can extract to a shared util if it continues to be useful.

internal/transform/scope-html_test.go
```diff
@@ -0,0 +1,123 @@
...
+		},
+		{
+			name:   "component className shorthand",
+			source: "<Component {className} />",
+			want:   `<Component className={className + " astro-XXXXXX"}></Component>`,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			nodes, err := tycho.ParseFragment(strings.NewReader(tt.source), &tycho.Node{Type: astro.ElementNode, DataAtom: atom.Body, Data: atom.Body.String()})
```
natemoo-re, line 112: `ParseFragment` with the context of a `atom.Body` node will skip the `html/head/body` wrappers. Handy tip for other test suites.

internal/transform/scope-html.go
drwpow: As long as we’re not adding extra chars at the end (which we aren’t) I’d just leave it 🤷🏻

internal/transform/scope-html_test.go
drwpow: Nice! Yeah printer test probably needs to be broken out at some point

internal/transform/scope-html_test.go
```diff
@@ -0,0 +1,123 @@
...
+			}
+		}
+		buf.WriteString(fmt.Sprintf(`></%s>`, node.Data))
+	}
+	return buf.String()
+}
+
+func TestScopeHTML(t *testing.T) {
+	// note: the tests have hashes inlined because it’s easier to read
+	// note: this must be valid CSS, hence the empty "{}"
```
drwpow, line 53: This note is old right?

internal/transform/scope-html_test.go
```diff
@@ -0,0 +1,123 @@
...
+		},
+		{
+			name:   "component className expression",
+			source: `<Component className={"test"} />`,
+			want:   `<Component className={"test" + " astro-XXXXXX"}></Component>`,
+		},
+		{
+			name:   "component className shorthand",
+			source: "<Component {className} />",
+			want:   `<Component className={className + " astro-XXXXXX"}></Component>`,
```
drwpow, line 107: Very nice!

drwpow (approved):
LGTM! Nice tests!

PR #23, opened 2021-09-22
Add support for Global styles

natemoo-re:
When encountering a `<style global>` element, we'll handle it properly and won't scope any elements in the same file.

Closes https://github.com/snowpackjs/astro/issues/1295

### DIFF
--- a/internal/print-to-source.go
+++ b/internal/print-to-source.go
@@ -1,0 +1,52 @@
+package astro
+
+import (
+	"fmt"
+	"strings"
+)
+
+func PrintToSource(buf *strings.Builder, node *Node) {
+	switch node.Type {
+	case DocumentNode:
+		for c := node.FirstChild; c != nil; c = c.NextSibling {
+			PrintToSource(buf, c)
+		}
+	case TextNode:
+		buf.WriteString(node.Data)
+	case ElementNode:
+		buf.WriteString(fmt.Sprintf(`<%s`, node.Data))
+		for _, attr := range node.Attr {
+			if attr.Namespace != "" {
+				buf.WriteString(attr.Namespace)
+				buf.WriteString(":")
+			}
+			buf.WriteString(" ")
+			switch attr.Type {
+			case QuotedAttribute:
+				buf.WriteString(attr.Key)
+				buf.WriteString("=")
+				buf.WriteString(`"` + attr.Val + `"`)
+			case EmptyAttribute:
+				buf.WriteString(attr.Key)
+			case ExpressionAttribute:
+				buf.WriteString(attr.Key)
+				buf.WriteString("=")
+				buf.WriteString(`{` + strings.TrimSpace(attr.Val) + `}`)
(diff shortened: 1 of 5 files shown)


internal/transform/scope-css.go
```diff
@@ -6,18 +6,29 @@ import (
...
 			continue
 		}
-		n.Attr = append(n.Attr, tycho.Attribute{
+		for _, attr := range n.Attr {
+			if attr.Key == "global" &&
+				(attr.Type == astro.EmptyAttribute) ||
+				(attr.Type == astro.ExpressionAttribute && attr.Val == "true") ||
+				(attr.Type == astro.QuotedAttribute && (attr.Val == "" || attr.Val == "true")) {
+				continue outer
+			}
```
natemoo-re, line 28: Skips `<style global>` nodes from scoping, returns a flag that we shouldn't scope HTML elements

internal/transform/transform.go
```diff
@@ -16,11 +16,11 @@ func Transform(doc *tycho.Node, opts TransformOptions) {
...
-		ScopeStyle(doc.Styles, opts)
-
-		walk(doc, func(n *tycho.Node) {
-			ScopeElement(n, opts)
-		})
+		if shouldScope := ScopeStyle(doc.Styles, opts); shouldScope {
+			walk(doc, func(n *tycho.Node) {
+				ScopeElement(n, opts)
+			})
+		}
```
natemoo-re, line 23: Conditionally scope HTML elements, rather than always

internal/print-to-source.go
```diff
@@ -0,0 +1,52 @@
...
+				buf.WriteString("=`" + strings.TrimSpace(attr.Val) + "`")
+			}
+		}
+		buf.WriteString(`>`)
+		for c := node.FirstChild; c != nil; c = c.NextSibling {
+			PrintToSource(buf, c)
+		}
+		buf.WriteString(fmt.Sprintf(`</%s>`, node.Data))
+	}
+}
```
natemoo-re, line 52: Extracted this to a shared `internal/` file. Only used for tests, but I was getting an import cycle when I tried to put it in the `test_utils` package.

It's nice how straight-forward this is!

FredKSchott (approved):
This one could probably use an extra look from @drwpow but overall lgtm

internal/transform/scope-css.go
```diff
@@ -6,18 +6,29 @@ import (
...
+outer:
 	for _, n := range styles {
 		if n.DataAtom != a.Style {
 			continue
 		}
-		n.Attr = append(n.Attr, tycho.Attribute{
+		for _, attr := range n.Attr {
+			if attr.Key == "global" &&
+				(attr.Type == astro.EmptyAttribute) ||
+				(attr.Type == astro.ExpressionAttribute && attr.Val == "true") ||
```
drwpow, line 25: Ooh nice

drwpow (approved):
LGTM!

PR #24, opened 2021-09-22
Fix text following expression

natemoo-re:
Previously, an `EndExpressionToken` in the `textIM` would revert the parser to its' original InsertionMode. Turns out, that's the wrong thing to do! If we're in `textIM`, anything after the expression should remain in the `textIM`. The InsertionMode will be reset automatically.

Closes https://github.com/snowpackjs/astro/issues/1362

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -1609,9 +1609,7 @@ func textIM(p *parser) bool {
 	case EndExpressionToken:
 		p.addLoc()
 		p.oe.pop()
-		p.im = p.originalIM
-		p.originalIM = nil
-		return p.tok.Type == EndExpressionToken
+		return true
 	}
 	p.im = p.originalIM
 	p.originalIM = nil
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -407,6 +407,16 @@ import Widget2 from '../components/Widget2.astro';`},
   </head><body></body></html>`,
 			},
 		},
+		{
+			name:   "text after title expression",
+			source: `<title>a {expr} b</title>`,
+			want: want{
+				imports:     "",
+				frontmatter: []string{},
+				styles:      []string{},
+				code:        `<html><head><title>a ${expr} b</title></head><body></body></html>`,
+			},
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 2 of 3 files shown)


internal/parser.go
```diff
@@ -1609,9 +1609,7 @@ func textIM(p *parser) bool {
 	case EndExpressionToken:
 		p.addLoc()
 		p.oe.pop()
-		p.im = p.originalIM
-		p.originalIM = nil
```
natemoo-re, line 1613: This was the culprit!

internal/parser.go
```diff
@@ -1609,9 +1609,7 @@ func textIM(p *parser) bool {
...
-		return p.tok.Type == EndExpressionToken
+		return true
 	}
 	p.im = p.originalIM
 	p.originalIM = nil
```
natemoo-re, line 1615: The insertion mode is reset automatically anyway, it doesn't need to be handled in the `EndExpressionToken` case

internal/token_test.go
```diff
@@ -280,6 +280,11 @@ func TestExpressions(t *testing.T) {
 			"{false ? <div>#f</div> : <div>#t</div>}",
 			[]TokenType{StartExpressionToken, TextToken, StartTagToken, TextToken, EndTagToken, TextToken, StartTagToken, TextToken, EndTagToken, EndExpressionToken},
 		},
+		{
+			"title",
+			"<title>test {expr} test</title>",
+			[]TokenType{StartTagToken, TextToken, StartExpressionToken, TextToken, EndExpressionToken, TextToken, EndTagToken},
+		},
```
natemoo-re, line 287: Sanity check: this is tokenizing correctly

internal/printer/printer_test.go
```diff
@@ -407,6 +407,16 @@ import Widget2 from '../components/Widget2.astro';`},
...
+		{
+			name:   "text after title expression",
+			source: `<title>a {expr} b</title>`,
+			want: want{
+				imports:     "",
+				frontmatter: []string{},
+				styles:      []string{},
+				code:        `<html><head><title>a ${expr} b</title></head><body></body></html>`,
+			},
+		},
```
natemoo-re, line 419: This is passing now!

FredKSchott: nice! would be great if we could add another test for `<title>a {expr} b {expr} c</title>`

FredKSchott (approved):

PR #25, opened 2021-09-22
Sort slot keys for stable output

natemoo-re:
Our `slot` tests were flaky because they relied on unstable output. This fixes the `slot` output so that it is always alphabetical.

This shouldn't have an effect on userland code because slots are explicitly placed by the user. They never access this object directly. This just fixes our test suite.


natemoo-re:
Merging to fix the test suite!

PR #27, opened 2021-10-05
Allow slot names to be expressions

matthewp:
Adds support for:

```astro
<div slot={name}>contents</div>
```


natemoo-re (approved):
Looks awesome—congrats on your first Compiler PR! 🎉 

Also Svelte doesn't even support this, so this is super cool!

PR #31, opened 2021-10-06
Allow top-level conditional expressions

matthewp:
This fixes https://github.com/snowpackjs/astro/issues/1493

You can do stuff like:

```astro
{cond && <span></span>}
```

At the top-level now.

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -732,11 +732,10 @@ func beforeHeadIM(p *parser) bool {
 		// Ignore the token.
 		return true
 	case StartExpressionToken:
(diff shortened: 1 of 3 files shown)


internal/parser.go
```diff
@@ -2358,6 +2357,28 @@ func afterAfterFramesetIM(p *parser) bool {
 	return true
 }
 
+func expressionIM(p *parser) bool {
```
natemoo-re, line 2360: Nice! This is much cleaner.

natemoo-re (approved):
Good solve! LGTM

PR #33, opened 2021-10-06
Fix Release Script

natemoo-re:
- set access to "public" as [recommended](https://github.com/atlassian/changesets/blob/main/docs/config-file-options.md#access-restricted--public) by the docs
- run `git push --follow-tags` as [recommended](https://github.com/atlassian/changesets/blob/main/docs/command-line-options.md#git-tags) by the docs


matthewp:
I'm been trying to fix this in main since I assumed there would be several things, so let's hope this gets it!

PR #35, opened 2021-10-06
Handle multiple conditionals at top-level

matthewp:
This is a continuation of #31, when in the head continue to use the expressionIM.

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -868,7 +868,7 @@ func inHeadIM(p *parser) bool {
 	case StartExpressionToken:
 		p.addExpression()
 		p.setOriginalIM()
-		p.im = textIM
+		p.im = expressionIM
 		return true
 	case EndExpressionToken:
 		p.addLoc()
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -499,12 +499,13 @@ import Widget2 from '../components/Widget2.astro';`},
 		},
 		{
 			name:   "condition expressions at the top-level",
-			source: `{cond && <span></span>}`,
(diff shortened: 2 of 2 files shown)


internal/parser.go
```diff
@@ -868,7 +868,7 @@ func inHeadIM(p *parser) bool {
 	case StartExpressionToken:
 		p.addExpression()
 		p.setOriginalIM()
-		p.im = textIM
+		p.im = expressionIM
```
natemoo-re, line 871: This is definitely correct

internal/printer/printer_test.go
```diff
@@ -499,12 +499,13 @@ import Widget2 from '../components/Widget2.astro';`},
...
 			name:   "condition expressions at the top-level",
-			source: `{cond && <span></span>}`,
+			source: `{cond && <span></span>}{cond && <strong></strong>}`,
+			only:   true,
 			want: want{
 				imports:     "",
 				frontmatter: []string{},
 				styles:      []string{},
-				code:        "<html><head>${cond && $$render`<span></span>`}</head><body></body></html>",
+				code:        "<html><head>${cond && $$render`<span></span>`}${cond && $$render`<strong></strong>`}</head><body></body></html>",
```
natemoo-re, line 508: ...but I'm realizing that this maybe isn't. Should top-level expressions go inside of the `head` or the `body`? I feel like we can't really know at compile time.

natemoo-re (commented):
Probably should have caught this on the last PR, but I'm a bit torn on if this is the desired behavior.

natemoo-re:
If you want to merge this and revisit this discussion in the future, that's fine too!

PR #41, opened 2021-10-07
chore: fix release script

natemoo-re:
This should build the package now!

### DIFF
--- a/.changeset/pretty-suns-sip.md
+++ b/.changeset/pretty-suns-sip.md
@@ -1,2 +1,5 @@
 ---
+"@astrojs/compiler": patch
 ---
+
+Actually include _any_ files?


package.json
```diff
@@ -9,6 +9,7 @@
   "scripts": {
     "build": "make astro-wasm && lerna run build --scope @astrojs/compiler",
     "build:compiler": "lerna run build --scope @astrojs/compiler",
+    "prerelease": "yarn build",
```
natemoo-re, line 12: That's all we needed, I think!

.changeset/pretty-suns-sip.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
matthewp, line 2: i love the simplicity here but I think you forgot something :)

matthewp (approved):

PR #48, opened 2021-10-08
Pass slots into createAstro

matthewp:
`$$slots` is needed by createAstro to create the `Astro.slots` object.


natemoo-re (approved):
LGTM!

PR #50, opened 2021-10-12
Improve JS scanner

natemoo-re:
Previously our algorithm for scanning JS frontmatter was naively breaking functions apart when it encountered an `await` keyword or the `Astro` global inside of functions. This adds some lexical awareness to the scanner to avoid that situation.

I also added some really basic test cases to ensure that there are no regressions.

### DIFF
--- a/.changeset/early-planes-press.md
+++ b/.changeset/early-planes-press.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+Improve JS scanning capabilities


internal/js_scanner/js_scanner.go
```diff
@@ -167,6 +228,42 @@ func readLineComment() {
 	}
 }
 
+func readTemplateString() {
```
drwpow, line 231: Are these pulled from somewhere? Or did you write them?

internal/js_scanner/js_scanner_test.go
```diff
@@ -0,0 +1,97 @@
...
+		{
+			name:   "basic",
+			source: `const value = "test"`,
+			want:   -1,
+		},
+		{
+			name: "basic split",
+			source: `const a = "test"
+const b = await fetch();`,
+			want: 17,
```
drwpow, line 25: Nice tests! Love testing assignment and different appearances of `async`.

drwpow: nit: test file needs `gofmt`

internal/js_scanner/js_scanner.go
drwpow: Reason I ask is because I’d like to either link to documentation where these originated, or I think we should add more comments here (such as examples of what each function is / isn’t meant to handle)

PR #51, opened 2021-10-12
Add hydration map to compiled output

matthewp:
This adds a hydration map to the compiled output. This is a list of all imports and the specifier. This will be used to find the component's client path for hydration purposes.

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -378,7 +378,7 @@ const someProps = {
 // Full Astro Component Syntax:
 // https://docs.astro.build/core-concepts/astro-components/
 import * as $$module1 from '../components/Counter.jsx';
-const $$hydrationMap = $$createHydrationMap('index.astro', [{ module: $$module1, specifier: '../components/Counter.jsx' }], [Counter]);`},
+const $$hydrationMap = $$createHydrationMap(import.meta.url, [{ module: $$module1, specifier: '../components/Counter.jsx' }], [Counter]);`},
 				styles: []string{fmt.Sprintf(`{props:{"data-astro-id":"HMNNHVCQ"},children:%s:root{font-family:system-ui;padding:2em 0;}.counter{display:grid;grid-template-columns:repeat(3,minmax(0,1fr));place-items:center;font-size:2em;margin-top:2em;}.children{display:grid;place-items:center;margin-bottom:2em;}%s}`, BACKTICK, BACKTICK)},
 				code: `<html lang="en">
   <head>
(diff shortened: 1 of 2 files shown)


internal/js_scanner/js_scanner.go
```diff
@@ -190,3 +240,21 @@ func readCommentWhitespace(br bool) byte {
 	}
 	return c
 }
+
+func readString(start int, quoteChar byte) int {
+	var c byte
+
+MainLoop:
```
natemoo-re, line 247: I don't think this loop is necessary!

internal/js_scanner/js_scanner.go
```diff
@@ -190,3 +240,21 @@ func readCommentWhitespace(br bool) byte {
...
+	for ; pos < len(source)-1; pos++ {
+		c = source[pos]
+		switch true {
+		case c == '\\':
+			pos++
+			continue
+		case c == quoteChar:
+			break MainLoop
```
natemoo-re, line 255: You can just `return pos` here

internal/printer/printer_test.go
```diff
@@ -554,6 +557,7 @@ import 'test';
 			hash := tycho.HashFromSource(code)
 			transform.Transform(doc, transform.TransformOptions{Scope: hash}) // note: we want to test Transform in context here, but more advanced cases could be tested separately
 			result := PrintToJS(code, doc, transform.TransformOptions{
+				Filename:    "index.astro",
```
natemoo-re, line 560: Is filename now required? Could we avoid this by using `import.meta.url` instead?

internal/printer/print-to-js.go
```diff
@@ -479,6 +485,43 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 	}
 }
 
+func printComponentImports(p *printer, doc *Node, source []byte) {
```
natemoo-re, line 488: Generally like to put these `printer` utils in `printer/printer.go` as a method on the `*printer` itself!

natemoo-re (commented):
Looks good! A few small comments, but I'm on board with this general approach.

internal/printer/printer_test.go
matthewp: Probably ideally the compiler should have some option like `Pathname` which would be `/src/components/Counter.jsx`. If you provide that it's what we use, otherwise we use `import.meta.pathname` (Vite). Happy to chat about this more, but Filename is definitely a placeholder only for the moment.

internal/printer/print-to-js.go
matthewp: Done!

PR #63, opened 2021-10-14
Bugfix: handle RegExp in Astro syntax

drwpow:
Tests added! Was investigating our `Code` internal component and realized this broke the compiler:

```
  // Replace "shiki" css variable naming with "astro".
  html = html.replace(/style="(background-)?color: var\(--shiki-/g, 'style="$1color: var(--astro-code-');
```

### DIFF
--- a/internal/token.go
+++ b/internal/token.go
@@ -726,9 +726,9 @@ scriptDataDoubleEscapeEnd:
 	goto scriptDataDoubleEscaped
 }
 
-// readComment reads the next comment token starting with "<!--". The opening
+// readHTMLComment reads the next comment token starting with "<!--". The opening
 // "<!--" has already been consumed.
-func (z *Tokenizer) readComment() {
+func (z *Tokenizer) readHTMLComment() {
 	z.data.Start = z.raw.End
 	defer func() {
 		if z.data.End < z.data.Start {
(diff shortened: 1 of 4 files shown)


internal/token.go
```diff
@@ -791,112 +791,74 @@ func (z *Tokenizer) readUntilCloseAngle() {
...
+	// template literal
 	case '`':
-		z.readTemplateLiteralString()
+		// Note that we DO NOT have to handle `${}` here because our expression
+		// behavior already handles `{}`. Technically incorrect, but it works.
+		z.readUntilChars([]byte{'`'})
 	}
 }
 
-func (z *Tokenizer) readSingleQuoteString() {
```
drwpow, line 803: Did a bit of cleanup in the tokenizer. We had a bunch of different utilties that all basically performed the same action:

- readSingleQuoteString()
- readDoubleQuoteString()
- readTemplateLiteralString()
- readJSSingleLineComment()
- readJSMultiLineComment()
- readRegExp()

All they did was just keep scanning until a specific character was encountered. That part has just been centralized into:

```
z.readUntilChars([]byte{'\'', '\r', '\n'})
```

This will read until any of these characters are encountered, handles escaped characters, and handles errors a bit more gracefully than the previous methods (hopefully fewer panics).

drwpow:
cc @matthewp I’m going to merge this & publish a new compiler version, but feel free to review post-merge if you’d like!

TL;DR no tests changed, no behavior changed, only a bugfix. All tests pass.

PR #65, opened 2021-10-15
Rename hydrationMap to metadata and add hoisted scripts

matthewp:
## Changes

- This changes the `$$hydrationMap` object to be `$$metadata` instead.
- `$$metadata` is exported so that the build (and potentially other tools) can read information about a component such as which hoisted scripts it contains and which hydrated components it has.

## Testing

Tests all updated.

## Docs

N/A

### DIFF
--- a/.changeset/tasty-onions-arrive.md
+++ b/.changeset/tasty-onions-arrive.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": minor
+---
+
+Renames the hydrationMap to \$\$metadata and makes it be exported


.changeset/tasty-onions-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/compiler": minor
+---
+
+Renames the hydrationMap to \$\$metadata and makes it be exported
```
natemoo-re, line 5: Nit: don't need to escape the `$` character

PR #67, opened 2021-10-19
[wip] Fix <style> tag in frontmatter

drwpow:
## Changes

If you declare a dynamic `<style>` tag in frontmatter, we technically don’t outline that behavior but IMO we should just ignore it / not try and run style processing on it. But this PR isn’t to outline that behavior; it’s just to fix breaking the compiler.

## Testing

Test added but code changes still needed


drwpow:
Nevermind; this seems to have been fixed in `main`! Closing.

PR #68, opened 2021-10-19
Add JS linting

drwpow:
## Changes

I am once again adding linting to a project! 

Obligatory disclaimer: **I don’t care what the lint rules are!** (other than perhaps `no-shadow` because that one bites me a lot). I am open to changing them at any time.

But I **do care about noisy/arbitrary/useless diffs**. This is only a means to that end. And that always happens on a project without any coding style enforced.

## Testing

Adds a `lint-js` test to lint JS on each PR


.github/workflows/ci.yml
```diff
@@ -49,3 +49,19 @@ jobs:
...
+
+  build-js:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        run: yarn
+        run: yarn build:compiler
+
+  lint-js:
```
drwpow, line 61: New `yarn lint` CI action

PR #69, opened 2021-10-19
Add style preprocessing

natemoo-re:
## Changes

- This exposes a `preprocessStyle` function to the compiler to preprocess styles.
- ~~Currently, this function has to be sync because `async` in Go turns out to be pretty weird. I tried.~~

Here are some handy Go + WASM resources specifically around Promises (that may or may not be useful)
- [Go, WebAssembly, HTTP requests and Promises: A guide to interacting with JavaScript from Go/Wasm](https://withblue.ink/2020/10/03/go-webassembly-http-requests-and-promises.html)
- [How to wait a js async function from golang wasm?](https://stackoverflow.com/questions/68426700/how-to-wait-a-js-async-function-from-golang-wasm)
- [Go by Example: WaitGroups](https://gobyexample.com/waitgroups)

## Testing

We don't have automatic WASM tests yet, but we should think about adding them!

## Docs

N/A


internal/transform/preprocess.go
```diff
@@ -0,0 +1,31 @@
...
+	dataOrPromise := stylePreprocess.Invoke(n.FirstChild.Data, getQuotedAttr(n, "lang"))
+	str := dataOrPromise.String()
+
+	fmt.Println(str)
+	// if str != "" {
+	// 	res, _ := wasm_utils.Await(dataOrPromise)
+	// 	newValue = res[0]
+	// } else {
+	// 	newValue = dataOrPromise
+	// }
```
natemoo-re, line 28: This is commented out, but if the `stylePreprocess` function is `async` we need to block Go until it resolves. 
`wasm_utils.Await(dataOrPromise)` mostly works, but the top-level function needs to wait for it as well. I assume WaitGroups will come in handy here.

lib/compiler/node/index.ts
```diff
@@ -3,13 +3,26 @@ import { promises as fs } from 'fs';
 import Go from "./wasm_exec.js";
 import { fileURLToPath } from "url";
 
+const sharedFunctions = Object.create(null);
+sharedFunctions.stylePreprocess = (value: string) => value;
```
natemoo-re, line 7: Default `stylePreprocess` function is just an identity function

lib/compiler/node/index.ts
```diff
@@ -3,13 +3,26 @@ import { promises as fs } from 'fs';
...
+export const init = async (options: any) => {
+  for (const key of Object.keys(sharedFunctions)) {
+    if (typeof options[key] !== 'undefined') {
+      sharedFunctions[key] = options[key];
+    }
+  }
+
+  return ensureServiceIsRunning();
+}
+
```
natemoo-re, line 18: Adds an `init` option that can provide WASM with any callbacks we might need. Just `stylePreprocess` for now.

lib/compiler/node/index.ts
```diff
@@ -39,6 +52,10 @@ const instantiateWASM = async (
 };
 
 const startRunningService = async () => {
+  for (const key of Object.keys(sharedFunctions)) {
+    const globalKey = `__astro_${key}`;
+    (globalThis as any)[globalKey] = (sharedFunctions as any)[key];
+  }
```
natemoo-re, line 58: Don't love this but exposing functions on `globalThis` is extremely straightforward. I couldn't seem to pass the function directly because then you end up in `wasm_exec.js` memory allocation hell.

cmd/astro-wasm/astro-wasm.go
```diff
@@ -77,47 +87,97 @@ type TransformResult struct {
...
+			return nil
+		})
+		defer handler.Release()
 
-	return vert.ValueOf(TransformResult{
-		Code: string(result.Output),
-		Map:  "",
+		// Create and return the Promise object
+		promiseConstructor := js.Global().Get("Promise")
+		return promiseConstructor.New(handler)
```
natemoo-re, line 180: Small refactor to return a `Promise` from `transform`.

cmd/astro-wasm/astro-wasm.go
```diff
@@ -77,47 +87,97 @@ type TransformResult struct {
...
+						Type: astro.DocumentNode,
+					}
+					for i := 0; i < len(nodes); i++ {
+						n := nodes[i]
+						doc.AppendChild(n)
+					}
+				}
+
+				// Hoist styles and scripts to the top-level
+				transform.ExtractScriptsAndStyles(doc)
```
natemoo-re, line 133: Main difference here is that we're splitting up the `transform` step into preprocessing and postprocessing.

cmd/astro-wasm/astro-wasm.go
```diff
@@ -77,47 +87,97 @@ type TransformResult struct {
...
+					if transformOptions.PreprocessStyle.IsUndefined() != true {
+						for i, style := range doc.Styles {
+							wg.Add(1)
+							i := i
+							go preprocessStyle(i, style, transformOptions, wg.Done)
+						}
+					}
+				}
+				// Wait for all the style goroutines to finish
+				wg.Wait()
```
natemoo-re, line 148: This spawns workers that run the async `preprocessStyle` functions and waits until they are all complete to continue.

cmd/astro-wasm/astro-wasm.go
```diff
@@ -77,47 +87,97 @@ type TransformResult struct {
...
+
+				// Perform CSS and element scoping as needed
+				transform.Transform(doc, transformOptions)
```
natemoo-re, line 151: At this point, our CSS is (presumably) actually CSS and we can scope it!

internal/transform/utils.go
```diff
@@ -15,3 +17,37 @@ func hasTruthyAttr(n *astro.Node, key string) bool {
...
+	for _, attr := range n.Attr {
+		switch attr.Type {
+		case astro.QuotedAttribute:
+			attrs.Set(attr.Key, attr.Val)
+		case astro.EmptyAttribute:
+			attrs.Set(attr.Key, true)
+		}
+	}
+	return attrs
+}
```
natemoo-re, line 41: This function returns compile-time known attributes as a JS object, which can then be passed to the preprocessors.

internal/wasm_utils/wasm_utils.go
```diff
@@ -0,0 +1,31 @@
...
+
+	awaitable.Call("then", thenFunc).Call("catch", catchFunc)
+
+	select {
+	case result := <-then:
+		return result, nil
+	case err := <-catch:
+		return nil, err
+	}
+}
```
natemoo-re, line 31: This is some magic I don't fully understand, borrowed from [StackOverflow](https://stackoverflow.com/questions/68426700/how-to-wait-a-js-async-function-from-golang-wasm). But it works the same as `await` in JS!

lib/compiler/shared/types.ts
```diff
@@ -4,6 +4,7 @@ export interface TransformOptions {
   sourcefile?: string;
   sourcemap?: boolean | 'inline' | 'external' | 'both';
   as?: 'document'|'fragment';
+  preprocessStyle: (content: string, attrs: Record<string, string>) => Promise<string>
```
natemoo-re, line 7: That's it, that's the feature!

lib/compiler/test.js
```diff
@@ -1,18 +1,52 @@
...
+        // HOLY CRAP THIS ACTUALLY WORKS!
+        preprocessStyle: async (value, attrs) => {
+          let x = i++;
+          if (!attrs.lang) {
+            return null;
+          }
+          console.log(`Starting to preprocess style ${x} as ${attrs.lang}`);
+          await sleep(3000);
+          console.log(`Finished preprocessing ${x}`);
+          return value.replace('color', 'background');
```
natemoo-re, line 39: Should be very straightforward.

`preprocessStyle` is an async function that gets the contents of the `style` element and any compile-time known attrs (we only care about `lang` now, but maybe others in the future!)

Simulating some expensive work with `sleep` here.

internal/wasm_utils/wasm_utils.go
FredKSchott: lol I'm both scared and thrilled!

cmd/astro-wasm/astro-wasm.go
```diff
@@ -1,23 +1,30 @@
...
 	"golang.org/x/net/html/atom"
 )
 
+var done chan bool
+
 func main() {
-	js.Global().Set("__astro_transform", js.FuncOf(Transform))
-	<-make(chan bool)
+	done = make(chan bool, 1)
+	js.Global().Set("__astro_transform", Transform())
```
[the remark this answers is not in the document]
matthewp: Not a big deal for this PR, but you can pass functions to wasm through an imports object: https://developer.mozilla.org/en-US/docs/WebAssembly/Using_the_JavaScript_API#preparing_the_example

PR #71, opened 2021-10-20
[wip] Add failing test for functional components in FrontMatter

drwpow:
## Changes

Adds failing test for one problem with the `<Debug>` component. Need some help in identifying what the tokens should be.


internal/token_test.go
```diff
@@ -169,6 +169,17 @@ func TestFrontmatter(t *testing.T) {
...
+		{
+			"function components w/ conditional rendering",
+			`---
+			const getSummary = (key) => {
+				return <>
+					{key && <><span class={` + "`" + `${className} key` + "`" + `}>{key}</span></>}
+				</>;
+			};
+			---`,
+			[]TokenType{FrontmatterFenceToken, TextToken, TextToken, StartTagToken, TextToken, StartExpressionToken, TextToken, StartTagToken, StartTagToken, StartExpressionToken, TextToken, TextToken, EndExpressionToken, StartExpressionToken, TextToken, EndExpressionToken, EndTagToken, EndTagToken, EndExpressionToken, TextToken, EndTagToken, TextToken, TextToken, FrontmatterFenceToken},
```
drwpow, line 181: Is this right? Or should it mostly be text (i.e. not try and flag internal expressions like start tags & end tags?

drwpow:
> Is implementing this even possible without a JS parser? I wasn't involved in the previous implementation, so maybe I'm misunderstanding how it worked before. If this becomes a big thing, like we start needing to parse JS, then I think we should revisit how this feature works.

We are parsing JS already because we allow JSX-style tags in frontmatter. But it’s partial parsing, that only cares about the boundaries of HTML and treats everything else as text. Still, this is needed to get the `<Debug />` component working.

PR #76, opened 2021-10-20
Add WASM test

drwpow:
## Changes

Tests the version of the compiler built automatically by CI

## Testing

Tests should pass (🤞)

PR #77, opened 2021-10-20
fix: ignore style preprocessor undefined/null return

natemoo-re:
## Changes

- Does what it says in the title

## Testing

N/A but can add if #76 goes first!

## Docs

N/A


drwpow:
Comment added [here](https://github.com/snowpackjs/astro-compiler-next/pull/76/files#diff-72146427232a47c8f1e5f1d60c4dc5c520dc9f92a23d28c6041431f19a836d66R94)

PR #125, opened 2021-10-31
fixing parsing for CSS variables in style parsing

tony-sull:
## Changes

Updates the scoped CSS transform to support CSS variables.  Currently, the colon is lost during the transform and `--bg: red` is output as `--bg red`

## Testing

Added a test to make cover CSS variable parsing

## Docs

bug fix only


natemoo-re (approved):
🙌🏻 Amazing catch! Thanks for diving into this one!

PR #128, opened 2021-10-31
Add support for `client:only` directive

natemoo-re:
## Changes

- Adds support for `client:only` directive
- Requires updates in `astro/internal`'s `renderComponent` function to gracefully handle this case. PR in progress to `astro`. 
- [] TODO: remove matching import from `metadata`, presumably cannot be SSR'd due to top-level `window` access.
  - Should top-level Component import be removed as well? `esbuild` will treeshake it out automatically.
- [] TODO: determine client import path and client export name ahead of time, since `metadata` won't have that info.

## Testing

Tests added

## Docs

No docs needed

### DIFF
--- a/internal/printer/print-to-js.go
+++ b/internal/printer/print-to-js.go
@@ -302,6 +302,7 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 	}
 
 	isComponent := (n.Component || n.CustomElement) && n.Data != "Fragment"
+	isClientOnly := isComponent && transform.HasAttr(n, "client:only")
 	isSlot := n.DataAtom == atom.Slot
 
 	p.addSourceMapping(n.Loc[0])
@@ -319,6 +320,8 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 	} else if !isSlot {
 		if n.CustomElement {
 			p.print(fmt.Sprintf("'%s'", n.Data))
+		} else if isClientOnly {
+			p.print("null")
 		} else {
 			p.print(n.Data)
 		}
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -149,6 +149,37 @@ import * as $$module1 from '../components';`,
 			},
 		},
 		{
+			name: "client:only component",
+			source: `---
+import { Component } from '../components';
+---
+<html>
+  <head>
+    <title>Hello world</title>
+  </head>
+  <body>
+    <Component client:only />
+  </body>
+</html>`,
+			want: want{
+				imports: "",
+				frontmatter: []string{
+					`import { Component } from '../components';
+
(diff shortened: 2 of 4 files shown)


internal/printer/print-to-js.go
```diff
@@ -319,6 +320,8 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 	} else if !isSlot {
 		if n.CustomElement {
 			p.print(fmt.Sprintf("'%s'", n.Data))
+		} else if isClientOnly {
+			p.print("null")
```
drwpow, line 324: Makes sense to me

internal/transform/transform.go
```diff
@@ -66,7 +66,7 @@ func ExtractScript(doc *tycho.Node, n *tycho.Node) {
 						id = fmt.Sprintf("'%s'", id)
 					}
 
-					if strings.HasPrefix(attr.Key, "client:") {
+					if strings.HasPrefix(attr.Key, "client:") && attr.Key != "client:only" {
```
drwpow, line 69: 👍🏻

internal/printer/printer.go
```diff
@@ -263,12 +263,41 @@ func (p *printer) printComponentMetadata(doc *astro.Node, source []byte) {
...
+	loc, statement := js_scanner.NextImportStatement(source, 0)
 	for loc != -1 {
-		p.print(fmt.Sprintf("\nimport * as $$module%v from '%s';", modCount, specifier))
-		specs = append(specs, specifier)
-		loc, specifier = js_scanner.NextImportSpecifier(source, loc)
-		modCount++
+		isClientOnly := false
+		for _, n := range doc.Metadata.ClientOnlyComponents {
+			for _, imported := range statement.Imports {
+				if imported.LocalName == n.Data {
```
drwpow, line 271: Great job checking for renamed components

drwpow (approved):
This looks great! I like the added test here. This seems difficult to test for here, but re-enabling the `client:only` test in Astro should be enough assurance this is working after SSR

internal/printer/printer_test.go
```diff
@@ -145,6 +142,31 @@ import * as $$module1 from '../components';`,
...
+</html>`,
+			want: want{
+				frontmatter: []string{"import { Component } from '../components';"},
+				// Specifically do NOT render any metadata here, we need to skip this import
+				code: `<html>
+  <head>
+    <title>Hello world</title>
+  </head>
+  <body>
+    ${` + RENDER_COMPONENT + `($$result,'Component',null,{"client:only":true,"client:component-path":"../components","client:component-export":"Component"})}
```
drwpow, line 169: 🎉 Awesome! Love to see that `null`

matthewp:
lgtm

PR #129, opened 2021-11-02
fix: script/style sibling issue

natemoo-re:
## Changes

- Previously, scripts/styles that were siblings of each other appeared to only be processed up until the first script or style tag.  
- I assumed this was a tokenization or parser bug, but it was in our `transform` module!
- Previously, we would add a style to `doc.Styles` and immediately remove it from it's parent node. Doing this _during_ the recursive `walk` function corrupted the children, meaning following nodes would be left in place.
- The fix was to **not remove** hoisted `style` or `script` elements until **after** the `doc` was `walk`ed.

## Testing

Added a test

## Docs

Bug fix only

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -665,6 +665,7 @@ func beforeHTMLIM(p *parser) bool {
 			p.im = inBodyIM
 			if p.hasSelfClosingToken {
 				p.oe.pop()
+				p.acknowledgeSelfClosingTag()
 			}
 			return true
 		}
@@ -755,6 +756,7 @@ func inHeadIM(p *parser) bool {
 			if p.hasSelfClosingToken {
 				p.addLoc()
 				p.oe.pop()
+				p.acknowledgeSelfClosingTag()
 			}
 			return true
 		}
@@ -783,6 +785,7 @@ func inHeadIM(p *parser) bool {
 			if p.hasSelfClosingToken {
 				p.addLoc()
 				p.oe.pop()
+				p.acknowledgeSelfClosingTag()
 			}
 			return true
 		case a.Noframes, a.Style:
@@ -790,6 +793,7 @@ func inHeadIM(p *parser) bool {
 			if p.hasSelfClosingToken {
 				p.addLoc()
 				p.oe.pop()
+				p.acknowledgeSelfClosingTag()
 			}
 			return true
 		case a.Head:
@@ -826,6 +830,7 @@ func inHeadIM(p *parser) bool {
 			p.addElement()
 			return true
 		}
+
 		switch p.tok.DataAtom {
 		case a.Head:
 			p.addLoc()
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -929,13 +929,13 @@ import * as $$module1 from 'react-bootstrap';`},
 </head>
 <div />
 `,
-			want: want{
-				styles: []string{
-					"{props:{\"global\":true},children:`div { color: red }`}",
-					"{props:{\"data-astro-id\":\"EX5CHM4O\"},children:`div.astro-EX5CHM4O{color:green;}`}",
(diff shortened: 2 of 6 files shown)


internal/transform/transform.go
```diff
@@ -23,67 +23,79 @@ func Transform(doc *tycho.Node, opts TransformOptions) *tycho.Node {
...
-		if n.Type == tycho.ElementNode {
-			switch n.DataAtom {
-			case a.Style:
-				// Do not extract <style> inside of SVGs
-				if n.Parent != nil && n.Parent.DataAtom == atom.Svg {
-					return
-				}
-				doc.Styles = append(doc.Styles, n)
-				// Remove local style node
-				n.Parent.RemoveChild(n)
```
natemoo-re, line 44: This was the problem!

internal/transform/transform.go
```diff
@@ -23,67 +23,79 @@ func Transform(doc *tycho.Node, opts TransformOptions) *tycho.Node {
...
+			if n.Parent != nil && n.Parent.DataAtom == atom.Svg {
+				return
 			}
+			doc.Styles = append(doc.Styles, n)
 		}
 	})
+	// Important! Remove styles from original location *after* walking the doc
+	for _, style := range doc.Styles {
+		style.Parent.RemoveChild(style)
+	}
```
natemoo-re, line 53: This was the solution!

jonathantneal (approved):
Yes. Looks good.

internal/transform/transform.go
```diff
@@ -23,67 +23,79 @@ func Transform(doc *tycho.Node, opts TransformOptions) *tycho.Node {
 	shouldScope := len(doc.Styles) > 0 && ScopeStyle(doc.Styles, opts)
 	walk(doc, func(n *tycho.Node) {
 		ExtractScript(doc, n)
+		AddComponentProps(doc, n)
```
natemoo-re, line 26: Split this into its own function for clarity

internal/transform/transform.go
```diff
@@ -23,67 +23,79 @@ func Transform(doc *tycho.Node, opts TransformOptions) *tycho.Node {
...
 		if shouldScope {
 			ScopeElement(n, opts)
 		}
 	})
+
+	// Important! Remove scripts from original location *after* walking the doc
+	for _, script := range doc.Scripts {
+		script.Parent.RemoveChild(script)
+	}
+
```
natemoo-re, line 36: This was the fix for scripts

internal/printer/printer_test.go
```diff
@@ -920,6 +920,22 @@ import * as $$module1 from 'react-bootstrap';`},
...
 		},
 		{
+			name: "Mixed style siblings",
+			source: `
+<head>
+	<style global>div { color: red }</style>
+	<style>div { color: green }</style>
+	<style>div { color: blue }</style>
+</head>
+<div />
```
natemoo-re, line 930: This was failing previously!

internal/parser.go
```diff
@@ -755,6 +756,7 @@ func inHeadIM(p *parser) bool {
 			if p.hasSelfClosingToken {
 				p.addLoc()
 				p.oe.pop()
+				p.acknowledgeSelfClosingTag()
```
natemoo-re, line 759: Caught a few unrelated bugs around self-closing tags

internal/transform/transform.go
drwpow: Wow this parser is slick! Love the flexibility of this.

internal/transform/transform.go
drwpow: Ack! Good catch!

drwpow (approved):
Very nice fix!

okikio:
Has this been published, it might be the cause of the bug we are facing on the repl, while testing on Safari

PR #130, opened 2021-11-02
Fix: `style` ordering

natemoo-re:
## Changes

- Followup to https://github.com/snowpackjs/astro/pull/1721. 
- Rather than reverse the final order from the `astro` side, the correct fix is to reverse the order per-file in the compiler.

**Previously,** text was _red_.
Screen Shot 2021-11-02 at 5 13 56 PM

**With this change,** text is _yellow_.

## Testing

Tests updated, but not testing this specifically

## Docs

Bug fix only

### DIFF
--- a/.changeset/quiet-dolphins-beam.md
+++ b/.changeset/quiet-dolphins-beam.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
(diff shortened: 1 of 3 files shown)


drwpow (approved):
👍🏻 Makes sense to me

PR #131, opened 2021-11-02
Preserve slots inside `<custom-element>`

natemoo-re:
## Changes

Fixes https://github.com/snowpackjs/astro/issues/1717.

Previously, we would compile all slots to use our custom Astro slot handling. For `custom-elements`, `slot` has different semantics and should be left up to the custom element itself.

This change passes through any slotted elements as written. 

## Testing

Test added

## Docs

Bug fix only

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -964,6 +964,13 @@ import * as $$module1 from 'react-bootstrap';`},
 				code: `<html><head></head><body>${$$renderComponent($$result,'Component',Component,{},{"named": () => $$render` + BACKTICK + `${$$renderComponent($$result,'Fragment',Fragment,{"slot":"named"},{"default": () => $$render` + BACKTICK + `<div>Default</div><div>Named</div>` + BACKTICK + `,})}` + BACKTICK + `,})}</body></html>`,
 			},
 		},
+		{
+			name:   "Preseve slots inside custom-element",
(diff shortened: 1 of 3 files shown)


internal/printer/printer_test.go
```diff
@@ -964,6 +964,13 @@ import * as $$module1 from 'react-bootstrap';`},
 				code: `<html><head></head><body>${$$renderComponent($$result,'Component',Component,{},{"named": () => $$render` + BACKTICK + `${$$renderComponent($$result,'Fragment',Fragment,{"slot":"named"},{"default": () => $$render` + BACKTICK + `<div>Default</div><div>Named</div>` + BACKTICK + `,})}` + BACKTICK + `,})}</body></html>`,
 			},
 		},
+		{
+			name:   "Preseve slots inside custom-element",
+			source: `<body><my-element><div slot=name>Name</div><div>Default</div></my-element></body>`,
+			want: want{
+				code: `<html><head></head><body>${$$renderComponent($$result,'my-element','my-element',{},{"default": () => $$render` + BACKTICK + `<div slot="name">Name</div><div>Default</div>` + BACKTICK + `,})}</body></html>`,
```
drwpow, line 971: Clever! Nice test, too

drwpow (approved):

graynorton:
Should the docs below be updated to note this nuance of behavior?

https://docs.astro.build/core-concepts/astro-components/#slots

PR #134, opened 2021-11-03
fix: better handling for text inside of elements inside of expressions

natemoo-re:
## Changes

- Followup to #122, which was an incomplete solution.
- This fixes handling of `'`, `"`, `` ` ``, and `/` character inside of elements that are inside of expressions.
- Closes https://github.com/snowpackjs/astro/issues/1478

## Testing

Tests added

## Docs

Bugfix only


matthewp:
lgtm

PR #135, opened 2021-11-05
Fix namespaced attribute printing

natemoo-re:
## Changes

- We had a bug in our `printAttribute` function that would print namespaced attributes in an invalid way. 
- This is fixed, unblocking attributes like `xmlnm:xlink`

## Testing

Test added

## Docs

Bug fix only


jonathantneal (approved):
Yes!

PR #137, opened 2021-11-05
Implement `getStaticPaths` hoisting

natemoo-re:
## Changes

- Fixes `getStaticPaths` behavior, because it allows `getStaticPaths` anywhere in the file. The function is hoisted to the top of the file.

## Testing

Tests added

## Docs

N/A Backwards compat

### DIFF
--- a/internal/printer/print-to-js.go
+++ b/internal/printer/print-to-js.go
@@ -5,7 +5,6 @@
 package printer
 
 import (
-	"errors"
 	"fmt"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


internal/printer/print-to-js.go
```diff
@@ -117,13 +116,17 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 				p.addSourceMapping(n.Loc[0])
 				if renderBodyStart == -1 {
 					p.addSourceMapping(c.Loc[0])
-					if js_scanner.AccessesPrivateVars([]byte(c.Data)) {
-						panic(errors.New("Variables prefixed by \"$$\" are reserved for Astro's internal usage!"))
-					}
```
jonathantneal, line 122: I like this change. Does the compiler now resolve conflicts differently?

jonathantneal (approved):
Let's lex!

internal/printer/print-to-js.go
natemoo-re: Nope, this warning never worked properly. I'd like to automatically namespace things so this isn't even necessary (_eventually_)

PR #138, opened 2021-11-06
fix: script hoist behavior

natemoo-re:
## Changes

See https://github.com/snowpackjs/astro/pull/1743

- When a component only included `<script hoist>`, we would print malformed JS because the `doc` was empty when we tried to print it.
- Now, if we hoist out all the contents, we inject an empty `FrontmatterNode` with an empty `TextNode` child.
- This allows the `printToJs` function to keep doing it's thing and returns valid JS.

## Testing

This is not tested because it is a difficult edge case to test for with our current setup

## Docs

Bug fix

### DIFF
--- a/internal/printer/print-to-js.go
+++ b/internal/printer/print-to-js.go
@@ -105,6 +105,7 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 
 	// Render frontmatter (will be the first node, if it exists)
 	if n.Type == FrontmatterNode {
+		fmt.Println("FrontmatterNode")
 		for c := n.FirstChild; c != nil; c = c.NextSibling {
 			if c.Type == TextNode {
 				p.printInternalImports(p.opts.InternalURL)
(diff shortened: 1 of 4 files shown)


internal/printer/print-to-js.go
```diff
@@ -105,6 +105,7 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 
 	// Render frontmatter (will be the first node, if it exists)
 	if n.Type == FrontmatterNode {
+		fmt.Println("FrontmatterNode")
```
matthewp, line 108: Woops

internal/printer/print-to-js.go
```diff
@@ -113,9 +114,13 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 				// `renderBodyStart` will be the index where we should split the frontmatter.
 				// If we don't encounter any of those, `renderBodyStart` will be `-1`
 				renderBodyStart := js_scanner.FindRenderBody([]byte(c.Data))
-				p.addSourceMapping(n.Loc[0])
+				if len(n.Loc) > 0 {
+					p.addSourceMapping(n.Loc[0])
```
drwpow, line 118: Nice!

drwpow (approved):
LGTM!

PR #140, opened 2021-11-11
feat: use `$$metadata.resolvePath` to support `client:only` directive

natemoo-re:
## Changes

- Adds `$$metadata.resolvePath()` to `client:only` components.
- Needed to unblock https://github.com/snowpackjs/astro/pull/1716

## Testing

Tests updated

## Docs

N/A, internal implementation


matthewp:
Looks good to me.

PR #142, opened 2021-11-12
Import meta env

drwpow:
## Changes
Given the following input:
```
---
export async function getStaticPaths() {
  const data = await fetch(`${import.meta.env.PUBLIC_API_URL}/api/posts`)
    .then((res) => res.json())
    .then(() => {
      console.log('data received')
    });
}
---
```

This is what it’d generate:

```diff
+ import * as $$module1 from 'data received';
  
  export async function getStaticPaths() {
    const data = await fetch(`${import.meta.env.PUBLIC_API_URL}/api/posts`)
      .then((res) => res.json())
      .then(() => {
        console.log('data received')
      });
  }
```

Basically, `import.meta.env` would be accidentally hoisted up as an import. As to why the `console.log` got the import I’m just guessing it’s because it’s the next string or something 🤷🏻 

Anyway, test added & bug sqooshed

## Testing

Test added; was failing before

## Docs

No docs needed

### DIFF
--- a/internal/js_scanner/js_scanner.go
+++ b/internal/js_scanner/js_scanner.go
@@ -359,6 +359,11 @@ func NextImportStatement(source []byte, pos int) (int, ImportStatement) {
 				if next == js.MulToken {
 					currImport.ExportName = string(nextValue)
 				}
+
+				// if this is import.meta.*, ignore (watch for first dot)
+				if next == js.DotToken && len(specifier) == 0 {
+					break
+				}
 			}
 		}
 
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -1058,6 +1058,101 @@ import * as $$module1 from 'react-bootstrap';`},
 				code: `<html><head></head><body><svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><rect xlink:href="#id"></rect></svg></body></html>`,
 			},
 		},
+		{
+			name: "Import scanning",
+			source: fmt.Sprintf(`---
+import Header from '../../components/Header.jsx'
+import Footer from '../../components/Footer.astro'
+import ProductPageContent from '../../components/ProductPageContent.jsx';
+
+export async function getStaticPaths() {
+  let products = await fetch(%s${import.meta.env.PUBLIC_NETLIFY_URL}/.netlify/functions/get-product-list%s)
+    .then(res => res.json()).then((response) => {
+      console.log('--- built product pages ---')
+      return response.products.edges
+    });
+
+  return products.map((p, i) => {
+    return {
+      params: {pid: p.node.handle},
+      props: {product: p},
+    };
+  });
+}
+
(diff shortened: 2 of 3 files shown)


internal/js_scanner/js_scanner.go
```diff
@@ -359,6 +359,11 @@ func NextImportStatement(source []byte, pos int) (int, ImportStatement) {
 				if next == js.MulToken {
 					currImport.ExportName = string(nextValue)
 				}
+
+				// if this is import.meta.*, ignore (watch for first dot)
+				if next == js.DotToken && len(specifier) == 0 {
```
drwpow, line 364: Tried to do a “safe” check here—basically if we encounter a '.' after “import” before we encounter anything else, don’t treat this as an import.

internal/printer/printer_test.go
```diff
@@ -45,8 +47,9 @@ type want struct {
 	styles         []string
 	scripts        []string
 	getStaticPaths string
+	code           string
+	skipHoist      bool // HACK: sometimes `getStaticPaths()` appears in a slightly-different location. Only use this if needed!
```
drwpow, line 51: So if there’s nothing between imports and exports, nothing will get hoisted because nothing needs to be. Unfortunately in our tests that means that `Astro` global and `$$metadata` can appear on either side, so this was my quick hack just for the test I added.

internal/printer/printer_test.go
```diff
@@ -181,13 +182,9 @@ import * as ns from '../components';
   </body>
 </html>`,
 			want: want{
-				frontmatter: []string{
-					`import * as ns from '../components';
-
-import * as $$module1 from '../components';`,
```
drwpow, line 187: Cleanup: you no longer need to specify these! 🎉 They’re just auto-generated from `metadata.modules` below.

internal/printer/printer_test.go
```diff
@@ -1093,6 +1160,19 @@ import * as $$module1 from 'react-bootstrap';`},
 			if len(tt.want.frontmatter) > 0 {
 				toMatch += test_utils.Dedent(tt.want.frontmatter[0])
 			}
+			// Fixes some tests where getStaticPaths appears in a different location
+			if tt.want.skipHoist == true && len(tt.want.getStaticPaths) > 0 {
+				toMatch += "\n\n"
+				toMatch += strings.TrimSpace(test_utils.Dedent(tt.want.getStaticPaths)) + "\n"
```
drwpow, line 1166: Fortunately the `skipHoist` “hack“ isn’t too bad. It only duplicates the `getStaticPaths` printing, which is workable.

The other additions here handle the cleanup of having to declare modules twice.

internal/printer/printer_test.go
drwpow: To be clear: it’s probably not a bug; just that hoisting can affect the print-to-js order which affects our tests

matthewp:
lgtm

natemoo-re (approved):
Whoops. LGTM!

PR #144, opened 2021-11-12
fix: drop JSX/TSX support from Frontmatter

natemoo-re:
## Changes

- This ended up being a pretty small change in the tokenizer, basically to ignore the current logic that any `<` characters should start an element.
- Didn't remove any logic from the parser/renderer that might handle this case. Those are dead code paths now, can cleanup later.

## Testing

Existing tests updated to reflect our new approach

## Docs

N/A

### DIFF
--- a/internal/token_test.go
+++ b/internal/token_test.go
@@ -265,23 +265,23 @@ func TestFrontmatter(t *testing.T) {
 			[]TokenType{FrontmatterFenceToken, TextToken, FrontmatterFenceToken, TextToken, StartTagToken, TextToken, EndTagToken, TextToken},
 		},
 		{
-			"tokenizes elements inside",
+			"does not tokenize elements inside",
 			`
 			---
 			const a = <div />;
 			---
 			`,
-			[]TokenType{FrontmatterFenceToken, TextToken, SelfClosingTagToken, TextToken, FrontmatterFenceToken},
+			[]TokenType{FrontmatterFenceToken, TextToken, TextToken, FrontmatterFenceToken},
 		},
 		{
-			"elements can have expression as child in frontmatter",
+			"no elements or expressions in frontmatter",
 			`
 			---
 			const contents = "foo";
 			const a = <div>{contents}</div>;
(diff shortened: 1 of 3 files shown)


internal/token_test.go
```diff
@@ -265,23 +265,23 @@ func TestFrontmatter(t *testing.T) {
...
 		{
-			"tokenizes elements inside",
+			"does not tokenize elements inside",
 			`
 			---
 			const a = <div />;
 			---
 			`,
-			[]TokenType{FrontmatterFenceToken, TextToken, SelfClosingTagToken, TextToken, FrontmatterFenceToken},
+			[]TokenType{FrontmatterFenceToken, TextToken, TextToken, FrontmatterFenceToken},
```
drwpow, line 274: 👍🏻

internal/token_test.go
```diff
@@ -265,23 +265,23 @@ func TestFrontmatter(t *testing.T) {
...
-			"elements can have expression as child in frontmatter",
+			"no elements or expressions in frontmatter",
 			`
 			---
 			const contents = "foo";
 			const a = <div>{contents}</div>;
 			---
 			`,
-			[]TokenType{FrontmatterFenceToken, TextToken, TextToken, StartTagToken, StartExpressionToken, TextToken, EndExpressionToken, EndTagToken, TextToken, FrontmatterFenceToken},
+			[]TokenType{FrontmatterFenceToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, FrontmatterFenceToken},
```
drwpow, line 284: Yeah to your PR comment, not important for this PR at all, but we could probably collapse all the text tokens into one now.

drwpow (approved):
Nice!

internal/token_test.go
natemoo-re: The parser collapses any adjacent TextTokens into a single TextNode, so I'm not sure if we need to tackle this in the tokenizer.

PR #145, opened 2021-11-12
fix: expressions inside of <select>

natemoo-re:
## Changes

- Closes https://github.com/snowpackjs/astro/issues/1807
- Fixes expressions inside of `select`

## Testing

Added two tests for common use cases.

## Docs

Bug fix

### DIFF
--- a/.changeset/polite-ties-destroy.md
+++ b/.changeset/polite-ties-destroy.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fix regression with expressions inside of <select> elements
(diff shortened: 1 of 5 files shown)


jonathantneal (approved):
Wonderful! It looks like [this fix also covers `<textarea>`](https://github.com/snowpackjs/astro-compiler-next/blob/10dc62cdf5b1daa158e05b3797ae6b9f4c71d4ab/internal/parser.go#L1226).

Would it be good to add a test for `<textarea>` as well, to be sure it remains covered?

natemoo-re:
@jonathantneal Just added tests and a fix for `textarea`.

jonathantneal:
@natemoo-re, hurray! Good fix.

PR #146, opened 2021-11-15
Fix for weird \x00 issue

matthewp:
## Changes

- Fixes weird issue where import statements were removed.
- Using a pointer to the Metadata object rather than putting it directly on the Node (maybe tinygo has trouble with that pattern?)

## Testing

No, but will fix tests in `astro` and remove a workaround.

## Docs

N/A


natemoo-re (approved):
LGTM! Nice catch.

drwpow:
HOW DID YOU FIGURE THIS OUT!?!?!?!?111

aykevl:
Interesting! I know little about Astro but this sounds like a bug in TinyGo. If you manage to reduce this to a small reproducer and report it upstream that would be great!

PR #149, opened 2021-11-16
fix: comments and strings inside of expression attributes

natemoo-re:
Supersedes https://github.com/snowpackjs/astro/issues/1833

## Changes

- Previously, comments and strings weren't being handled inside of expression attributes
- We also weren't able to support comments inside of a tag, but now we can (because they're expression attributes)

## Testing

Tokenizer tests added

## Docs

Bug fix


drwpow (approved):
Looking good! Nice test

PR #151, opened 2021-11-18
fix: handle `defer` in LIFO order

natemoo-re:
## Changes

- We have been noticing lots of `call to released function` errors lately...
- After some digging, it turns out that `preprocessStyle` was queueing `defer`'d callbacks incorrectly.
- This updates `preprocessStyle` to properly account for `defer`'s LIFO behavior.

## Testing

Tested manually, very hard to automatically test

## Docs

Bug fix only


matthewp:
lgtm

internal_wasm/utils/utils.go
```diff
@@ -9,25 +9,27 @@ import (
 // See https://stackoverflow.com/questions/68426700/how-to-wait-a-js-async-function-from-golang-wasm
 func Await(awaitable js.Value) ([]js.Value, []js.Value) {
 	then := make(chan []js.Value)
-	defer close(then)
 	thenFunc := js.FuncOf(func(this js.Value, args []js.Value) interface{} {
 		then <- args
 		return nil
 	})
+	// defers are called LIFO!
```
drwpow, line 16: 🙇🏻 Thank you! This is very helpful

internal_wasm/utils/utils.go
```diff
@@ -9,25 +9,27 @@ import (
...
+	// This will `close` before `Release()`
+	defer thenFunc.Release()
```
drwpow, line 18: Unknown, but how does the `.Call("then" …).Call("catch")` work? Do we need to wait to release `thenFunc` until `catch` has been called? Or no?

lib/compiler/test/basic.test.mjs
```diff
@@ -43,7 +43,6 @@ div {
 `,
     {
       sourcemap: true,
-      // HOLY CRAP THIS ACTUALLY WORKS!
```
drwpow, line 46: RIP

drwpow (approved):
Looks good! Only have a question but this definitely seems like an improvement

internal_wasm/utils/utils.go
natemoo-re: All `defers` will be called after `return`, so they clean themselves up automatically.

PR #153, opened 2021-11-18
Change how imports are printed

matthewp:
## Changes

- Do it the hard way
- This fixes the OSX `\x00` issues. I don't know why. I just know that it does.

## Testing

Fixes tests in /astro

## Docs

N/A


jonathantneal (approved):
I get it.

drwpow (approved):
Whoa! Could you explain… uh…

> I don't know why. I just know that it does.

😶 nevermind carry on good good good

PR #154, opened 2021-11-18
Fix leading `style` and `script hoist`

natemoo-re:
## Changes

- In most cases, our parser is smart enough to detect when the only element is a `Component` node and skip generating the implied `html`, `head`, and `body` tags. This happens if the `Component` comes first.
- If there is a `style` or `script hoist` tag above the `Component` node, our parser isn't able to lookahead and determine if `html`, `head`, and `body` should be generated. They always are. 
- This PR adds a `transform` step that corrects the parser output. This happens here because this is the step where we walk the document and make any mutations based on contextual information.

## Testing

New test suite added to `transform`

## Docs

Bug fix

### DIFF
--- a/.changeset/heavy-masks-jump.md
+++ b/.changeset/heavy-masks-jump.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fix handling of top-level component nodes with leading styles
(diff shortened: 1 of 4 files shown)


drwpow (approved):
Nice fix! Good tests, too.

PR #156, opened 2021-11-18
[wip] tinygo 0.21.0

drwpow:
## Changes

This PR just tracks if the next version of tinygo is working or not. If there’s a big ❌ at the bottom… you can take a wild guess

## Testing

No tests added.

## Docs

No docs

### DIFF
--- a/.changeset/tiny-vans-help.md
+++ b/.changeset/tiny-vans-help.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+chore: Upgrade tinygo to 0.21.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,8 +37,8 @@ jobs:
 
       - name: Get tinygo runtime
         run: |
-          wget -P /tmp/ https://github.com/tinygo-org/tinygo/releases/download/v0.20.0/tinygo_0.20.0_amd64.deb
-          sudo dpkg -i /tmp/tinygo_0.20.0_amd64.deb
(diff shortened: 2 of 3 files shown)


natemoo-re:
👀 Hooray, new WASM bugs!

drwpow:
> 👀 Hooray, new WASM bugs!

loool

janimo:
0.21 only works if building with -gc leaking. The default conservative gc seems to have a regression and the .wasm file is also almost 300K larger than with 0.20.

matthewp:
Looks like we're not going to be merging this for now. Thanks for the info @janimo, we'll keep an eye on improvements on the Go side, if you happen to find out that they've fixed these regressions please let us know if you can.

aykevl:
> and the .wasm file is also almost 300K larger than with 0.20.

That's a significant increase!
I suspect this is because of [Asyncify](https://emscripten.org/docs/porting/asyncify.html) which we now use for concurrency (in short, the old method was subtly broken but that only became visible once we upgraded to LLVM 12). If you don't need concurrency, you can probably shrink the binary size a lot using `-scheduler=none`.

> 0.21 only works if building with -gc leaking. The default conservative gc seems to have a regression

Can you please make a small reproducer and report this upstream? We haven't seen any GC issues lately and I can't think of anything that would have broken GC - except if Asyncify somehow managed to confuse things.

janimo:
@aykevl I remember trying with -scheduler=none, and while it resulted in a really small binary, it crashed regardless of the -gc option used. The only testcase I can think of is cloning this repo, make astro-wasm, npm install and npm run test which crashes. I didn't look deeper, sorry.

PR #158, opened 2021-11-20
fix: expressions inside of `<table>`

natemoo-re:
## Changes

- Closes https://github.com/snowpackjs/astro/issues/1939
- We weren't handling expressions inside of the `table` or `tbody` insertion modes, which was causing expressions to be fostered to another parent.
- This PR implements expression handling inside of the `table` and `tbody` insertion modes. Importantly (not reflected in original report) this change will not create an implicit `tbody` or `thead` if the only child of `table` is an expression.

## Testing

Tests added

## Docs

Bug fix only

### DIFF
--- a/.changeset/spotty-buckets-accept.md
+++ b/.changeset/spotty-buckets-accept.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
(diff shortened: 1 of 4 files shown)


drwpow (approved):
Nice test!

PR #159, opened 2021-11-20
Fix `form > textarea` insertion mode

natemoo-re:
## Changes

- Closes https://github.com/snowpackjs/astro/issues/1953
- Added a failing test, then made sure it passed.
- Added a fix that reset the `originalIM` if we hit `StartTagToken` inside of `inBodyIM` when `originalIM != nil`. We now properly handle this edge case.

## Testing

Test added

## Docs

Bug fix only


drwpow (approved):
Nice fix!

PR #162, opened 2021-11-22
fix: make all element checks case sensitive

natemoo-re:
## Changes

- Closes https://github.com/snowpackjs/astro/issues/1974
- Turns out not all of our tag checks were case-sensitive, so `Base` was being treated as a void tag (`base`).
- This fixes the case-sensitivity of the tokenizer.

## Testing

Test added. Test that _was_ passing but should have been failing updated to reflect the correct behavior.

## Docs

Bug fix only.

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -969,7 +969,7 @@ import { Container, Col, Row } from 'react-bootstrap';
 			want: want{
 				frontmatter: []string{`import { Container, Col, Row } from 'react-bootstrap';`},
 				metadata:    metadata{modules: []string{`{ module: $$module1, specifier: 'react-bootstrap' }`}},
-				code:        "${$$renderComponent($$result,'Container',Container,{},{\"default\": () => $$render`${$$renderComponent($$result,'Row',Row,{},{\"default\": () => $$render`${$$renderComponent($$result,'Col',Col,{})}<h1>Hi!</h1>`,})}`,})}\n.",
+				code:        "${$$renderComponent($$result,'Container',Container,{},{\"default\": () => $$render`${$$renderComponent($$result,'Row',Row,{},{\"default\": () => $$render`${$$renderComponent($$result,'Col',Col,{},{\"default\": () => $$render`<h1>Hi!</h1>`,})}`,})}`,})}",
 			},
 		},
 		{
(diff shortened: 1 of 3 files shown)


internal/printer/printer_test.go
```diff
@@ -969,7 +969,7 @@ import { Container, Col, Row } from 'react-bootstrap';
 			want: want{
 				frontmatter: []string{`import { Container, Col, Row } from 'react-bootstrap';`},
 				metadata:    metadata{modules: []string{`{ module: $$module1, specifier: 'react-bootstrap' }`}},
-				code:        "${$$renderComponent($$result,'Container',Container,{},{\"default\": () => $$render`${$$renderComponent($$result,'Row',Row,{},{\"default\": () => $$render`${$$renderComponent($$result,'Col',Col,{})}<h1>Hi!</h1>`,})}`,})}\n.",
```
natemoo-re, line 972: This was wrong! Should have caught that sooner!

internal/printer/printer_test.go
```diff
@@ -1194,6 +1194,13 @@ const items = ["Dog", "Cat", "Platipus"];
 				code: `<html><body>${$$renderComponent($$result,'Component',Component,{},{"default": () => $$render` + BACKTICK + `<form><textarea></textarea></form>` + BACKTICK + `,})}</body></html>`,
 			},
 		},
+		{
+			name:   "slot inside of Base",
+			source: `<Base title="Home"><div>Hello</div></Base>`,
+			want: want{
+				code: `${$$renderComponent($$result,'Base',Base,{"title":"Home"},{"default": () => $$render` + BACKTICK + `<div>Hello</div>` + BACKTICK + `,})}`,
```
natemoo-re, line 1201: Prior to this change, `Base` could not accept any slots because it was being closed as if it were the void tag `<base>`.

internal/token.go
```diff
@@ -915,9 +915,6 @@ loop:
 		}
 		for i := 0; i < len(s); i++ {
 			c := z.buf[z.data.Start+i]
-			if 'A' <= c && c <= 'Z' {
-				c += 'a' - 'A'
-			}
```
natemoo-re, line 920: Missed a few spots. This is essentially Go's `.toLowerCase()` for a `char`.

drwpow (approved):
Nice!

jonathantneal (approved):

PR #163, opened 2021-11-23
Fix issue where components can only add one style or script

jonathantneal:
## Changes

- Fixes issue where only one script or style may be added from an Astro component.
  - **Reason**: Astro Compiler spreads arguments to `Set#add` which takes one argument and not a spread.
  - **Solution**: Iterate scripts or styles and add them individually to their respective set.

## Testing

Tests are updated in `printer_test.go`

I would love to learn how I could write a specific test that would have caught this issue.

## Docs

bug fix only

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -36,7 +36,7 @@ export default $$Component;`
 var STYLE_PRELUDE = "const STYLES = [\n"
 var STYLE_SUFFIX = "];\nfor (const STYLE of STYLES) $$result.styles.add(STYLE));\n"
 var SCRIPT_PRELUDE = "const SCRIPTS = [\n"
-var SCRIPT_SUFFIX = "];\nfor (const SCRIPT of SCRIPTS) $$result.scripts.add(STYLE));\n"
+var SCRIPT_SUFFIX = "];\nfor (const SCRIPT of SCRIPTS) $$result.scripts.add(SCRIPT));\n"
 var CREATE_ASTRO_CALL = "const $$Astro = $$createAstro(import.meta.url, 'https://astro.build');\nconst Astro = $$Astro;"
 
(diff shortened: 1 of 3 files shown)


internal/printer/printer_test.go
```diff
@@ -34,9 +34,9 @@ var SUFFIX = fmt.Sprintf("%s;", BACKTICK) + `
 });
 export default $$Component;`
 var STYLE_PRELUDE = "const STYLES = [\n"
-var STYLE_SUFFIX = "];\n$$result.styles.add(...STYLES);\n"
+var STYLE_SUFFIX = "];\nfor (const STYLE of STYLES) $$result.styles.add(STYLE));\n"
 var SCRIPT_PRELUDE = "const SCRIPTS = [\n"
-var SCRIPT_SUFFIX = "];\n$$result.scripts.add(...SCRIPTS);\n"
+var SCRIPT_SUFFIX = "];\nfor (const SCRIPT of SCRIPTS) $$result.scripts.add(STYLE));\n"
```
matthewp, line 39: should be `.add(SCRIPT)`

matthewp (commented):
lgtm overall, one thing that should fix the tests.

matthewp:
Also needs a changeset.

internal/printer/printer_test.go
jonathantneal:
```suggestion
var SCRIPT_SUFFIX = "];\nfor (const SCRIPT of SCRIPTS) $$result.scripts.add(SCRIPT));\n"
```

jonathantneal:
Thanks for the review! Patched and changeset added, @matthewp.

matthewp (approved):
lgtm

natemoo-re (approved):
Nice catch, thank you!

PR #164, opened 2021-11-23
Support expressions in SVG elements

jonathantneal:
## Changes

- Fixes an issue where expressions did not work within SVG elements.
- Follows behavior of `inBodyIM`.

## Testing

- added simple test — `<svg>{title ?? null}</svg>`.
- added nested test — `<svg>{title ? <title>{title}</title> : null}</svg>`.

## Docs

bug fix only

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -2533,7 +2533,6 @@ func ignoreTheRemainingTokens(p *parser) bool {
 
 const whitespaceOrNUL = whitespace + "\x00"
 
-// TODO: handle expressions in SVG
 // Section 12.2.6.5
 func parseForeignContent(p *parser) bool {
 	switch p.tok.Type {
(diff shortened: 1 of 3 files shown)


internal/parser.go
```diff
@@ -2533,7 +2533,6 @@ func ignoreTheRemainingTokens(p *parser) bool {
 
 const whitespaceOrNUL = whitespace + "\x00"
 
-// TODO: handle expressions in SVG
```
natemoo-re, line 2536: 😄

internal/parser.go
```diff
@@ -2612,6 +2611,14 @@ func parseForeignContent(p *parser) bool {
 			}
 		}
 		return true
+	case StartExpressionToken:
+		p.reconstructActiveFormattingElements()
+		p.addExpression()
```
natemoo-re, line 2616: Just FYI we do have `expressionIM`, but I think we can circle back to implementing it here. I tried it out and it was causing some problems with `setOriginalIM()` being called twice.

I think our insertion mode algorithm probably needs to be updated to use a stack for improved spec compliance. Again, outside of the scope of this PR.

PR #167, opened 2021-11-23
Fix leading `<style>` regression

natemoo-re:
## Changes

- Closes https://github.com/snowpackjs/astro/issues/1989
- The problem here was a regression in the handling of leading `<script hoist>` or `<style>` nodes (implemented in https://github.com/snowpackjs/astro-compiler/pull/154)
- The fix required two changes:
  - Tracking which elements are implicitly generated while parsing the document.
  - Being more intelligent about repairing a document after elements have been hoisted out

## Testing

Many tests added

## Docs

Bug fix only

### DIFF
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -39,20 +39,35 @@ func Transform(doc *tycho.Node, opts TransformOptions) *tycho.Node {
 	// Since we can't detect a "component-only" file until after `parse`, we need to handle
 	// them here. The component will be hoisted to the root of the document, `html` and `head` will be removed.
 	if opts.As != "Fragment" {
-		hasBody := false
 		var onlyComponent *tycho.Node
+		var rootNode *tycho.Node
 		walk(doc, func(n *tycho.Node) {
-			if n.Component && n.Parent != nil && n.Parent.DataAtom == a.Head {
-				onlyComponent = n
-			} else if !hasBody && n.DataAtom == a.Body {
-				hasBody = true
+			if p := n.Parent; n.Component && p != nil && (p.DataAtom == a.Head || p.DataAtom == a.Body) {
+				if !hasSiblings(n) {
+					onlyComponent = n
+				}
+				return
+			}
+			if n.DataAtom == a.Html && (!IsImplictNode(n) || childCount(n) == 1) {
+				rootNode = n
+				return
 			}
 		})
 
-		if !hasBody && onlyComponent != nil {
-			onlyComponent.Parent.RemoveChild(onlyComponent)
-			doc.AppendChild(onlyComponent)
-			doc.RemoveChild(onlyComponent.PrevSibling)
+		if rootNode == nil {
+			rootNode = doc
+		}
+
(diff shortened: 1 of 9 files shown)


internal/transform/transform.go
```diff
@@ -39,20 +39,35 @@ func Transform(doc *tycho.Node, opts TransformOptions) *tycho.Node {
...
-				hasBody = true
+			if p := n.Parent; n.Component && p != nil && (p.DataAtom == a.Head || p.DataAtom == a.Body) {
+				if !hasSiblings(n) {
+					onlyComponent = n
+				}
+				return
+			}
+			if n.DataAtom == a.Html && (!IsImplictNode(n) || childCount(n) == 1) {
+				rootNode = n
+				return
```
natemoo-re, line 53: If the user explicitly authored `<html></html>` as the root element, don't remove it.

internal/transform/transform.go
```diff
@@ -39,20 +39,35 @@ func Transform(doc *tycho.Node, opts TransformOptions) *tycho.Node {
...
+		if onlyComponent != nil {
+			p := onlyComponent.Parent
+			if IsImplictNode(p) {
+				onlyComponent.Parent.RemoveChild(onlyComponent)
+				rootNode.AppendChild(onlyComponent)
+				rootNode.RemoveChild(onlyComponent.PrevSibling)
+				if rootNode.FirstChild != nil && IsImplictNode(rootNode.FirstChild) {
+					rootNode.RemoveChild(rootNode.FirstChild)
+				}
+			}
```
natemoo-re, line 70: We're being much smarter now. This behavior only kicks in when the parent `html`, `head`, or `body` element was implicitly generated.

internal/transform/transform.go
```diff
@@ -153,3 +168,27 @@ func walk(doc *tycho.Node, cb func(*tycho.Node)) {
...
+		return false
+	}
+
+	if sibling := n.NextSibling; sibling != nil {
+		if sibling.Type == tycho.TextNode {
+			return strings.TrimSpace(n.NextSibling.Data) != ""
+		} else {
+			return sibling.Type != tycho.CommentNode
+		}
+	}
```
natemoo-re, line 183: Empty text nodes are barely even siblings, amirite?

internal/transform/utils.go
```diff
@@ -25,6 +25,25 @@ func HasAttr(n *astro.Node, key string) bool {
 	return false
 }
 
+func IsImplictNode(n *astro.Node) bool {
+	return HasAttr(n, astro.ImplicitNodeMarker)
+}
+
+func IsImplictNodeMarker(attr astro.Attribute) bool {
+	return attr.Key == astro.ImplicitNodeMarker
+}
```
natemoo-re, line 34: These utils are a much nicer abstraction than checking for `\x00implicit`

matthewp:
Interesting solution. Hope it works! :)

matthewp:
lgtm

jonathantneal (approved):

PR #170, opened 2021-11-23
Fix typo

natemoo-re:
## Changes

- Files with a single root component but no trailing whitespace were throwing because I had a copy+paste typo.

## Testing

Tests added to catch this in the future

## Docs

Bug fix

### DIFF
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -184,7 +184,7 @@ func hasSiblings(n *tycho.Node) bool {
 
 	if sibling := n.PrevSibling; sibling != nil {
 		if sibling.Type == tycho.TextNode {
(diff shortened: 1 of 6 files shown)


internal/transform/transform.go
```diff
@@ -184,7 +184,7 @@ func hasSiblings(n *tycho.Node) bool {
 
 	if sibling := n.PrevSibling; sibling != nil {
 		if sibling.Type == tycho.TextNode {
-			return strings.TrimSpace(n.NextSibling.Data) != ""
+			return strings.TrimSpace(n.PrevSibling.Data) != ""
```
natemoo-re, line 187: D'oh. This was a tough one to track down.

jonathantneal (approved):

PR #174, opened 2021-11-24
Fix CSS comments

drwpow:
## Changes

Fixes https://github.com/withastro/astro/issues/1992. Simple fix.

## Testing

Test added.

## Docs

Bug fix only

### DIFF
--- a/internal/transform/scope-css.go
+++ b/internal/transform/scope-css.go
@@ -47,6 +47,8 @@ outer:
 				} else {
 					break walk // an unrecoverable error occurred, or the end has been reached
 				}
+			case css.CommentGrammar:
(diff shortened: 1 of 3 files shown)


internal/transform/scope-css.go
```diff
@@ -47,6 +47,8 @@ outer:
 				} else {
 					break walk // an unrecoverable error occurred, or the end has been reached
 				}
+			case css.CommentGrammar:
```
drwpow, line 50: Default case below adds a semicolon. Simple as that.

matthewp:
lgtm!

natemoo-re (approved):
Nice fix!

PR #176, opened 2021-11-27
Remove GOPATH references.

janimo:
The code does not need to be checked out under GOPATH to build.

## Changes
Remove unneeded pkg/ prerequisites in Makefile, and references to GOPATH from the docs.
With new Go versions and using modules, packages don't need to be checked out under GOPATH.

## Testing

make astro
make astro-wasm

## Docs
this affects dev docs only.


natemoo-re:
Thanks, appreciate the PR!

PR #177, opened 2021-11-29
feat: implement new `fragment` behavior to support `head` and `body`

natemoo-re:
## Changes

- Fixes https://github.com/withastro/astro/issues/2031
- Allows `body` and `head` (any arbitrary tag) inside of components when using `as: "fragment"`.
- This may become the default behavior (eventually).

## Testing

WASM test added as that's the only area this behavior is surfaced

## Docs

Bug fix

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -1082,6 +1082,10 @@ func inBodyIM(p *parser) bool {
 		}
 		switch p.tok.DataAtom {
 		case a.Html:
+			if p.inTemplateFragmentContext() {
+				p.addElement()
+				return true
+			}
 			if p.oe.contains(a.Template) {
 				return true
 			}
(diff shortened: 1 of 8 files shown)


jonathantneal (approved):

internal/parser.go
```diff
@@ -1082,17 +1082,25 @@ func inBodyIM(p *parser) bool {
 		}
 		switch p.tok.DataAtom {
 		case a.Html:
+			if p.inTemplateFragmentContext() {
```
drwpow, line 1085: 👍🏻 This makes sense to me

internal/parser.go
```diff
@@ -2157,7 +2175,7 @@ func inSelectInTableIM(p *parser) bool {
 // Section 12.2.6.4.18.
 func inTemplateIM(p *parser) bool {
 	switch p.tok.Type {
-	case TextToken, CommentToken, DoctypeToken:
+	case TextToken, CommentToken, DoctypeToken, FrontmatterFenceToken:
```
drwpow, line 2178: Ooh good catch

drwpow (approved):

matthewp:
lgtm

PR #179, opened 2021-11-29
Pass projectRoot through to `createAstro`

matthewp:
## Changes

- Passes `projectRoot` through to the `createAstro` function in internals.
- This will be used to make URLs created by `Astro.resolve` be projectRoot relative
- `/src/...` instead of `/Users/name/...`

## Testing

Tested in Astro mostly, tests updated here.

## Docs

N/A


jonathantneal (approved):
Yes, please!

drwpow (approved):

PR #182, opened 2021-11-29
Add invalid CSS tests

drwpow:
## Changes

I was tracking down some compiler weirdness with CSS and wanted to add some tests for invalid CSS. Fortunately, these pass! So I think it’s worth adding just to track against regressions here.

No code changes.

## Testing

test only

## Docs

no docs


matthewp:
lgtm

natemoo-re (approved):
LGTM!

PR #184, opened 2021-11-30
fix:fragment frontmatter is optional

natemoo-re:
## Changes

- Missed the case where a component file didn't have any frontmatter.
- Added a `nil` check.
- Added a test!

## Testing

WASM test added

## Docs

Bug fix only


matthewp:
lgtm

PR #186, opened 2021-12-01
Fix expressions inside `<template>`

natemoo-re:
## Changes

- Fix infinite loop in `fragment` mode
- Expressions were not being handled inside of `template`, which was causing an infinite loop on top-level expressions without a parent element.

## Testing

WASM test added

## Docs

Bug fix only

### DIFF
--- a/lib/compiler/test/slotted-api.test.mjs
+++ b/lib/compiler/test/slotted-api.test.mjs
@@ -1,0 +1,23 @@
+/* eslint-disable no-console */
+import { transform } from '@astrojs/compiler';
+
+const contents = `
+{Astro.slots.a && <div id="a">
+  <slot name="a" />
+</div>}
+`
+
+async function run() {
+  const result = await transform(contents, {
+    sourcemap: true,
+    as: 'fragment',
+    site: undefined,
+    sourcefile: 'MoreMenu.astro',
+    sourcemap: 'both',
+    internalURL: 'astro/internal',
+  });
+
+  console.log(result.code);
+}
+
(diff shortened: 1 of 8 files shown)


lib/compiler/test/slotted-api.test.mjs
```diff
@@ -0,0 +1,23 @@
...
+  const result = await transform(contents, {
+    sourcemap: true,
+    as: 'fragment',
+    site: undefined,
+    sourcefile: 'MoreMenu.astro',
+    sourcemap: 'both',
+    internalURL: 'astro/internal',
+  });
+
+  console.log(result.code);
```
drwpow, line 20: Remove?

lib/compiler/test/slotted-api.test.mjs
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable no-console */
+import { transform } from '@astrojs/compiler';
+
+const contents = `
+{Astro.slots.a && <div id="a">
```
drwpow, line 5: Could this be a Go test? Or is this better was a WASM test?

drwpow (approved):
LGTM!

lib/compiler/test/slotted-api.test.mjs
natemoo-re: Our go tests don't do a great job of testing `as: 'fragment'` behavior right now because they're handled in the WASM entrypoint.

I'd like to fix that but it's a follow-up IMO.

PR #200, opened 2021-12-02
Fix edge case with conditional scripts

natemoo-re:
## Changes

- `XElement` was breaking because it had an expression with a `script` in it.
- This fixes that edge case

## Testing

Minimal test added

## Docs

Bug fix only

### DIFF
--- a/.changeset/unlucky-queens-accept.md
+++ b/.changeset/unlucky-queens-accept.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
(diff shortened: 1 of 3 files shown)


drwpow (approved):
I’m not sure I fully understand this but I see a good test so I’m clicking the green button

matthewp (approved):
lgtm

jonathantneal (approved):

aFuzzyBear:
💚 this,

PR #206, opened 2021-12-03
fix: regression with active formatting elements

natemoo-re:
## Changes

- Closes https://github.com/withastro/astro/issues/2104
- Fixes regression with active formatting elements when Components follow expressions.

## Testing

Test added, as well as an additional regression test

## Docs

Bug fix only

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -1337,9 +1337,7 @@ func inBodyIM(p *parser) bool {
 		case a.Caption, a.Col, a.Colgroup, a.Frame, a.Tbody, a.Td, a.Tfoot, a.Th, a.Thead, a.Tr:
 			// Ignore the token.
 		default:
-			if !isComponent(p.tok.Data) {
-				p.reconstructActiveFormattingElements()
-			}
+			p.reconstructActiveFormattingElements()
 			p.addElement()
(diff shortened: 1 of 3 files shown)


internal/parser.go
```diff
@@ -2499,6 +2499,9 @@ func frontmatterIM(p *parser) bool {
 }
 
 func inExpressionIM(p *parser) bool {
+	// p.afe (active formatting elements) should be cleared when entering an expression
+	p.clearActiveFormattingElements()
```
natemoo-re, line 2503: This is the fix! When we encounter an expression, we should clear the context of `p.afe` (active formatting elements) to avoid mutating the authored content.

internal/parser.go
```diff
@@ -2553,9 +2556,6 @@ func inExpressionIM(p *parser) bool {
 		return true
 	}
 	p.im = p.originalIM
-	if p.im == nil {
-		p.im = inBodyIM
-	}
```
natemoo-re, line 2558: Unrelated fix, but this could obscure a valid panic if author did not set `originalIM` before entering `inExpressionIM`.

drwpow (approved):
Nice fix! Always reassuring to see something fixed by simplifying the code

PR #208, opened 2021-12-06
Add support for static CSS extraction behind a flag

matthewp:
## Changes

- This add the ability for CSS to be extracted statically, rather than as a side-effect of render().
- This is intended for static builds and SSR support.
- This option is needed: `{ experimentalStaticExtraction: true }`
- Works by making CSS another return result of the compiler and then by adding import statements.
  - Note that these import statements only work in Vite. In other environments we should probably not add them; maybe a `target: 'vite' | '?'` would be useful for this.

## Testing

Tests added for CSS extraction. In the future if this becomes on by default we'll want to update the old tests.

## Docs

N/A


internal/printer/printer_css_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	astro "github.com/withastro/compiler/internal"
+	"github.com/withastro/compiler/internal/test_utils"
+	"github.com/withastro/compiler/internal/transform"
+)
+
+func TestPrinterCSS(t *testing.T) {
+	longRandomString := ""
+	for i := 0; i < 4080; i++ {
+		longRandomString += string(NON_WHITESPACE_CHARS[rand.Intn(len(NON_WHITESPACE_CHARS))])
```
drwpow, line 17: You might not need this… this was to test that length bug where our parser only parsed `4080` characters at a time

drwpow (approved):
LGTM! Not blocking but if you could add a wasm test if you have time that’d be great. Great Go tests though!

natemoo-re (approved):
Nice! Looks like a sound approach.

PR #211, opened 2021-12-07
Fix issue with import assertions creating additional imports

jonathantneal:
## Changes

- Fixes issue where imports with `assert`.
- Imports with `assert` would generate additional imports for the string tokens within the assertion object.

## Testing

printer test added

## Docs

bug fix only


FredKSchott:
LGTM but I'm not super familiar with this code!

internal/printer/printer.go
```diff
@@ -349,7 +356,11 @@ func (p *printer) printComponentMetadata(doc *astro.Node, source []byte) {
 		if i > 1 {
 			p.print(", ")
 		}
-		p.print(fmt.Sprintf("{ module: $$module%v, specifier: '%s' }", i, specs[i-1]))
+		asrt := "{}"
+		if asrts[i-1] != "" {
+			asrt = asrts[i-1]
+		}
+		p.print(fmt.Sprintf("{ module: $$module%v, specifier: '%s', assert: %s }", i, specs[i-1], asrt))
```
matthewp, line 363: doesn't this need to be quoted? `"%s"`?

matthewp: oh never mind, i see from the test below that it doesn't

matthewp (commented):
lgtm. Curious what the use-case is making this part of the metadata. Don't see any reason why not to have it though.

jonathantneal:
> Curious what the use-case is making this part of the metadata.

That’s a great callout, @matthewp. From what I observed, this is a current requirement of the library. It may also make sense to keep it. Here’s what I observed:

- Imports are duplicated as a namespace import, `import * as $$module%v from '%s'%s`. <sup>[reference](https://github.com/withastro/compiler/pull/211/files#diff-58de1575bcd3abab85340eb61810590c68a061ec2cad49d8cf3864fe402a64c1R338)</sup>
- The second import is generated from import metadata.
- To have the imports match, the assertion metadata is added.

internal/printer/printer.go
```diff
@@ -329,8 +330,14 @@ func (p *printer) printComponentMetadata(doc *astro.Node, source []byte) {
 			}
 		}
 		if !isClientOnlyImport {
-			p.print(fmt.Sprintf("\nimport * as $$module%v from '%s';", modCount, statement.Specifier))
+			assertions := ""
+			if statement.Assertions != "" {
+				assertions += " assert "
+				assertions += statement.Assertions
+			}
+			p.print(fmt.Sprintf("\nimport * as $$module%v from '%s'%s;", modCount, statement.Specifier, assertions))
```
jonathantneal, line 338: About this change:

- Imports are duplicated as a namespace import.
- The second import is generated from import metadata.
- To have the imports match, the assertion metadata is added.

matthewp: Oh, can you make it so the assertions are only added if there are assertions? This code needs to run in environments that don't support assertions.

jonathantneal: Yes, where `import` is used, it only adds `assert` if one exists.

matthewp (approved):

natemoo-re (approved):
LGTM! Congrats on the PR!

matthewp:
@jonathantneal happy to pair with you on getting this in.

PR #220, opened 2021-12-13
Add client hydration to component metadata

matthewp:
## Changes

- This adds a set of hydration directives to the component's metadata.
- This will be used to know which directives we need to build, for the static build.

## Testing

Tests updated.

## Docs

N/A


jonathantneal (approved):
Yea! This seems helpful.

PR #227, opened 2021-12-28
fix(#212, #214): improve expressions inside of active formatting elements

natemoo-re:
## Changes

- Closes https://github.com/withastro/compiler/issues/212
  Closes https://github.com/withastro/compiler/issues/214
- Fixes how expressions are handled inside of active formatting elements

## Testing

- Minimal tests added for open issues

## Docs

N/A, bugfix only

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -2499,8 +2499,14 @@ func frontmatterIM(p *parser) bool {
 }
 
 func inExpressionIM(p *parser) bool {
-	// p.afe (active formatting elements) should be cleared when entering an expression
-	p.clearActiveFormattingElements()
+	// p.afe (active formatting elements) should be cleared to `<a>` tags when entering an expression
+	if p.afe.contains(a.A) {
+		for {
+			if n := p.afe.pop(); len(p.afe) == 0 || n.DataAtom == a.A {
+				break
+			}
+		}
+	}
 
 	switch p.tok.Type {
 	case ErrorToken:
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -1281,10 +1281,17 @@ const items = ["Dog", "Cat", "Platipus"];
 			},
 		},
 		{
-			name:   "anchor expressions",
+			name:   "anchor inside expression",
 			source: `{true && <a>expr</a>}`,
 			want: want{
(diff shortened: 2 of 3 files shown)


internal/parser.go
```diff
@@ -2499,8 +2499,14 @@ func frontmatterIM(p *parser) bool {
 }
 
 func inExpressionIM(p *parser) bool {
-	// p.afe (active formatting elements) should be cleared when entering an expression
-	p.clearActiveFormattingElements()
```
natemoo-re, line 2503: This was a regression introduced while trying to fix expressions with `<a>` tags!

internal/parser.go
```diff
@@ -2499,8 +2499,14 @@ func frontmatterIM(p *parser) bool {
...
+	// p.afe (active formatting elements) should be cleared to `<a>` tags when entering an expression
+	if p.afe.contains(a.A) {
+		for {
+			if n := p.afe.pop(); len(p.afe) == 0 || n.DataAtom == a.A {
+				break
+			}
+		}
+	}
```
natemoo-re, line 2509: Instead of clearing _all_ active formatting elements, now we clear up to the last `<a>` tag.

internal/printer/printer_test.go
```diff
@@ -1259,6 +1259,34 @@ const items = ["Dog", "Cat", "Platipus"];
...
+				code: `<html><head></head><body><a>${expr}</a></body></html>`,
+			},
+		},
+		{
+			name:   "anchor expressions",
+			source: `{true && <a>expr</a>}`,
+			want: want{
+				code: `<html><head></head><body>${true && $$render` + BACKTICK + `<a>expr</a>` + BACKTICK + `}</body></html>`,
+			},
+		},
```
natemoo-re, line 1289: All of these are minimal cases taken from the issues.

jonathantneal (approved):
I don’t fully grok this one, but I trust this one. It’s because I don’t understand what makes an `<a>` special for expressions

natemoo-re:
@jonathantneal That's fair! This behavior is the hairiest part of the codebase IMO. I'm not totally clear on why `<a>` is special, either, but there is a long trail of bugs to confirm that this is necessary.

Would love to revisit this in the future to clarify the _why_ here.

PR #228, opened 2021-12-28
fix(#221): fix tokenization of `/` inside expression attributes

### LINKED ISSUE #221: 🐛 BUG: Can't perform maths using variables inside the JSX {} operator

0.5.7

yarn

Ubuntu

### Describe the Bug

I am facing the following error when trying to set `(height * 100) / width` as the value of any attribute.

image

And this error when used inside an element,

image

https://stackblitz.com/edit/github-hua66y?file=src%2Fpages%2Findex.astro

natemoo-re:
## Changes

- Closes #221
- Fixes tokenization of expression attributes when `/` is used (now breaks on closing `}`, if it exists)

## Testing

Tests added

## Docs

Bug fix only

### DIFF
--- a/internal/token.go
+++ b/internal/token.go
@@ -793,7 +793,7 @@ find_next:
 }
 
 // read RegExp expressions and comments (starting from '/' byte)
-func (z *Tokenizer) readCommentOrRegExp() {
+func (z *Tokenizer) readCommentOrRegExp(boundaryChars []byte) {
 	c := z.readByte() // find next character after '/' to know how to handle it
 	switch c {
 	// single-line commment (ends on newline)
@@ -813,7 +813,7 @@ func (z *Tokenizer) readCommentOrRegExp() {
 	// RegExp
 	default:
 		z.raw.End--
-		z.readUntilChar([]byte{'/', '\r', '\n'})
+		z.readUntilChar(append([]byte{'/', '\r', '\n'}, boundaryChars...))
 	}
 }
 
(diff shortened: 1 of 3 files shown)


internal/token.go
```diff
@@ -1236,7 +1236,13 @@ func (z *Tokenizer) readTagAttrExpression() {
 				if next == '/' {
 					panic("Block comments (//) are not allowed inside of expressions")
 				}
-				z.readCommentOrRegExp()
+				// Also stop when we hit a '}' character (end of attribute expression)
+				z.readCommentOrRegExp([]byte{'}'})
```
natemoo-re, line 1240: Previously, we were only doing `z.readCommentOrRegExp()` which didn't know what to do with a single `/` character (it assumed it was part of a RegExp literal.

Now `readCommentOrRegExp` accepts additional characters that can be considered a boundary, like `}`.

internal/token.go
```diff
@@ -1236,7 +1236,13 @@ func (z *Tokenizer) readTagAttrExpression() {
...
+				// If we exit on a '}', ignore the final character here
+				lastChar := z.buf[z.data.End-1 : z.data.End][0]
+				if lastChar == '}' {
+					z.data.End--
+				}
```
natemoo-re, line 1245: This checks if the final character was `}`, in which case we need to do `z.data.End--` to back up and re-process the character in this loop.

jonathantneal (approved):
Solid. Appreciate the comment above the check logic, too.

PR #229, opened 2021-12-28
fix(#224): address issue with expression tokenization

### LINKED ISSUE #224: 🐛 BUG: Astro compiler fails after first curly brace closure in an expression

0.21.13

npm

Mac

### Describe the Bug

The following Astro file gets extra `$$render` functions added to the second and third functions within the expression.

```astro
{
	() => {
		let generate = (input) => {
			let a = () => { return; };
			let b = () => { return; };
			let c = () => { return; };
		};
	}
}
```

It gets turned into something like:

```astro
{
	() => {
		let generate = (input) => {
			let a = () => { return; };
			let b = () => $$render`${ return; }`;
			let c = () => $$render`${ return; }`;
		};
	}
}
```

Credit to @okikio for discovering the issue. 🙏

https://stackblitz.com/edit/github-gjpgyq

natemoo-re:
## Changes

- Closes #224
- Fixes logical error in expression tokenization step. It worked in most situations but #224 was a reproducible edge case where it failed.

## Testing

Added tokenizer and printer tests to verify that the case in the issue works. Existing tests ensure this is not a regression.

## Docs

N/A, bug fix only.

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -410,6 +410,34 @@ $$render` + "`" + `<div${$$addAttribute(color, "id")}>color</div>` + "`" + `
 			},
 		},
 		{
+			name: "expressions with multiple curly braces",
+			source: `
+<div>
+{
+	() => {
+		let generate = (input) => {
+			let a = () => { return; };
+			let b = () => { return; };
+			let c = () => { return; };
+		};
+	}
+}
+</div>`,
+			want: want{
+				code: `<html><head></head><body><div>
+${
+	() => {
+		let generate = (input) => {
+			let a = () => { return; };
+			let b = () => { return; };
+			let c = () => { return; };
+		};
+	}
+}
+</div></body></html>`,
+			},
+		},
+		{
 			name: "slots (basic)",
 			source: `---
 import Component from "test";
--- a/internal/token.go
+++ b/internal/token.go
@@ -1674,9 +1674,9 @@ expression_loop:
 				z.tt = TextToken
 				return z.tt
 			}
-			z.openBraceIsExpressionStart = true
 			z.expressionStack[len(z.expressionStack)-1]--
 			if z.expressionStack[len(z.expressionStack)-1] == -1 {
+				z.openBraceIsExpressionStart = true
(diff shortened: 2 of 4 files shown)


internal/token.go
```diff
@@ -1674,9 +1674,9 @@ expression_loop:
 				z.tt = TextToken
 				return z.tt
 			}
-			z.openBraceIsExpressionStart = true
 			z.expressionStack[len(z.expressionStack)-1]--
 			if z.expressionStack[len(z.expressionStack)-1] == -1 {
+				z.openBraceIsExpressionStart = true
```
natemoo-re, line 1679: Pretty simple fix! This edge case exposed that `z.openBraceIsExpressionStart = true` should only be reset when we've exhausted the entire `z.expressionStack`

internal/token_test.go
```diff
@@ -122,6 +122,17 @@ func TestBasic(t *testing.T) {
...
+		{
+			"expression with multiple returns",
+			`<div>{() => {
+	let generate = (input) => {
+		let a = () => { return; };
+		let b = () => { return; };
+		let c = () => { return; };
+	};
+}}</div>`,
+			[]TokenType{StartTagToken, StartExpressionToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, TextToken, EndExpressionToken, EndTagToken},
```
natemoo-re, line 134: This was previously being tokenized incorrectly.

internal/printer/printer_test.go
```diff
@@ -406,6 +406,34 @@ $$render` + "`" + `<div${$$addAttribute(color, "id")}>color</div>` + "`" + `
...
+}
+</div>`,
+			want: want{
+				code: `<html><head></head><body><div>
+${
+	() => {
+		let generate = (input) => {
+			let a = () => { return; };
+			let b = () => { return; };
+			let c = () => { return; };
```
natemoo-re, line 433: Added a printer test as a sanity check, too.

jonathantneal (approved):
Thank you! The fix was so small that I had to look twice!

PR #230, opened 2021-12-28
fix(#225): Do not treat `noscript` as raw

### LINKED ISSUE #225: 🐛 BUG: Components inside <noscript> are not parsed through the compiler

0.21.13

npm

Mac

### Describe the Bug

Any components in a `<noscript>` tag only pass in the source code. It does not get parsed through the compiler.

https://stackblitz.com/edit/github-8nksfd?file=src/pages/index.astro

natemoo-re:
## Changes

- Closes #225
- Removes `noscript` from the list of tags to tokenize as raw text

## Testing

Tests added in tokenizer and printer

## Docs

Bug fix only

### DIFF
--- a/internal/token.go
+++ b/internal/token.go
@@ -957,7 +957,7 @@ func (z *Tokenizer) readStartTag() TokenType {
 	case 'i':
 		raw = z.startTagIn("iframe")
 	case 'n':
-		raw = z.startTagIn("noembed", "noframes", "noscript")
+		raw = z.startTagIn("noembed", "noframes")
 	case 'p':
 		raw = z.startTagIn("plaintext")
 	case 's':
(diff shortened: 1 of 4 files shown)


internal/token_test.go
```diff
@@ -44,6 +44,11 @@ func TestBasic(t *testing.T) {
 			`<pkg.Item>`,
 			[]TokenType{StartTagToken},
 		},
+		{
+			"noscript component",
+			`<noscript><Component /></noscript>`,
+			[]TokenType{StartTagToken, SelfClosingTagToken, EndTagToken},
```
natemoo-re, line 50: Previously all contents inside of `noscript` were tokenized as `TextToken` instead of being tokenized as regular content.

internal/token.go
```diff
@@ -957,7 +957,7 @@ func (z *Tokenizer) readStartTag() TokenType {
 	case 'i':
 		raw = z.startTagIn("iframe")
 	case 'n':
-		raw = z.startTagIn("noembed", "noframes", "noscript")
+		raw = z.startTagIn("noembed", "noframes")
```
natemoo-re, line 960: Removing `noscript` from this `raw` logic fixes that!

jonathantneal (approved):
This makes sense. Server parsing versus client parsing. 👍

PR #231, opened 2021-12-28
fix(#213): @keyframes breaks following scoping

### LINKED ISSUE #213: 🐛 BUG: CSS @keyframes rules prevent scoping for all rules that follow

0.21.10

yarn

Mac

## Summary

Using `@keyframes` in a `<style>` block in an astro component breaks selector scoping for all rules after the `@keyframes` rule.

This behavior changed between 0.20 and 0.21.

### Example

Is this example, the `.my-para` declaration will be scoped properly. The `.another-para` will not be scoped.

```astro
<style>
  .my-para {
    color: blue;
  }

  @keyframes fadeIn {
    to {
      opacity: 100%;
    }
  }

  .another-para {
    color: red;
  }
</style>
```

That input CSS produces the following bundled CSS. **Note**: I've modified formatting here for readability

```css
.my-para.astro-FNZOVGOV {
  color: #00f;
}
@keyframes fadeIn {
  to {
    opacity: 100%;
  }
}
.another-para {
  color: red;
}
```

The expecation is that `.another-para` would be scoped `.another-para.astro-FNZOVGOV`. This was was the behavior in Astro pre 0.21.

## Further testing

From what I've been able to tell, the `@keyframes` rule breaks scoping for all rules that come after it. So, if we modify our above example to put `@keyframes` first, no styles are scoped

```astro
<style>
  @keyframes fadeIn {
    to {
      opacity: 100%;
    }
  }

  .my-para {
    color: blue;
  }

  .another-para {
    color: red;
  }
</style>
```

produces the following (again, formatting modified here for readability)

```css
@keyframes fadeIn {
  to {
    opacity: 100%;
  }
}
.my-para {
  color: #00f;
}
.another-para {
  color: red;
}
```

## Working around this

This is a bit flimsy, but you can work around this by putting all `@keyframes` rules at the bottom of any `<style>`.

```astro
<style>
  .my-para {
    color: blue;
  }

  .another-para {
    color: red;
  }

  @keyframes fadeIn {
    to {
      opacity: 100%;
    }
  }
</style>
```

Produces:

```css
.my-para.astro-FNZOVGOV {
  color: #00f;
}
.another-para.astro-FNZOVGOV {
  color: red;
}
@keyframes fadeIn {
  to {
    opacity: 100%;
  }
}
```

## Other @-rules

This seems to be isolated to `@keyframes`. I tested with `@supports` and `@font-face`. Neither of those caused this same, or other, issues.

My hunch is that the extra curly braces needed in `@keyframes` declarations trips things up.

https://codesandbox.io/s/clever-jennings-xlp4z?file=/src/pages/index.astro

natemoo-re:
## Changes

- Fixes #213
- Previously, any rules following `@keyframes` would not be scoped. This fixes that!

## Testing

Multiple tests add

## Docs

Bug fix only

### DIFF
--- a/internal/transform/scope-css.go
+++ b/internal/transform/scope-css.go
@@ -85,6 +85,17 @@ outer:
 
 					// if inside @keyframes, don’t transform what’s there
 					if isKeyframes {
+						if strVal == "{" {
+							keyframeCurlyCount++
+						} else if strVal == "}" {
+							keyframeCurlyCount--
+						}
+
+						// Inside of this case, we only want to break out when keyframeCurlyCount is -1
+						// since 0 is the default
+						if keyframeCurlyCount < 0 {
+							isKeyframes = false
+						}
 						out += strVal
 						continue
 					}
@@ -131,12 +142,16 @@ outer:
 						isBracket = false
 						out += strVal
 					case "{":
-						keyframeCurlyCount++
+						if isKeyframes {
+							keyframeCurlyCount++
+						}
 						isElement = true
 						isPseudoState = false
 						out += strVal
 					case "}":
-						keyframeCurlyCount--
+						if isKeyframes {
+							keyframeCurlyCount--
+						}
 						if keyframeCurlyCount == 0 {
 							isKeyframes = false
 						}
--- a/internal/transform/scope-css_test.go
+++ b/internal/transform/scope-css_test.go
@@ -189,6 +189,21 @@ func TestScopeStyle(t *testing.T) {
 			want:   "@keyframes shuffle{0%{transform:rotate(0deg);color:blue;}100%{transform:rotate(360deg};}}",
 		},
 		{
+			name:   "keyframes start",
+			source: "@keyframes shuffle{0%{transform:rotate(0deg);color:blue;}100%{transform:rotate(360deg};}} h1{} h2{}",
+			want:   "@keyframes shuffle{0%{transform:rotate(0deg);color:blue;}100%{transform:rotate(360deg};}}h1.astro-XXXXXX{}h2.astro-XXXXXX{}",
+		},
+		{
+			name:   "keyframes middle",
+			source: "h1{} @keyframes shuffle{0%{transform:rotate(0deg);color:blue;}100%{transform:rotate(360deg};}} h2{}",
+			want:   "h1.astro-XXXXXX{}@keyframes shuffle{0%{transform:rotate(0deg);color:blue;}100%{transform:rotate(360deg};}}h2.astro-XXXXXX{}",
+		},
+		{
+			name:   "keyframes end",
+			source: "h1{} h2{} @keyframes shuffle{0%{transform:rotate(0deg);color:blue;}100%{transform:rotate(360deg};}}",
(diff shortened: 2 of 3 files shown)


internal/transform/scope-css.go
```diff
@@ -85,6 +85,17 @@ outer:
 
 					// if inside @keyframes, don’t transform what’s there
 					if isKeyframes {
+						if strVal == "{" {
+							keyframeCurlyCount++
+						} else if strVal == "}" {
+							keyframeCurlyCount--
+						}
```
natemoo-re, line 92: Previously, we were only doing `out += strVal` inside of this case, but that meant we never updated `keyframeCurlyCount` to reflect when we were done!

internal/transform/scope-css.go
```diff
@@ -85,6 +85,17 @@ outer:
...
+
+						// Inside of this case, we only want to break out when keyframeCurlyCount is -1
+						// since 0 is the default
+						if keyframeCurlyCount < 0 {
+							isKeyframes = false
+						}
```
natemoo-re, line 98: `keyframeCurlyCount == 0` is the default case, we are done when `keyframeCurlyCount == -1`.

internal/transform/scope-css.go
```diff
@@ -131,12 +142,16 @@ outer:
 						isBracket = false
 						out += strVal
 					case "{":
-						keyframeCurlyCount++
+						if isKeyframes {
+							keyframeCurlyCount++
+						}
```
natemoo-re, line 147: Only track this when we're inside `isKeyframes`, otherwise we might be starting at a weird number.

internal/transform/scope-css.go
```diff
@@ -131,12 +142,16 @@ outer:
...
 						isElement = true
 						isPseudoState = false
 						out += strVal
 					case "}":
-						keyframeCurlyCount--
+						if isKeyframes {
+							keyframeCurlyCount--
+						}
```
natemoo-re, line 154: Only track this when we're inside `isKeyframes`, otherwise we might be starting at a weird number.

internal/transform/scope-css_test.go
```diff
@@ -188,6 +188,21 @@ func TestScopeStyle(t *testing.T) {
...
 		},
+		{
+			name:   "keyframes start",
+			source: "@keyframes shuffle{0%{transform:rotate(0deg);color:blue;}100%{transform:rotate(360deg};}} h1{} h2{}",
+			want:   "@keyframes shuffle{0%{transform:rotate(0deg);color:blue;}100%{transform:rotate(360deg};}}h1.astro-XXXXXX{}h2.astro-XXXXXX{}",
+		},
+		{
+			name:   "keyframes middle",
+			source: "h1{} @keyframes shuffle{0%{transform:rotate(0deg);color:blue;}100%{transform:rotate(360deg};}} h2{}",
+			want:   "h1.astro-XXXXXX{}@keyframes shuffle{0%{transform:rotate(0deg);color:blue;}100%{transform:rotate(360deg};}}h2.astro-XXXXXX{}",
```
natemoo-re, line 199: It works!

jonathantneal (approved):
It works! Thank you.

PR #232, opened 2021-12-28
fix(#223): Do not prefix values inside of declarations

### LINKED ISSUE #223: 🐛 BUG: grid-template-* with [line-name] adds extra astro class

^0.21.10

npm

Mac

### Describe the Bug

When adding `grid-template-columns` in a style tag with line-names, like this:

`grid-template-columns: [content-start] 1fr [content-end];`

an extra .astro class is entered into the style value, generating this:

`grid-template-columns: .astro-4D2SJCT3[content-start] 1fr [content-end]`

which is invalid CSS.

https://stackblitz.com/edit/github-rfg3sg?file=src/components/Tour.astro

natemoo-re:
## Changes

- Fixes #223
- Improves our scoping algorithm to not touch values inside of declaration rules (like `grid-template-column`.

## Testing

Test added

## Docs

Bug fix only

### DIFF
--- a/internal/transform/scope-css.go
+++ b/internal/transform/scope-css.go
@@ -35,6 +35,7 @@ outer:
 
 		isKeyframes := false    // if we’re inside @keyframes, there’s nothing to scope
 		keyframeCurlyCount := 0 // keep track of open "{"s inside @keyframes
+		declaration := ""
 
 	walk:
 		for {
@@ -69,6 +70,7 @@ outer:
 					}
 				case css.DeclarationGrammar:
 					out += string(data) + ":"
+					declaration = string(data)
 				default:
 				}
 
@@ -122,8 +124,9 @@ outer:
 						isBracket = true
 						isElement = false
 						isPseudoState = false
-						// if there is no selector before an attribute selector, then assume "*"
-						if n == 0 {
+
+						// if there is no selector before an attribute selector and we're not in a delcaration, assume "*"
+						if n == 0 && declaration == "" {
 							out += scopeRule("", opts)
 						}
 						out += strVal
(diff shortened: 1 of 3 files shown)


internal/transform/scope-css.go
```diff
@@ -35,6 +35,7 @@ outer:
 
 		isKeyframes := false    // if we’re inside @keyframes, there’s nothing to scope
 		keyframeCurlyCount := 0 // keep track of open "{"s inside @keyframes
+		declaration := ""
```
natemoo-re, line 38: This tracks any `css.DeclarationGramar` nodes that we enter.

internal/transform/scope-css.go
```diff
@@ -69,6 +70,7 @@ outer:
 					}
 				case css.DeclarationGrammar:
 					out += string(data) + ":"
+					declaration = string(data)
```
natemoo-re, line 73: Set the `declaration` flag. Just used as a boolean flag.

internal/transform/scope-css.go
```diff
@@ -122,8 +124,9 @@ outer:
 						isBracket = true
 						isElement = false
 						isPseudoState = false
-						// if there is no selector before an attribute selector, then assume "*"
-						if n == 0 {
+
+						// if there is no selector before an attribute selector and we're not in a delcaration, assume "*"
+						if n == 0 && declaration == "" {
```
natemoo-re, line 129: Previously, this logic was designed to handle attribute selectors `[data-click]` but it _also_ unintentionally caught cases like:
```
grid-template-columns: [content-start] 1fr [content-end];
```

This fixes that by checking if we've already encountered a declaration.

jonathantneal (approved):
I’m so glad the fix for this is logical (with the "i am within a declaration value, do not touch" flag)

PR #243, opened 2022-01-13
Pass opts.Pathname to top-level createAstro

matthewp:
## Changes

- Changes createAstro from `createAstro(import.meta.url)` to `createAstro("/src/components/Foo.astro")`.
  - This aligns with what createMetadata receives.
  - This is needed because in the static build (and eventually SSR) `import.meta.url` points to the build JS files and not the original source location. Since we want to resolve browser URLs, passing the pathname string makes more sense.
## Testing

No tests needed updating

## Docs

N/A

### DIFF
--- a/.changeset/ninety-yaks-hide.md
+++ b/.changeset/ninety-yaks-hide.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': minor
+---
+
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):
LGTM!

Issue #242, opened 2022-01-07
🐛 BUG: Can't nest Astro components with `<slot/>`


tony-sull:
### What version of `@astrojs/compiler` are you using?

0.7.4

### What package manager are you using?

npm

### What operating system are you using?

Linux

### Describe the Bug

I am trying to use one Astro component as a child to another Astro component using slots.

I'm able to use plain old HTML in the parent component's slot, but when using an Astro component for the slot the child component is never rendered and is being thrown into the slot as a raw string

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-3bqujt-lpwy6r?file=src%2Fpages%2Findex.astro

tony-sull:
I assumed the slots for Astro components were handled in the compiler, looks like that's actually in the main `astro` project!  Filed [2346](https://github.com/withastro/astro/issues/2346) to track there since this shouldn't require a compiler fix

tony-sull:
May have been a little quick to assume this isn't the compiler.  I can't quite tell where Astro ends and the compiler begins, going to re-open this for now just in case the fix actually is here

One note on the bug, I realized that it only reproduces if the parent component is using props.  I.e. in the stackblitz repro, updating `FeaturesBlock.astro` so that it doesn't need to pass the `data` prop to `UIBlock` renders the slotted content fine

PR #245, opened 2022-01-13
Fix data-astro-raw logic

natemoo-re:
## Changes

- Closes https://github.com/withastro/compiler/issues/242 and https://github.com/withastro/astro/issues/2346
- Fixes logic error in `data-astro-raw` detection

## Testing

Multiple tests added

## Docs

Bug fix only

### DIFF
--- a/internal/token.go
+++ b/internal/token.go
@@ -931,18 +931,12 @@ loop:
 }
 
 func (z *Tokenizer) hasTag(s string) bool {
-loop:
 	for i := len(z.attr) - 1; i >= 0; i-- {
-
 		x := z.attr[i]
 		key := z.buf[x[0].Start:x[0].End]
-		for i := 0; i < len(key) && i < len(s); i++ {
-			c := key[i]
-			if c != s[i] {
-				continue loop
-			}
+		if string(key) == s {
+			return true
 		}
-		return true
 	}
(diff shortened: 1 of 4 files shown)


internal/token.go
```diff
@@ -931,18 +931,12 @@ loop:
...
-			c := key[i]
-			if c != s[i] {
-				continue loop
-			}
+		if string(key) == s {
+			return true
 		}
-		return true
 	}
 	return false
```
natemoo-re, line 941: This was the fix! `data-astro-raw` was previously matching any slice (from 0 to len) of `data-astro-raw`, so `data` would incorrectly be matched.

The fix was to simplify the loop logic and just do simple string comparison.

matthewp: oops

matthewp:
Great catch! lgtm

internal/token.go
Morritz: > This was the fix! `data-astro-raw` was previously matching any slice (from 0 to len) of `data-astro-raw`, so `data` would incorrectly be matched.
> 
> The fix was to simplify the loop logic and just do simple string comparison.
I didnt expect it to be nested this deep, I didnt even knew that the golang was used for the compiler

PR #247, opened 2022-01-18
Add hoisted scripts to TransformResult

matthewp:
## Changes

- This adds hoisted scripts to the `TransformResult`. In the static build these will work like styles, where in dev they are all external. For example `/src/components/Counter.astro?astro&type=script&index=0` will be one path.
- A change on the Astro side will bundle these in production (static build only).

## Testing

No added tests needed, only a change to the wasm API.

## Docs

N/A

### DIFF
--- a/.changeset/silly-laws-smoke.md
+++ b/.changeset/silly-laws-smoke.md
@@ -2,4 +2,4 @@
 '@astrojs/compiler': patch
 ---
 
-Adds a new scripts property to the TransformResult
+Adds a new property, `scripts`, to `TransformResult`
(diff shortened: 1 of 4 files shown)


.changeset/silly-laws-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Adds a new scripts property to the TransformResult
```
boehs, line 5:
```suggestion
Adds a new property, `scripts`, to `TransformResult`
```

boehs (changes requested):
Code looks high quality, have not tested locally however. Minor spelling :P

natemoo-re (approved):
Looks great!

PR #253, opened 2022-01-19
Better hoisted script types

matthewp:
## Changes

- Small improvement for the types on hoisted scripts in the TransformResult. Makes it so that if you know the `type` then the corresponding other prop is not possibly undefined.

## Testing

N/A

## Docs

N/A


jonathantneal (approved):
I love anything that avoids me seeing `undefined`. This feels like some kind of `Partial` trick.

PR #258, opened 2022-01-19
Revert "Add hoisted scripts to TransformResult (#247)"

matthewp:
This reverts commit 502f8b8cf13ce775cd196ee3b9e37003f6c3c053.

I believe this is causing astro to fail on Windows but have not tracked down why.


natemoo-re (approved):
LGTM!

PR #269, opened 2022-01-27
Implement `set:html` and `set:text` 

natemoo-re:
## Changes

- Implements the [`set:html` RFC](https://github.com/withastro/rfcs/blob/main/active-rfcs/0000-set-html.md)
- Instead of complicating the `astro/internal` logic, I found a way to implement these directives entirely in the compiler. `set:html={content}` is converted to a child expression. `set:text={content}` is converted to a child expression wrapped in `$$escapeHTML(content)`.
- Will require `escapeHTML` being added to `astro/internal`, [PR in core here](https://github.com/withastro/astro/pull/2479)

## Testing

Lots of tests added

## Docs

No docs yet, will need to add those.


internal/transform/transform.go
```diff
@@ -106,6 +111,59 @@ func ExtractStyles(doc *astro.Node) {
...
+		if n.Type == astro.ElementNode && HasSetDirective(n) {
+			for _, attr := range n.Attr {
+				if attr.Key == "set:html" || attr.Key == "set:text" {
+					nodes = append(nodes, n)
+					directives = append(directives, &attr)
+					return
+				}
+			}
+		}
+	})
```
natemoo-re, line 127: Find `set:html` or `set:text` attributes

internal/transform/transform.go
```diff
@@ -106,6 +111,59 @@ func ExtractStyles(doc *astro.Node) {
...
+
+	if len(nodes) > 0 {
+		for i, n := range nodes {
+			directive := directives[i]
+			n.RemoveAttribute(directive.Key)
```
natemoo-re, line 132: Remove the attribute so it isn't printed

internal/transform/transform.go
```diff
@@ -106,6 +111,59 @@ func ExtractStyles(doc *astro.Node) {
...
+			loc = append(loc, directive.ValLoc)
+			data := directive.Val
+			if directive.Key == "set:text" {
+				data = fmt.Sprintf("$$escapeHTML(%s)", data)
+			}
+			expr.AppendChild(&astro.Node{
+				Type: astro.TextNode,
+				Data: data,
+				Loc:  loc,
+			})
```
natemoo-re, line 149: The directive attribute is converted to a child expression node

jonathantneal (approved):
This is so clean!

Am I right there’s a second stage to this from the `astro` repo to implement `escapeHTML`?

internal/transform/transform.go
```diff
@@ -106,6 +111,59 @@ func ExtractStyles(doc *astro.Node) {
...
+
+			shouldWarn := false
+			// Remove all existing children
+			for c := n.FirstChild; c != nil; c = c.NextSibling {
+				if !shouldWarn {
+					shouldWarn = c.Type == astro.CommentNode || (c.Type == astro.TextNode && len(strings.TrimSpace(c.Data)) != 0)
+				}
+				n.RemoveChild(c)
+			}
```
natemoo-re, line 158: If there are existing children, we remove them. We only throw a warning if the children are element nodes or non-empty text nodes.

internal/transform/transform.go
```diff
@@ -118,6 +176,9 @@ func ExtractStyles(doc *astro.Node) {
 
 func ExtractScript(doc *astro.Node, n *astro.Node) {
 	if n.Type == astro.ElementNode && n.DataAtom == a.Script {
+		if HasSetDirective(n) {
+			return
+		}
```
natemoo-re, line 181: Per the RFC, the existence of `set:html` or `set:text` means the `<script>` will be printed literally

internal/transform/transform.go
```diff
@@ -92,6 +94,9 @@ func Transform(doc *astro.Node, opts TransformOptions) *astro.Node {
 func ExtractStyles(doc *astro.Node) {
 	walk(doc, func(n *astro.Node) {
 		if n.Type == astro.ElementNode && n.DataAtom == a.Style {
+			if HasSetDirective(n) {
+				return
+			}
```
natemoo-re, line 99: Same for `<style>`, these will be literally printed on the page.

internal/printer/printer.go
```diff
@@ -221,7 +226,7 @@ func (p *printer) printStyleOrScript(n *astro.Node) {
 }
 
 func (p *printer) printAttribute(attr astro.Attribute) {
-	if attr.Key == "define:vars" {
+	if attr.Key == "define:vars" || attr.Key == "set:text" || attr.Key == "set:html" {
 		return
```
natemoo-re, line 230: If one of these slips through (possible if there are duplicates or if `set:html` and `set:text` are mixed), we ignore them and do not print them as attributes.

internal/printer/printer.go
```diff
@@ -60,6 +61,7 @@ func (p *printer) printInternalImports(importSpecifier string) {
 	p.print("createAstro as " + CREATE_ASTRO + ",\n  ")
 	p.print("createComponent as " + CREATE_COMPONENT + ",\n  ")
 	p.print("renderComponent as " + RENDER_COMPONENT + ",\n  ")
+	p.print("escapeHTML as " + ESCAPE_HTML + ",\n  ")
```
natemoo-re, line 64: Need to add `escapeHTML` function to `astro/internal`. PR coming shortly.

natemoo-re:
> Am I right there’s a second stage to this from the astro repo to implement escapeHTML?

That's correct @jonathantneal! I'll link that when it's up.

boehs (approved):
Should this be marked as a draft till then?

matthewp:
Looks great!

PR #270, opened 2022-01-27
Implement `Astro.self`

natemoo-re:
## Changes

- Implements the [Recursive Components RFC](https://github.com/withastro/rfcs/blob/main/active-rfcs/0000-recursive-components.md)
- Also implements better naming scheme for components, where the name will be generated from the file pathname if possible.

## Testing

WASM test added, `Astro.self` added to test output.

## Docs

Needs docs.


jonathantneal (approved):
I’m always confused by `opts.opts` but this looks good. L32 of printer_test.go tells me what I expect. :)

natemoo-re:
Totally agree, @jonathantneal! Would be great to clean that up.

boehs (approved):
From a code quality standpoint this is spot on!

matthewp (approved):
lgtm

PR #274, opened 2022-01-31
Add `parse` function, emit AST

natemoo-re:
## Changes

- Adds a function named `parse` which can emit an AST
- Adds `@astrojs/compiler/utils` with a `walk` function and `is` type guards.
- Adds fully typed AST Nodes
**Example**

```js
import { parse } from '@astrojs/compiler';
import { walk, is } from '@astrojs/compiler/utils';

const result = await parse(source, {
  position: false, // defaults to `true`
});

walk(result.ast, (node) => {
  // `tag` nodes are `element` | `custom-element` | `component`
  if (is.tag(node)) {
    console.log(node.name);
  }
})
```

## Testing

- Minimal Go test scaffolded out
- Simple JS test added

## Docs

README has been updated. Docs should probably remain internal for a while until this is stabilized.


lib/compiler/README.md
```diff
@@ -12,7 +12,15 @@ npm install @astrojs/compiler
 
 ## Usage
 
-_Note: API will change before 1.0! Use at your own discretion._
+_Note: Public APIs are likely to change before 1.0! Use at your own discretion._
+
+#### Transform `.astro` to valid TypeScript
+
+The Astro compiler can convert `.astro` syntax to a JavaScript Module whose default export generates HTML.
```
jonathantneal, line 19:
```suggestion
The Astro compiler can convert `.astro` syntax to a TypeScript Module whose default export generates HTML.
```

Per the note below that reads "_The output code may need an additional post-processing step to generate valid **JavaScript**._"

jonathantneal (approved):
I can’t possibly fathom it all without a walk thru, but this looks phenomenal.

cmd/astro-wasm/astro-wasm.go
```diff
@@ -133,6 +152,36 @@ func preprocessStyle(i int, style *astro.Node, transformOptions transform.Transf
...
+		}
+		doc = &astro.Node{
+			Type: astro.DocumentNode,
+		}
+		for i := 0; i < len(nodes); i++ {
+			n := nodes[i]
+			doc.AppendChild(n)
+		}
+
+		result := printer.PrintToJSON(source, doc, parseOptions)
```
natemoo-re, line 177: This is very similar to the `Transform` function other than the way we print. Instead of generating a JS module from our AST, we have a new printer that generates stringified JSON.

internal/printer/print-to-json.go
```diff
@@ -0,0 +1,241 @@
...
+		str += `,"position":{`
+		str += fmt.Sprintf(`"start":{"line":%d,"column":%d,"offset":%d}`, n.Position.Start.Line, n.Position.Start.Column, n.Position.Start.Offset)
+		if n.Position.End.Line != 0 {
+			str += fmt.Sprintf(`,"end":{"line":%d,"column":%d,"offset":%d}`, n.Position.End.Line, n.Position.End.Column, n.Position.End.Offset)
+		}
+		str += "}"
+	}
+	str += "}"
+	return str
+}
```
natemoo-re, line 106: Just want to note that our `String()` function is implemented manually, but recurses through the AST just fine.

I had originally used `json.Marshal` but it relies heavily on Reflection which is not well supported in `tinygo`. This is more performant, anyway.

internal/printer/print-to-json.go
```diff
@@ -0,0 +1,241 @@
...
+		if hasChildren {
+			for c := n.FirstChild; c != nil; c = c.NextSibling {
+				renderNode(p, parent, c, opts)
+			}
+		}
+		return
+	}
+	var node ASTNode
+
+	node.Position = positionAt(p, n, opts)
```
natemoo-re, line 193: Position data is not _super_ robust at the moment and might be wrong in some places. This should work but we'll need a cleanup pass to normalize this data in the parser.

lib/compiler/shared/ast.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export interface ParentLikeNode extends BaseNode {
+  type: 'element' | 'component' | 'custom-element' | 'expression' | 'root';
+  children: Node[];
+}
+
+export interface LiteralNode extends BaseNode {
+  type: 'text' | 'doctype' | 'comment' | 'frontmatter';
+  value: string;
+}
```
natemoo-re, line 29: This is a pretty good approximation of our AST! Intentionally very close to [`rehype`](https://github.com/rehypejs/rehype) which is built on [`unist`](https://github.com/syntax-tree/unist).

lib/compiler/shared/ast.ts
```diff
@@ -0,0 +1,85 @@
...
+export interface AttributeNode extends BaseNode {
+  type: 'attribute';
+  kind: 'quoted' | 'empty' | 'expression' | 'spread' | 'shorthand' | 'template-literal';
+  name: string;
+  value: string;
+}
+
+export interface DirectiveNode extends Omit<AttributeNode, 'type'> {
+  type: 'directive';
+}
```
natemoo-re, line 44: One choice worth calling out—I've split `attributes` and `directives` into two separate nodes. This should make it easy to find and validate official directive without needing to loop through all the existing attributes. The compiler filters them out for you.

natemoo-re (commented):
Leaving some comments for reference

matthewp (approved):
lgtm

PR #275, opened 2022-01-31
Support define:vars with static extraction

matthewp:
## Changes

- If using `<style define:vars>`, those get added to the STYLES array.

## Testing

Added tests for styles and scripts when using StaticExtraction flag.

## Docs

Bug fix only


natemoo-re (approved):
LGTM!

PR #282, opened 2022-02-14
feat: `data-astro-raw` => `is:raw`

natemoo-re:
## Changes

- Was investigating an issue with `data-astro-raw` that turned out to not be a problem with the compiler! Woohoo!
- This formalizes support for `data-astro-raw` attribute as the `is:raw` directive.
- `is:raw` attribute will not be printed in the output, it is just a compiler directive.

## Testing

Tests added

## Docs

TBD.

### DIFF
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -172,7 +172,7 @@ func (p *printer) printAttributesToObject(n *astro.Node) {
 		if i != 0 {
 			p.print(",")
 		}
-		if a.Key == "set:text" || a.Key == "set:html" {
+		if a.Key == "set:text" || a.Key == "set:html" || a.Key == "is:raw" {
 			continue
 		}
 		switch a.Type {
@@ -249,7 +249,7 @@ func (p *printer) printStyleOrScript(opts RenderOptions, n *astro.Node) {
 }
 
 func (p *printer) printAttribute(attr astro.Attribute) {
-	if attr.Key == "define:vars" || attr.Key == "set:text" || attr.Key == "set:html" {
+	if attr.Key == "define:vars" || attr.Key == "set:text" || attr.Key == "set:html" || attr.Key == "is:raw" {
 		return
 	}
 
--- a/internal/token.go
+++ b/internal/token.go
@@ -930,7 +930,7 @@ loop:
 	return false
 }
 
-func (z *Tokenizer) hasTag(s string) bool {
+func (z *Tokenizer) hasAttribute(s string) bool {
 	for i := len(z.attr) - 1; i >= 0; i-- {
 		x := z.attr[i]
 		key := z.buf[x[0].Start:x[0].End]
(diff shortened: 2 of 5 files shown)


internal/printer/printer.go
```diff
@@ -172,7 +172,7 @@ func (p *printer) printAttributesToObject(n *astro.Node) {
 		if i != 0 {
 			p.print(",")
 		}
-		if a.Key == "set:text" || a.Key == "set:html" {
+		if a.Key == "set:text" || a.Key == "set:html" || a.Key == "is:raw" {
```
natemoo-re, line 175: Don't print these in the output.

internal/printer/printer.go
```diff
@@ -249,7 +249,7 @@ func (p *printer) printStyleOrScript(opts RenderOptions, n *astro.Node) {
 }
 
 func (p *printer) printAttribute(attr astro.Attribute) {
-	if attr.Key == "define:vars" || attr.Key == "set:text" || attr.Key == "set:html" {
+	if attr.Key == "define:vars" || attr.Key == "set:text" || attr.Key == "set:html" || attr.Key == "is:raw" {
```
natemoo-re, line 252: Don't print these in the output.

internal/token.go
```diff
@@ -930,7 +930,7 @@ loop:
 	return false
 }
 
-func (z *Tokenizer) hasTag(s string) bool {
+func (z *Tokenizer) hasAttribute(s string) bool {
```
natemoo-re, line 933: Renamed `hasTag` to `hasAttribute` as it was a confusing name!

internal/token.go
```diff
@@ -962,7 +962,10 @@ func (z *Tokenizer) readStartTag() TokenType {
 		raw = z.startTagIn("xmp")
 	}
 	if !raw {
-		raw = z.hasTag("data-astro-raw")
+		raw = z.hasAttribute("data-astro-raw")
+	}
```
natemoo-re, line 966: Still supports `data-astro-raw` for now!

internal/token.go
```diff
@@ -962,7 +962,10 @@ func (z *Tokenizer) readStartTag() TokenType {
...
+	if !raw {
+		raw = z.hasAttribute("is:raw")
```
natemoo-re, line 968: Add support for `is:raw` directive

internal/token.go
boehs: Should there be a deprecation warning?

natemoo-re: Probably but would be more helpful to handle that part in `core`

boehs (approved):

matthewp (approved):

natemoo-re:
Going to bring this up on the RFC call, started a discussion https://github.com/withastro/rfcs/discussions/114 with more detail in the interest of transparency.

I think this is a safe change since it was an undocumented and private API, but it never hurts to get more input.

jonathantneal (approved):
I believe this name is subject to change, but let’s experiment. :)

Issue #277, opened 2022-01-28
🐛 BUG: memory access out of bounds with KaTeX


JuanM04:
### What version of `astro` are you using?

0.22.19

### What package manager are you using?

npm, yarn

### What operating system are you using?

Linux

### Describe the Bug

When using markdown and installing `katex`, `remark-math` and `rehype-katex`, some math equations throws a wasm error: `memory access out of bounds`. Specifically, some functions like `\frac{1}{2}`, or `\sqrt[3]{27}` (interestingly, `\sqrt{4}` doesn't throw)

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-vrzdws?file=src/pages/index.md

JuanM04:
Thanks! I was just narrowing down where the bug happens until I've reached this minimal example: https://stackblitz.com/edit/node-ldbkbx?file=index.js

JuanM04:
Thanks! I was just narrowing down where the bug happens until I've reached this minimal example: https://stackblitz.com/edit/node-ldbkbx?file=index.js

natemoo-re:
Thanks for the amazing reproductions! I'm taking a look at this today ☺️

JuanM04:
After seeing your PR I've made a workaround to the original plugin:

```js
import { visit } from "unist-util-visit"
import rehypeKatex from "rehype-katex"

export default () => {
  return (tree, ...args) => {
    rehypeKatex({
      macros: {
        "\\unit": "\\ \\text{#1}",
      },
    })(tree, ...args)

    visit(tree, "element", (node) => {
      const className = node.properties?.className || []

      if (className.includes("math")) {
        node.properties["data-astro-raw"] = true
      }
    })
  }
}

```

JuanM04:
@natemoo-re The error persists after upgrading to 0.10.2. I've updated the reproduction

PR #283, opened 2022-02-15
Treat `{}` as raw text inside of `<math>`

natemoo-re:
## Changes

- Closes https://github.com/withastro/compiler/issues/277.
- Improves Astro's `MathML` support by treating `{}` characters as raw text. 
- Note that this introduces a new tokenization mode where _only expressions are disabled_. All children are still tokenized properly, but `{` is a `TextToken` rather than `StartExpressionToken`. This means that Astro compiler directives like `set:html` will continue to work inside of `<math>`.
- There are likely few users impacted by this change directly, but more users who use something like [KaTeX](https://katex.org/) (where `{}` represent significant characters) that will benefit from this indirectly.

## Testing

Tokenizer tests added!

## Docs

Bug fix. Should this be documented somewhere?

### DIFF
--- a/internal/token.go
+++ b/internal/token.go
@@ -269,6 +269,10 @@ type Tokenizer struct {
 	// token: one that treats "<p>" as text instead of an element.
 	// rawTag's contents are lower-cased.
 	rawTag string
+	// noExpressionTag is the "math" in "<math>". If non-empty, any instances
+	// of "{" will be treated as raw text rather than an StartExpressionToken.
+	// noExpressionTag's contents are lower-cased.
+	noExpressionTag string
 	// stringStartChar is the character that opened the last string: ', ", or `
 	// stringStartChar byte
 	// stringIsOpen will be true while in the context of a string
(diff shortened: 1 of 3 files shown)


internal/token.go
```diff
@@ -946,12 +955,14 @@ func (z *Tokenizer) hasTag(s string) bool {
...
 	// Several tags flag the tokenizer's next token as raw.
-	c, raw := z.buf[z.data.Start], false
+	c, raw, noExpression := z.buf[z.data.Start], false, false
 	switch c {
 	case 'i':
 		raw = z.startTagIn("iframe")
 	case 'n':
 		raw = z.startTagIn("noembed", "noframes")
+	case 'm':
+		noExpression = z.startTagIn("math")
```
jonathantneal, line 965: 👍 There it is.

natemoo-re: Thanks for commenting, forgot to do my usual pre-review comments.

PR #289, opened 2022-02-15
Fix conditional special elements

natemoo-re:
## Changes

- Closes https://github.com/withastro/compiler/issues/284
- Fixes conditional rendering of special elements like `iframe` and `noscript`.
- These were setting the parser `context`, which then meant `p.im()` would throw.

## Testing

Tests added

## Docs

Bug fix only

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -2763,7 +2763,7 @@ func (p *parser) parseCurrentToken() {
 		if p.inForeignContent() {
 			consumed = parseForeignContent(p)
 		} else {
-			if p.im == nil && p.context != nil {
+			if p.im == nil {
 				p.im = inBodyIM
 			}
 			consumed = p.im(p)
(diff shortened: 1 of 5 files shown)


internal/parser.go
```diff
@@ -2763,7 +2763,7 @@ func (p *parser) parseCurrentToken() {
 		if p.inForeignContent() {
 			consumed = parseForeignContent(p)
 		} else {
-			if p.im == nil && p.context != nil {
+			if p.im == nil {
```
natemoo-re, line 2766: This is the fix! Special elements like `iframe` and `noscript` were setting `p.context` which meant `p.im(p)` below would throw.

internal/parser.go
```diff
@@ -2763,7 +2763,7 @@ func (p *parser) parseCurrentToken() {
...
 				p.im = inBodyIM
 			}
 			consumed = p.im(p)
```
natemoo-re, line 2769: This was causing the `nil pointer dereference` error because, well, `p.im` was `nil`.

PR #290, opened 2022-02-16
chore: upgrade tinygo to `0.22.0`

natemoo-re:
## Changes

- Upgrades `tinygo` from `0.20.x` to `0.22.x`. Lots of WASM fixes have landed in TinyGo in the interim!
- Spent a long time wrestling with what turned out to be garbage collection issues. GC was too aggressive with [`-gc convservative`](https://github.com/tinygo-org/tinygo/issues/2187) (the default) and did not work with [`-gc none`](https://github.com/tinygo-org/tinygo/issues/2495), but works as it did in `0.20.x` when setting `-gc leaking`.
  - Despite the scary name, I _think_ this is the setting we want. See [the docs for more info](https://tinygo.org/docs/reference/usage/important-options/).

## Testing

Tests continue to pass!

## Docs

N/A


cmd/astro-wasm/astro-wasm.go
```diff
@@ -22,20 +22,22 @@ import (
 var done chan bool
 
 func main() {
-	js.Global().Set("__astro_transform", Transform())
-	// This ensures that the WASM doesn't exit early
-	<-make(chan bool)
+	js.Global().Set("@astrojs/compiler", make(map[string]interface{}))
+	module := js.Global().Get("@astrojs/compiler")
+	module.Set("transform", Transform())
```
natemoo-re, line 27: This is much cleaner. Instead of doing `__astro_transform`, we create `globalThis["@astrojs/compiler"]["transform"]` (and others in the future)

cmd/astro-wasm/astro-wasm.go
```diff
@@ -148,102 +150,104 @@ func Transform() interface{} {
 		handler := js.FuncOf(func(this js.Value, args []js.Value) interface{} {
 			resolve := args[0]
 
-			var doc *astro.Node
+			go func() {
+				var doc *astro.Node
```
natemoo-re, line 154: Spawning this in a goroutine now

lib/compiler/browser/index.ts
```diff
@@ -52,14 +52,7 @@ const startRunningService = async (wasmURL: string) => {
...
 
-  const apiKeys = new Set(['transform']);
-  const service: any = Object.create(null);
-
-  for (const key of apiKeys.values()) {
-    const globalKey = `__astro_${key}`;
-    service[key] = (globalThis as any)[globalKey];
-    delete (globalThis as any)[globalKey];
-  }
+  const service: any = (globalThis as any)['@astrojs/compiler'];
```
natemoo-re, line 55: Much more straight-forward.

lib/compiler/browser/wasm_exec.ts
```diff
@@ -280,6 +280,7 @@ export default class Go {
               let iov_ptr = iovs_ptr + iovs_i * 8; // assuming wasm32
               let ptr = mem().getUint32(iov_ptr + 0, true);
               let len = mem().getUint32(iov_ptr + 4, true);
+              nwritten += len;
```
natemoo-re, line 283: Changes in this file are upstream from `tinygo`.

lib/compiler/node/wasm_exec.ts
```diff
@@ -11,33 +11,43 @@ import fs from 'fs';
...
         const [sec, nsec] = process.hrtime();
         return sec * 1000 + nsec / 1000000;
       },
     },
-    enumerable: true,
-  },
-});
+  });
+}
+// End of polyfills for common API.
```
natemoo-re, line 46: Ran into some problems where `@astropub/webapi` polyfilling `crypto` would cause this to throw. Switched to verify that patching needs to happen before doing it.

lib/compiler/node/wasm_exec.ts
```diff
@@ -164,6 +174,7 @@ export default class Go {
               let iov_ptr = iovs_ptr + iovs_i * 8; // assuming wasm32
               let ptr = mem().getUint32(iov_ptr + 0, true);
               let len = mem().getUint32(iov_ptr + 4, true);
+              nwritten += len;
```
natemoo-re, line 177: Changes in below here are upstream from `tinygo`.

.github/workflows/ci.yml
```diff
@@ -43,12 +43,12 @@ jobs:
 
       - name: Get tinygo runtime
         run: |
-          wget -P /tmp/ https://github.com/tinygo-org/tinygo/releases/download/v0.20.0/tinygo_0.20.0_amd64.deb
-          sudo dpkg -i /tmp/tinygo_0.20.0_amd64.deb
+          wget -P /tmp/ https://github.com/tinygo-org/tinygo/releases/download/v0.22.0/tinygo_0.22.0_amd64.deb
+          sudo dpkg -i /tmp/tinygo_0.22.0_amd64.deb
```
natemoo-re, line 47: Updates `tinygo` to 0.22.0

.github/workflows/ci.yml
```diff
@@ -43,12 +43,12 @@ jobs:
...
 
       - name: Build WASM
         run: |
-          tinygo build -o ./lib/compiler/astro.wasm -target wasm ./cmd/astro-wasm/astro-wasm.go
+          tinygo build -gc leaking -scheduler asyncify -no-debug -o ./lib/compiler/astro.wasm -target wasm ./cmd/astro-wasm/astro-wasm.go
```
natemoo-re, line 51: As I mentioned, `-gc leaking` is necessary to avoid over-eager garbage collection that was cleaning up functions before we had an opportunity to call them.

This is because we're spawning goroutines and invoking a `Promise.resolve` callback inside of it. If we don't stop garbage collection from happening, `resolve` is undefined by the time we try to invoke it.

natemoo-re:
Closing in favor of #291 as some stress testing revealed memory issues with `-gc leaking` would happen almost immediately.

Issue #285, opened 2022-02-14
🐛 BUG: bus error 10... RuntimeError: memory access out of bounds


jladmore:
### What version of `astro` are you using?

0.23.0-next.6

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

We are using alpine components embedded into the astro page. We have found after upgrading to 0.23.0-next.6 we get a bus error 10 on our Mac and RuntimeError: memory access out of bounds on stackblitz.

The issue relates to the size of the html block in an alpinejs template entity. If we make it smaller it works ok. If we use another astro component say CartItems.astro and include the block this way it works. When we have the full block between the template we get the error and crash..

There seems to be more information from the stackblitz runtime, than what we get from our Mac build.  

This works fine in 0.22.20

Thanks

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-nvdxee

natemoo-re:
Confirming that #291 seems to fix this. See https://stackblitz.com/edit/withastro-astro-zd8qwj?file=package.json

Issue #278, opened 2022-02-02
🐛 BUG: HTML Lines being Truncated after 700 characters


aFuzzyBear:
### What version of `@astrojs/compiler` are you using?

v22.20

### What package manager are you using?

yarn

### What operating system are you using?

wsl2; ubuntu 20.04

### Describe the Bug

This is a really weird edge case I have found, and I believe it has to do with the compiler doing something funny.

I have found that there seems to be a upper limit which is far lower than the actual theoretical limit placed by the language, on the total number of characters that are printed to a single line in the html source code. 

I have found where it does truncate the string it does so in a manner that breaks the string itself, its very peculiar, for further information I append this small screenshot that demonstrates what It is I am seeing, alongside a reproduction. 

If this could be resolved I would really appreciate it, since it seems to of inadvertently placed a character limit on how much scripting XElement can do, which borks the entire component after 700characters, 
![unknown](https://user-images.githubusercontent.com/28299972/152169721-9e229f81-9482-4ec2-bc84-b7a9ef378b3f.png)

### Link to Minimal Reproducible Example

https://codesandbox.io/s/vigorous-platform-p2jtq?file=/src/components/ProblemChild.astro

natemoo-re:
Quite confident that this will be closed by https://github.com/withastro/compiler/pull/291

Issue #256, opened 2022-01-19
Bug track: nil pointer dereference


boehs:
### What operating system are you using?

Linux

### Describe the Bug

Recently, a wave of support threads in discord have been created discussing an issue. It has been impossible to reproduce, as it only occurs on certain linux devices. Known platforms issue *may* occur on

- some devices running Arch Linux
- Cloudflare
- Netlify
- Vercel

(note, using one of these platforms is not enough to trigger this bug, there are other unknown conditions that lead it to work on some devices but not others)

---

The following related errors can be experienced based on the contents of a file:

1.

```
wasm://wasm/0027!!!!!:1
RuntimeError: null function or function signature mismatch
    at wasm://wasm/00273ac2:wasm-function[293]:0x135!!
    at wasm://wasm/00273ac2:wasm-function[300]:0x142!!
    at Go._resume (file:///home/evan/.local/share/pnpm-global/5/node_modules/.pnpm/@astrojs+compiler@0.7.4/node_modules/@astrojs/compiler/node/wasm_exec.js:374:28)
    at file:///home/evan/.local/share/pnpm-global/5/node_modules/.pnpm/@astrojs+compiler@0.7.4/node_modules/@astrojs/compiler/node/wasm_exec.js:384:16
    at new Promise (<anonymous>)
    at syscall/js.valueNew (file:///home/evan/.local/share/pnpm-global/5/node_modules/.pnpm/@astrojs+compiler@0.7.4/node_modules/@astrojs/compiler/node/wasm_exec.js:274:54)
    at wasm://wasm/0027!!!!:wasm-function[321]:0x155!!
    at wasm://wasm/0027!!!!:wasm-function[821]:0x6!!!!
    at wasm://wasm/0027!!!!:wasm-function[293]:0x135!!
    at wasm://wasm/0027!!!!:wasm-function[300]:0x142!!
```

(note: ! indicates inconsistencies between reports)

2.

`RuntimeError: Error: Uh oh, the Astro compiler encountered an unrecoverable error!`

3.

Most common!

```
panic: runtime error: nil pointer dereference
RuntimeError: Error: Uh oh, the Astro compiler encountered an unrecoverable error!
```

---

Working versions are also inconsistent. In my testing, I found the following astro versions

- 0.22.14
- 0.22.3
- <0.22.1
[...]

PR #291, opened 2022-02-16
Switch to Go's builtin WASM output

natemoo-re:
## Changes

- I was hopeful that #290 would fix a lot of our issues, but that doesn't seem to be the case!
- The `stress.test.js` file has been added. TinyGo with `-gc leaking` will crash after about 2500 invocations since it runs out of memory. TinyGo with `-gc conservative` (the default) is no longer working for us. So we're back to square one.
- Go's default WASM output, while large (almost 4 MB), seems to run incredibly smoothly, even as I increased the stress test to run hundreds of thousands of times.
- Thankfully, switching us over required almost no changes to our source code!
- I strongly suspect that some of our lingering issues like https://github.com/withastro/compiler/issues/285, https://github.com/withastro/compiler/issues/278, and https://github.com/withastro/compiler/issues/256 will be addressed by switching back to the official Go WASM output. Given how volatile each issue is and how difficult to reproduce they can be, it's reasonable to assume that they are caused by inconsistent memory issues that can happen in TinyGo's WASM output.

## Testing

Stress test added, all other tests continuing to pass

## Docs

Need to update `CONTRIBUTING.md`


jonathantneal (approved):
Thank you. 🥺🙏

matthewp:
Let's put it through its paces and test against withastro/astro and in the community. Worth it if it cuts down on bugs and memory issues.

PR #294, opened 2022-02-16
[WIP] Allow external scoping

jonathantneal:
## Changes

- Provides short path to fixing CSS parsing issues.
- Provides short path to supporting container queries.
- This allows external style scoping with an `externalScoping` flag.
- Passes `transformOptions` to `preprocessStyle`, which should allow PostCSS to apply the styling.

## Testing

help needed

## Docs

help needed

### DIFF
--- a/cmd/astro-wasm/astro-wasm.go
+++ b/cmd/astro-wasm/astro-wasm.go
@@ -79,6 +79,11 @@ func makeTransformOptions(options js.Value, hash string) transform.TransformOpti
 		projectRoot = "."
 	}
 
+	externalScoping := false
+	if jsBool(options.Get("externalScoping")) {
+		externalScoping = true
+	}
+
 	staticExtraction := false
 	if jsBool(options.Get("experimentalStaticExtraction")) {
 		staticExtraction = true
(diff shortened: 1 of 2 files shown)


cmd/astro-wasm/astro-wasm.go
```diff
@@ -129,7 +135,7 @@ func preprocessStyle(i int, style *astro.Node, transformOptions transform.Transf
 		return
 	}
 	attrs := wasm_utils.GetAttrs(style)
-	data, _ := wasm_utils.Await(transformOptions.PreprocessStyle.(js.Value).Invoke(style.FirstChild.Data, attrs))
+	data, _ := wasm_utils.Await(transformOptions.PreprocessStyle.(js.Value).Invoke(style.FirstChild.Data, attrs, transformOptions))
```
matthewp, line 138: is this needed? I think that ProcessStyle has access to all of these options on the JS side anyways. Is there something that it needs?

jonathantneal: I think so. This was to have the object containing the hash value (e.g. `astro-hash1d2c3b4a`).

matthewp: We talked about this, make sense.

natemoo-re:
@jonathantneal I'm still tracking this, just FYI. I really want to unblock this and allow power users to bring their own scoping.

natemoo-re:
I dug into this over the weekend! 

While this is definitely one possible route we could take to unblock modern CSS features, the best fix would be to use a better CSS parser that supports all of the features we need out of the box. This PR would mean users need to bring their own logic for scoping, which I can see being a pain. 

I just opened a PR to swap in the `esbuild` CSS parser https://github.com/withastro/compiler/pull/329 Going to close this in favor of that.

PR #295, opened 2022-02-17
Fixes hoisted scripts passed a variable

matthewp:
## Changes

- The static build uses the hoisted metadata to track which hoisted scripts are part of a page, in order to build them.
- This fixes it so that expressions are passed as the variable and not quoted.
- Part of https://github.com/withastro/astro/issues/2604

## Testing

Test added

## Docs

N/A


jonathantneal (approved):
I’m "glad" for this "PR".

No no I really am! 😂

PR #297, opened 2022-02-17
fix memory usage caused by duplicate instantiations

FredKSchott:
## Changes

- Previously, a bug caused us to instantiate multiple instances of the compiler
- Combined with our larger WASM output, v0.11.0 may have made an existing memory issue even worse

## Testing

- Tested in examples/blog-with-multiple-authors (3 markdown pages) and reduced 7 calls to instantiation down to 1. Seemed to have some noticeable impact on build time from 3sec -> 2.5 (hard to replicate reliably, so take that with a big grain of salt).

## Docs

- N/A

### DIFF
--- a/.changeset/rich-gifts-change.md
+++ b/.changeset/rich-gifts-change.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fix memory issue caused by duplicate WASM instantiations
--- a/lib/compiler/browser/index.ts
+++ b/lib/compiler/browser/index.ts
@@ -9,26 +9,28 @@ interface Service {
   transform: typeof types.transform;
 }
 
-let initializePromise: Promise<void> | undefined;
+let initializePromise: Promise<Service> | undefined;
 let longLivedService: Service | undefined;
 
-export const initialize: typeof types.initialize = (options) => {
+export const initialize: typeof types.initialize = async (options) => {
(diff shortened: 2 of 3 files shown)


FredKSchott:
Interesting debugging note: putting some `console.time()` calls around the old instantiation logic returned the following performance data: 
```
03:14 PM [config] Set "buildOptions.site" to generate correct canonical URLs and sitemap
1: 467.749ms
2: 631.179ms
3: 790.95ms
0: 859.724ms
4: 1.562s
5: 1.601s
6: 1.610s
```

This was probably all parallel wait time (since the total build only took 2.6sec), but it still shows just how long it took for this instantiation logic to complete. @matthewp this could even be behind the terrible `.md -> .astro` perf that you were seeing on large sites!

natemoo-re (commented):
Looks great, good catch! Would you mind updating the `browser/index.ts` file as well?

FredKSchott:
Browser shouldn't need an update, since you need to await `initialize()` before calling `transform()`. But I'll update the code to at least match some of the new data types in node.ts

FredKSchott:
@natemoo-re done! Mainly just a code cleanup since no bug existed in browser, but the cleanup did smooth out 1-2 rough edges. TBH these all probably need a fresh pair of eyes at some point in the future!

natemoo-re (approved):
Thank you, thank you 🙏🏻

PR #301, opened 2022-02-24
Handle self-closing raw elements

natemoo-re:
## Changes

- Fixes our handling of elements like `<title />` and `<script />`.
- These cases were very unlikely before, but with the introduction of `set:text` and `set:html` they are much more likely.
- This was a change needed in the tokenizer because `<title />` would read everything following as raw text since it never encountered a closing `</title>` tag.

## Testing

Tests added

## Docs

Bug fix only


jonathantneal (approved):
Thank you. 🙏

PR #303, opened 2022-02-24
Warn when using an expression on hoist `src` attr

matthewp:
## Changes

- Adds a warning when doing `<script hoist src={expr}></script>`.
- That's because we need to statically resolve these, explained here: https://github.com/withastro/astro/issues/2648

## Testing

mjs test added

## Docs

N/A

### DIFF
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -154,7 +154,7 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions) {
 					if attr.Type == astro.ExpressionAttribute {
 						if opts.StaticExtraction {
 							shouldAdd = false
-							fmt.Printf("%s: <script hoist> uses the expression {%s} on the src attribute and will be ignore. Use a static string on the src attribute instead.\n", opts.Filename, attr.Val)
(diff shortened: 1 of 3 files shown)


internal/transform/transform.go
```diff
@@ -140,16 +140,31 @@ func NormalizeSetDirectives(doc *astro.Node) {
...
 		// if <script hoist>, hoist to the document root
 		// If also using define:vars, that overrides the hoist tag.
 		if hasTruthyAttr(n, "hoist") && !HasAttr(n, "define:vars") {
+			shouldAdd := true
+			for _, attr := range n.Attr {
+				if attr.Key == "src" {
+					if attr.Type == astro.ExpressionAttribute {
+						if opts.StaticExtraction {
+							shouldAdd = false
+							fmt.Printf("%s: <script hoist> uses the expression {%s} on the src attribute and will be ignore. Use a static string on the src attribute instead.\n", opts.Filename, attr.Val)
```
FredKSchott, line 157: nit: `ignored`. 
Also nit: slight preference for `static string literal` or just `string literal`  as more correct here

FredKSchott (approved):
LGTM on the concept, seems inline with the ["finalize script behavior"](https://github.com/withastro/rfcs/blob/main/proposals/0008-style-script-behavior.md) RFC. I don't know Go code well enough to review the code though :)

PR #308, opened 2022-02-28
Client only metadata

matthewp:
## Changes

- Adds the metadata for client:only components that is needed for the build

## Testing

Tests updated

## Docs

N/A


natemoo-re (approved):
Looks great!

PR #310, opened 2022-02-28
Fix next `head` behavior

natemoo-re:
## Changes

- Turns out https://github.com/withastro/astro/pull/2436 was failing because the `<head>` was being removed from some of our output.
- The problem seems to be in the `astro.ParseFragment` function, but given recent changes the `astro.Parse` function should work just as well.

## Testing

WASM tests updated, new one added

## Docs

Bug fix


FredKSchott (approved):
LGTM once tests are passing!

PR #316, opened 2022-03-02
Refactor WASM tests

natemoo-re:
## Changes

- Switches WASM tests to use a proper test runner, [`uvu`](https://github.com/lukeed/uvu).
- Cleans up and organizes test suite

## Testing

This is that.

## Docs

Nope


FredKSchott:
If there are any trouble spots to look at, lmk. Otherwise, I'm happy to LGTM based on a quick runthrough

matthewp (approved):
👍

PR #321, opened 2022-03-11
AST: Always output the attribute property in an element node even if empty

antonyfaris:
## Changes

There was no attribute property in the element node when the element didn't have attributes. This adds the property with an empty array as its value.

## Testing

I added a test for this case.

## Docs


natemoo-re (approved):
This looks great!

PR #330, opened 2022-03-14
Zero-specificity Scoping

natemoo-re:
## Changes

- Implements [RFC0012](https://github.com/withastro/rfcs/blob/main/proposals/0012-scoped-css-with-preserved-specificity.md)
- Still uses our current `.astro-XXXXXX` classes, but changes generated styles to not change the specificity of your selector.
  - In practice, this means that what previously generated `div.astro-XXXXXX` will now generate `div:where(.astro-XXXXXX)` and preserve the authored specificity of the `div` selector.

- [Compat table for `:where`](https://caniuse.com/mdn-css_selectors_where)

CleanShot 2022-07-22 at 13 44 56@2x

## Testing

Tests updated!

## Docs

Need to mention in migration guide


internal/transform/scope-html.go
```diff
@@ -39,37 +41,38 @@ var NeverScopedSelectors map[string]bool = map[string]bool{
...
 				return
 			}
 		}
 	}
 	// If we didn't find an existing class attribute, let's add one
 	n.Attr = append(n.Attr, astro.Attribute{
-		Key: "class",
-		Val: "astro-" + opts.Scope,
+		Key:  "data-astro-scope",
+		Type: astro.QuotedAttribute,
```
matthewp, line 75: Just curious, why switch to a data attribute instead of a class name?

natemoo-re: Great question! It's not strictly necessary since `:where()` will reset both `.astro-XXXXXX` and `data-astro-scope="XXXXXX"` to zero.

The motivation is that this attribute would be much more explicit. Currently `class` has an odd behavior where we append the parent `.astro-XXXXXX` to it. If you spread `{...Astro.props}` onto the child, it will inherit that `.astro-XXXXXX` and **pierce scoped encapsulation** without users realizing it. Separating this out to a `data-` attribute means we can filter it from `props` and allow users to opt-in to piercing behavior with a directive. It also removes the special-casing of `className` on component nodes and matches Vue's approach (which uses an attribute like `data-v-469af010`).

natemoo-re: I switched this back to our original approach with `class`! Hopefully makes this a smaller change to get in for v1.0.0-rc.

matthewp (approved):

PR #347, opened 2022-04-02
Support new script/style behavior

natemoo-re:
## Changes

- Implements [RFC0016](https://github.com/withastro/rfcs/blob/main/proposals/0016-style-script-defaults.md)
- Adds support for `is:inline` and `is:global`
- Companion PR to https://github.com/withastro/astro/pull/2961

## Testing

Tests updated

## Docs

Covered by core PR

### DIFF
--- a/internal/printer/print-to-js.go
+++ b/internal/printer/print-to-js.go
@@ -401,7 +401,7 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 		p.print(`]`)
 	} else {
 		for _, a := range n.Attr {
-			if transform.IsImplictNodeMarker(a) {
+			if transform.IsImplictNodeMarker(a) || a.Key == "is:inline" {
 				continue
 			}
 			if a.Key == "slot" {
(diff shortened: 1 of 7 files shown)


internal/printer/print-to-js.go
```diff
@@ -401,7 +401,7 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 		p.print(`]`)
 	} else {
 		for _, a := range n.Attr {
-			if transform.IsImplictNodeMarker(a) {
+			if transform.IsImplictNodeMarker(a) || a.Key == "is:inline" {
```
natemoo-re, line 404: This strips `is:inline` directives from the output

internal/transform/scope-css.go
```diff
@@ -19,7 +19,7 @@ outer:
 		if n.DataAtom != a.Style {
 			continue
 		}
-		if hasTruthyAttr(n, "global") {
+		if hasTruthyAttr(n, "is:global") {
```
natemoo-re, line 22: Removed support for `global`, just supporting `is:global`. Is that what we want? It's minimal to support both.

natemoo-re: Added a migration warning for `global`, but still supporting both.

FredKSchott:
LGTM!

Issue #348, opened 2022-03-30
🐛 BUG: Escaped {} within nested JSX fragment


d11z:
### What version of `astro` are you using?

0.25.4

### What package manager are you using?

npm

### What operating system are you using?

Windows

### Describe the Bug

JSX fragments seem to confuse the compiler somehow. This (pseudo)code causes the `{label ? ...}` to be escaped and renders as plain text instead of evaluating the code.

```tsx
---
// const sections = [ ... ];
---
<div>
  {sections.map(({ label, items }) => (
    <>
      {label ? "truthy" : "falsy"}
	  <span>another element</span>
    </>
  ))}
</div>
```

Results in:
```
{label ? "truthy" : "falsy"}
another element
{label ? "truthy" : "falsy"}
another element
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-bpawmn?file=src/components/Sidebar.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #360, opened 2022-04-10
Move `openBraceIsExpressionStart` assignment before `<>` check

williamtetlow:
## Changes
- Closes https://github.com/withastro/compiler/issues/348

## Testing

Tokenizer test added
## Docs
bug fix only


natemoo-re (approved):
Amazing, thanks so much for the fix! This is a great catch.

PR #362, opened 2022-04-12
Fix: keep explicit Component passing for `client:only`

bholmesdev:
## Changes
- resolves issue logged in Astro core: https://github.com/withastro/astro/issues/2966
- for `client:only` components, keep passing the `Component` to the `renderComponent` function instead of stubbing as `null`. This preserves the Component import at the top of the file for static analysis by rollup. Verified this doesn't break any `client:only` behavior like accessing the global `window`

## Testing

Updated unit tests for `client:only` JS printer

## Docs

N/A

### DIFF
--- a/.changeset/old-oranges-doubt.md
+++ b/.changeset/old-oranges-doubt.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+For client:only components, keep passing Component to renderComponent for JS printer
--- a/internal/printer/print-to-js.go
+++ b/internal/printer/print-to-js.go
@@ -330,7 +330,6 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 
 	isFragment := n.Fragment
 	isComponent := isFragment || n.Component || n.CustomElement
-	isClientOnly := isComponent && transform.HasAttr(n, "client:only")
 	isSlot := n.DataAtom == atom.Slot
 	isImplicit := false
 	for _, a := range n.Attr {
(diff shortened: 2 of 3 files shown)


matthewp:
I don't think this is right, the point of `client:only` is to prevent it from running on the server. It might be passing in some cases because we do polyfill some global variables, but we can't assume that that's enough.

We should be building the client only components here: https://github.com/withastro/astro/pull/3095

I'm happy to take a look if you are stuck on this one.

bholmesdev:
> I don't think this is right, the point of `client:only` is to prevent it from running on the server. It might be passing in some cases because we do polyfill some global variables, but we can't assume that that's enough.
> 
> We should be building the client only components here: [withastro/astro#3095](https://github.com/withastro/astro/pull/3095)
> 
> I'm happy to take a look if you are stuck on this one.

But the top level component `import` is _still_ always present even before this change? This only passes the component to the `renderComponent` function as an argument, preventing `esbuild` from stripping the `import` as unused ([see here](https://github.com/withastro/astro/blob/main/packages/astro/src/vite-plugin-astro/index.ts#L166-L172)).

The only alternative I see is returning `clientOnlyComponents` as a property from the compiler. Otherwise, there's no way for our rollup plugins to know what components a given page is using. I personally prefer this current approach since it doesn't explicitly invoke the `Component`, and the top-level ESM import was already present before this PR.

bholmesdev:
Update: I'll attempt to avoid stripping the component import from core's end. Played around with some esbuild settings to no avail, but maybe I was missing something 🤷

bholmesdev:
Closing in favor of solution in Astro core!

PR #363, opened 2022-04-14
Add support for HTML minification

natemoo-re:
## Changes

- Adds a new `compact` option to `transform({})`. When `compact === true`, whitespace will be automatically collapsed in the output.
- This avoids the need for a post-processing hook to minify the generated HTML.
- This also avoids the need for special minification logic to handle Islands properly, as those will be output directly by the frameworks.
- If text node is entirely whitespace, it will be removed.
- If text node is prefixed or suffixed with multiple `\n` characters, they will be collapsed to a single `\n`
- If text node is prefixed or suffixed with multiple inline whitespace characters, they will be collapsed to a single ` `.

## Testing

Multiple tests added for Go and WASM.

## Docs

N/A


huan086:
I suggest that the `compact` property be made a string instead of a boolean. In the future, there could be other more aggressive whitespace removal.

E.g. An aggressive whitespace removal like React's would remove all whitespace between different Node

```
assert.match(await minify('<p>  <a href="#">  <code>   foo</code></a> bar   </p>'), '<p><a href="#"><code>foo</code></a>bar</p>');
assert.match(await minify('<p>  <a href="#">  <code>{'   '}foo</code></a> bar   </p>'), '<p><a href="#"><code>   foo</code></a>bar</p>');
```

I find this useful for formatting the HTML for readability, without introducing whitespace into the output HTML which would cause "gaps" between the nodes

FredKSchott (approved):
LGTM from a quick look over, but I'm no Go expert.

Also, can you confirm that sourcemaps are still working, or at least aren't significantly regressed with this change?

natemoo-re:
@FredKSchott just tested and this step happens after we attach the original source location to the token and before we point the sourcemap location to the generated line. So from my testing this didn't have any effect on the sourcemaps. 

I'll be adding sourcemap tests in a follow-up PR.

PR #364, opened 2022-04-14
Add support for comment block only expressions and comment blocks in attributes and props

williamtetlow:
## Changes
- Adds support for expressions that only contain a comment block
  - `{/* a comment */}`
- Adds support for comments in Shorthand
  - `<h1 {/* comment in shorthand */ id}>Hello</h1>`
- Adds support for Shorthand attributes that only contain a comment
  - `<h1 {/* id={prop commentd out} */} class={class}>Hello</h1>`  
- Adds panic when a comment block is unterminated (missing `*/`)

See tests https://github.com/withastro/compiler/pull/364/files#diff-d0dbf4b902747f2cdc45290c88eb59b04d03bebcb2bd8872a0b49284a9dac9d1R1810

## Testing
Tests added for
- Comment only expressions
- Comments in attributes/props
- Panic when a comment block is unterminated

## Docs

I don't think we need to add this to docs explicitly

## TODO
- [x] Add support for comments in attributes/props
- [x] Add support for comment blocks

### DIFF
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -170,14 +170,16 @@ func (p *printer) printFuncSuffix(opts transform.TransformOptions) {
 }
 
 func (p *printer) printAttributesToObject(n *astro.Node) {
+	lastAttributeSkipped := false
 	p.print("{")
 	for i, a := range n.Attr {
-		if i != 0 {
+		if i != 0 && !lastAttributeSkipped {
 			p.print(",")
 		}
 		if a.Key == "set:text" || a.Key == "set:html" || a.Key == "is:raw" {
 			continue
 		}
+		lastAttributeSkipped = false
 		switch a.Type {
 		case astro.QuotedAttribute:
 			p.addSourceMapping(a.KeyLoc)
@@ -204,8 +206,13 @@ func (p *printer) printAttributesToObject(n *astro.Node) {
 			p.addSourceMapping(loc.Loc{Start: a.KeyLoc.Start - 3})
 			p.print(`...(` + strings.TrimSpace(a.Key) + `)`)
 		case astro.ShorthandAttribute:
+			withoutComments := removeComments(a.Key)
+			if len(withoutComments) == 0 {
+				lastAttributeSkipped = true
+				continue
+			}
 			p.addSourceMapping(a.KeyLoc)
-			p.print(`"` + strings.TrimSpace(a.Key) + `"`)
+			p.print(`"` + withoutComments + `"`)
 			p.print(":")
 			p.addSourceMapping(a.KeyLoc)
 			p.print(`(` + strings.TrimSpace(a.Key) + `)`)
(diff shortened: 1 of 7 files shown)


williamtetlow:
I don't think this needs to be added to docs but let me know!

We need to update VSCode to use `//` when you click `ctrl + /` inside a tag `<something a="1"`.

I'll raise this in Discord

internal/token.go
```diff
@@ -1293,7 +1296,7 @@ func (z *Tokenizer) readTagAttrExpression() {
 			}
 		// Handle comments, strings within attrs
 		case '/', '"', '\'':
-			if len(z.attrTemplateLiteralStack) != 0 && c == '/' {
+			if z.attrTemplateLiteralStack[z.attrExpressionStack-1] != 0 && c == '/' {
```
williamtetlow, line 1299: ☝️ I updated logic here slightly.

`attrTemplateLiteralStack` was introduced in https://github.com/withastro/compiler/pull/280 but it looks like what we want to test here is if there is an opening backtick (so currently in a template literal) not if `attrTemplateLiteralStack` is empty

internal/printer/utils.go
```diff
@@ -53,3 +53,29 @@ func escapeSingleQuote(str string) string {
 func encodeDoubleQuote(str string) string {
 	return strings.Replace(str, `"`, "&quot;", -1)
 }
+
+// Remove comment blocks from string (e.g. "/* a comment */aProp" => "aProp")
+func removeComments(input string) string {
```
williamtetlow, line 58: This returns a string with all comments removed

For example...

```js
// input
"/*a comment */ hello /*a comment*/ world /*a comment*/"

// output
"hello world"`
```

internal/printer/printer.go
```diff
@@ -291,11 +298,15 @@ func (p *printer) printAttribute(attr astro.Attribute) {
...
+		withoutComments := removeComments(attr.Key)
+		if len(withoutComments) == 0 {
+			return
+		}
 		p.print(fmt.Sprintf("${%s(", ADD_ATTRIBUTE))
 		p.addSourceMapping(attr.KeyLoc)
 		p.print(strings.TrimSpace(attr.Key))
 		p.addSourceMapping(attr.KeyLoc)
-		p.print(`, "` + strings.TrimSpace(attr.Key) + `")}`)
+		p.print(`, "` + withoutComments + `")}`)
```
williamtetlow, line 309: This and [`ln 208`](https://github.com/withastro/compiler/pull/364/files#diff-58de1575bcd3abab85340eb61810590c68a061ec2cad49d8cf3864fe402a64c1L208) would resolve the key for shorthand attribute with comments like so `{/* a comment */ id} => { "/* a comment */ id":  id }`

We remove the comments now when printing the key to fix this

natemoo-re (approved):
YES! This is awesome work, thank you so much.

PR #393, opened 2022-05-13
fix(#380): fix json escape

### LINKED ISSUE #380: 🐛 BUG: `\` escape problem with parse method.

0.14.2

npm

Mac

### Describe the Bug

Using the parse method crashes because `\` is not escaped as JSON.

I expected the following template to be able to get the AST with the parse method, but it actually crashes.

```astro
---
const foo = '\''
---
{foo}
```

To reproduce it, clone the link and execute the following command.

https://github.com/ota-meshi/withastro-compiler-bug-2022-05

```
npm install
npm run test
```

Or run the following js.

```mjs
import {parse} from "@astrojs/compiler"
void main()

/** main */
async function main() {
    const contents = `
---
const foo = '\\''
---
{foo}`
    const result = await parse(contents)
    console.log(result)
}
```

https://github.com/ota-meshi/withastro-compiler-bug-2022-05

ota-meshi:
## Changes

Fixes #380

This PR fixes `escapeForJSON` to escape the backslash (`/`).

## Testing

I added a test case for a template with backslashes to `internal/printer/printer_test.go TestPrintToJSON` and verified that `go test -v ./internal/...` command passed.

## Docs

bug fix only.

### DIFF
--- a/.changeset/six-squids-wink.md
+++ b/.changeset/six-squids-wink.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fix json escape
--- a/internal/printer/print-to-json.go
+++ b/internal/printer/print-to-json.go
@@ -37,6 +37,8 @@ type ASTNode struct {
 }
 
 func escapeForJSON(value string) string {
+	backslash := regexp.MustCompile(`\\`)
+	value = backslash.ReplaceAllString(value, `\\`)
 	newlines := regexp.MustCompile(`\n`)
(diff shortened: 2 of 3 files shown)


ota-meshi:
The linting error doesn't seem to have anything to do with my changes. If you know I have to do something, please let me know.

natemoo-re:
Thanks so much for the PR @ota-meshi! The linting issue should be fixed by https://github.com/withastro/compiler/pull/397 and then I'll rerun this PR.

natemoo-re:
Looks great, thanks so much for the fix!! 🎉

PR #396, opened 2022-05-16
Do not hoist nested scripts or styles

natemoo-re:
## Changes

- Fixes https://github.com/withastro/compiler/issues/395 and a variety of other edge cases where `style` and `script` should not be hoisted out.
- Adheres behavior to [RFC0008](https://github.com/withastro/rfcs/blob/main/proposals/0008-style-script-behavior.md#new-rfc-behavior)
- In practice, this means that `script` and `style` elements that are not at the top-level of a template are treated as `is:inline`.

## Testing

Existing tests updated

## Docs

TODO


natemoo-re:
Closing in favor of https://github.com/withastro/compiler/pull/473

PR #399, opened 2022-05-16
fix(#391): inject astro scoped class when `class:list` is used

### LINKED ISSUE #391: 🐛 BUG: Tags with the class:list directive do not get the scoped style class applied

1.0.0-beta.12

None

npm

Linux (WSL Ubuntu)

### Describe the Bug

Placing a `<style>` tag inside a component scopes the styles at build time by applying a hashed class (`astro-${HASH}`)to all tags inside it. But elements that have the `class:list` directive do not appear to get the hashed class applied, making scoped styles impossible to use on those elements.

In the linked reproduction the `<h1>` tag gets the `'foo'` class via the `class:list` directive, and the scoped styles target the `.foo` class, but the styles aren't applied because the generated CSS is expecting the scoped styles class.

https://stackblitz.com/edit/github-rgs6pb?file=src%2Fpages%2Findex.astro&on=stackblitz

### Participation

- [X] I am willing to submit a pull request for this issue.

natemoo-re:
## Changes

- Injected `astro-XXXXXX` when `class:list` is used instead of `class`
- Fixes #391 

## Testing

Tests updated

## Docs

Bug fix only

### DIFF
--- a/.changeset/hip-games-trade.md
+++ b/.changeset/hip-games-trade.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fix Astro scoping when `class:list` is used
--- a/internal/transform/scope-html.go
+++ b/internal/transform/scope-html.go
@@ -1,6 +1,8 @@
 package transform
 
 import (
+	"fmt"
+
 	astro "github.com/withastro/compiler/internal"
 )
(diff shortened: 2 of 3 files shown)


matthewp (approved):
lgtm

PR #400, opened 2022-05-16
fix: properly handle scoped class injection with spread attributes

natemoo-re:
# Do not merge until core `spreadAttributes` is updated!

## Changes

- Long-standing bug that was tricky to fix.
- `<div {...Astro.props} />` would emit duplicate `class` attributes!
  - The compiler would scope this internally, so if `Astro.props` included `class` that would lead to duplicate `class` attributes and would throw in Vite.
- The solution is to update the compiler logic for injection, skipping `class` is a `...spread` attribute is found. This should be paired with an update to the `spreadAttributes` function to accept a third param of attributes that should alway be injected (`class`).

## Testing

Tests added

## Docs

Bug fix only!

### DIFF
--- a/.changeset/curvy-peas-buy.md
+++ b/.changeset/curvy-peas-buy.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fix long-standing bug where a `class` attribute inside of a spread prop will cause duplicate `class` attributes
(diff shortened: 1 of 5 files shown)


tony-sull (approved):
lgtm!  Clean fix for handling the class collision when spreading props :+1:

PR #402, opened 2022-05-17
Add TSX output

natemoo-re:
## Changes

- Add `convertToTSX` method to emit pure TSX that we can typecheck
- cc @Princesseuh 

## Testing

Simple test added, need more edge cases covered!

## Docs

TODO

### DIFF
--- a/internal/printer/print-to-tsx.go
+++ b/internal/printer/print-to-tsx.go
@@ -1,0 +1,191 @@
+package printer
+
+import (
+	"fmt"
+	"strings"
+
+	. "github.com/withastro/compiler/internal"
+	astro "github.com/withastro/compiler/internal"
+	"github.com/withastro/compiler/internal/sourcemap"
+	"github.com/withastro/compiler/internal/transform"
+)
+
+func PrintToTSX(sourcetext string, n *Node, opts transform.TransformOptions) PrintResult {
+	p := &printer{
+		opts:    opts,
+		builder: sourcemap.MakeChunkBuilder(nil, sourcemap.GenerateLineOffsetTables(sourcetext, len(strings.Split(sourcetext, "\n")))),
+	}
+
+	renderTsx(p, n, RenderOptions{
+		isRoot: true,
+		opts:   opts,
+	})
+	return PrintResult{
+		Output:         p.output,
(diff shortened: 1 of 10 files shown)


Princesseuh:
Here's a non-exhaustive list of things to take into account (not considering current state):

**Script, style and `is:raw`**
- Style tags and other tags with non-TS friendly content should be template literals (ex: Markdown, `is:raw`)
- Script tags should be turned into function calls (ex: `{() => {content}}`), as certain TypeScript features are able to parse the content of said functions (ex: document symbols), inlay hints too I believe? without needing any special care

**Attributes**
- Attributes starting with `@` are not allowed in JSX, as such they should be replaced with something else, this also applies to interface definitions, as TypeScript will complain about missing types otherwise. Thanks, JSX
- Shorthands props work differently in JSX than in Astro, props using the shorthand syntax should be converted to the long one

**Return type**
- The language server expect the name of the function to be a pascal-cased name (built from the file name) followed by the `__AstroComponent_` suffix (ex: `MyButton__AstroComponent_` from a file called `MyButton.astro`). This is used in order to be able to provide auto imports and quick fixes with imports, the suffix is used to provide special care for components (this is what powers the icons inside the completions!)
- The `props` property should be prefixed with an underscore (`_props`), otherwise TypeScript will complain about it being unused (which ends up in the user's diagnostics). This could be fixed in the language server if needed

Will edit with more details as I remember them, in the mean time, looking at [our current `astro2tsx` can be resourceful](https://github.com/withastro/language-tools/blob/main/packages/language-server/src/plugins/typescript/astro2tsx.ts)

internal/printer/print-to-tsx.go
```diff
@@ -0,0 +1,296 @@
...
+
+func isScript(p *astro.Node) bool {
+	return p.DataAtom == atom.Script
+}
+
+var ScriptMimeTypes map[string]bool = map[string]bool{
+	"module":                 true,
+	"text/typescript":        true,
+	"application/javascript": true,
+	"text/partytown":         true,
```
tony-sull, line 35: official partytown support!  😄

tony-sull (approved):
lgtm!  := shipIt

PR #403, opened 2022-05-19
Provide a good error when unable to find matching client:only import …

matthewp:
## Changes

- Checks to see if any client:only components did not have a matching import statement.
- If there are any, panic that the compiler cannot complete successful, providing a good error message for the scenario.

## Testing

- Test added when trying to use a frontmatter variable with client:only.

## Docs

N/A, bug fix.


natemoo-re (approved):
Nice, thanks for improving this!

PR #405, opened 2022-05-21
fix: don't parse expressions in <math> elements

### LINKED ISSUE #386: 🐛 BUG: Math in markdown throws errors or crashes Astro fully.

0.22.6

pnpm

macOS Monterey 12.1

### Describe the Bug

When using `remark-math` and `remark-katex` to extend the default markdown pipeline `@astrojs/markdown-remark` (as described in the docs),

- In v0.22.3 _curly braces used in formulas_ cause Astro to throw errors.
`$f\colon\mathbb R^x$` is translated correctly.
`$f\colon\mathbb R^{x}$` throws `Syntax error "x"`.

- In v0.22.6 Astro quits unexpectedly in both cases with shell error `zsh: bus error  pnpm dev`.

Please view the full example linked below.

EXPECTED BEHAVIOR:
Strings surrounded by `$`-signs should be ignored by Astro and handled by `remark-math` and `remark-katex`. Astro should not try to evaluate contents of curly braces as JS expressions when they are nested within a math context.

https://stackblitz.com/github/AlexLike/astro-math-example

virchau13:
## Changes

This should fix #386, as well as a seperate bug where self-closing tags would override the `noExpressionTag` context. The main issue with #386 was actually due to the parser calling `NextIsNotRawText`, which resets `noExpressionTag`, which I believe is incorrect ("raw text" as in a textarea is different from "raw text" in foreign `<math>` elements.)

This is my first attempt to write Go code, so I probably messed something up; feel free to point it out.

## Testing

The self-closing tag bug is tested in `token_test.go`. #386 is a parser bug, so I'm not actually sure how to test it (please let me know how!) 

## Docs

Bug fix only.


natemoo-re (approved):
Clean fix, thank you for catching this!

PR #406, opened 2022-05-23
fix: strip newlines at the end of component files

virchau13:
## Changes

This makes it so that newlines at the end of files are not considered. This is because it can lead to confusing behaviour; defining an Astro component with a newline at the end (which can be implicitly placed by most editors) adds an implicit newline to the final output, resulting in strange spaces in the final HTML.

For example:

```astro
<!-- src/components/Link.astro -->
<a href={Astro.props.href}><slot /></a>
```

```astro
---
/* src/pages/index.astro */
import Link from '../components/Link.astro';
---
<p>this is a <Link href="example.com">test</Link>!</p>
```

results in a strange space between the link text and the exclamation mark:
strange space example

This change simply makes it so that the three exposed JS functions all strip newlines from the end of the source text before attempting to use it.

## Testing

This was manually tested. Does this need automated testing?

## Docs

Bug fix only. (Does this need documentation?)

### DIFF
--- a/.changeset/chilly-chairs-enjoy.md
+++ b/.changeset/chilly-chairs-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': minor
+---
+
+Newlines at the end of Astro files are now stripped before the Astro file is processed.
--- a/.changeset/chilly-chairs-enjoy.md
+++ b/.changeset/chilly-chairs-enjoy.md
@@ -2,4 +2,4 @@
 '@astrojs/compiler': minor
 ---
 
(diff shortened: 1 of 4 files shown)


matthewp:
It looks like some tests are expecting this. I think it's likely that they really don't care, so you might want to take a look and update the tests.

natemoo-re (approved):
This is great! Thanks for providing a really clear use case where this behavior breaks expectations.

I updated the implementation to trim trailing space nodes without updating the source text (which could throw off source maps).

I also added a test suite to check for this behavior explicitly.

PR #410, opened 2022-05-24
Prevent including module metadata imports for side-effectual CSS

matthewp:
## Changes

- The compiler adds import statements for each user import in order to figure out which one maps to a component.
- CSS side-effectual imports should *not* be included in this metadata. Doing so prevents CSS try shaking this code away.
- This is part of switching to Vite's internal CSS plugin and fixing https://github.com/withastro/astro/issues/2146

## Testing

- Test added

## Docs

N/A, bug fix.

### DIFF
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -2,6 +2,7 @@ package printer
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	astro "github.com/withastro/compiler/internal"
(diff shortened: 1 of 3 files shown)


internal/printer/printer.go
```diff
@@ -43,6 +44,7 @@ var RESULT = "$$result"
 var SLOTS = "$$slots"
 var FRAGMENT = "Fragment"
 var BACKTICK = "`"
+var styleModuleSpecExp = regexp.MustCompile(`(\.css|\.sass|\.scss)$`)
```
natemoo-re, line 47: Any reason not to match this list?

https://github.com/withastro/astro/blob/main/packages/astro/src/core/render/util.ts#L26-L36

natemoo-re (commented):
Looks good, I would just make sure we're matching `core` in terms of what constitutes a CSS import

internal/printer/printer.go
matthewp: Nope, can add all of those 👍

PR #415, opened 2022-05-31
fix(#413): fix wasm `parse` to save attribute namespace

### LINKED ISSUE #413: 🐛 BUG: wasm `parse()` does not return the attribute namespace.

0.15.1

npm

Mac

### Describe the Bug

The namespace is not included in the attribute node of AST obtained by the wasm `parse` method.

I think the name of the AST node for the third attribute in the following example should be `xmlns:xlink`.
(Or we need the namespace property? )

Example:

```astro
<svg 
  viewBox="0 0 147 40"
  xmlns="http://www.w3.org/2000/svg"
  xmlns:xlink="http://www.w3.org/1999/xlink">
</svg>
```

AST:

```json
{
  "type": "root",
  "children": [
    {
      "type": "element",
      "name": "svg",
      "attributes": [
        {
          "type": "attribute",
          "kind": "quoted",
          "name": "viewBox",
          "value": "0 0 147 40"
        },
        {
          "type": "attribute",
          "kind": "quoted",
          "name": "xmlns",
          "value": "http://www.w3.org/2000/svg"
        },
        {
          "type": "attribute",
          "kind": "quoted",
          "name": "xlink",
          "value": "http://www.w3.org/1999/xlink"
        }
      ],
      "children": [
        {
          "type": "text",
          "value": "\n"
        }
      ]
    },
    {
      "type": "text",
      "value": "\n"
    }
  ]
}
```

We probably need to handle the namespace in the following places:

https://github.com/withastro/compiler/blob/d5dd438d3f0aa5f005e9b4d7127230711ab126a1/internal/printer/print-to-json.go#L207-L213

https://stackblitz.com/edit/github-au6cvz?file=src%2Fpages%2Findex.astro&on=stackblitz

https://github.com/ota-meshi/withastro-compiler-bug-2022-05-31

ota-meshi:
## Changes

Fixes #413

This PR fixes the wasm `parse` method to save the attribute namespace.

## Testing

I added a test case for a template with attribute namespace to `internal/printer/printer_test.go TestPrintToJSON` and verified that `go test -v ./internal/...` command passed.

## Docs

bug fix only.

---

I'm not familiar with the go lang, so I don't know best practices such as variable declarations. If I'm doing something strange by convention, please let me know.

### DIFF
--- a/internal/printer/print-to-json.go
+++ b/internal/printer/print-to-json.go
@@ -204,11 +204,9 @@ func renderNode(p *printer, parent *ASTNode, n *Node, opts t.ParseOptions) {
 			}
 
 			for _, attr := range n.Attr {
-				var name string
+				name := attr.Key
 				if attr.Namespace != "" {
-					name = attr.Namespace + ":" + attr.Key
-				} else {
-					name = attr.Key
+					name = fmt.Sprintf("%s:%s", attr.Namespace, attr.Key)
 				}
 				attrNode := ASTNode{
 					Type:     "attribute",


internal/printer/print-to-json.go
```diff
@@ -204,11 +204,17 @@ func renderNode(p *printer, parent *ASTNode, n *Node, opts t.ParseOptions) {
 			}
 
 			for _, attr := range n.Attr {
+				var name string
+				if attr.Namespace != "" {
+					name = attr.Namespace + ":" + attr.Key
+				} else {
+					name = attr.Key
+				}
```
natemoo-re, line 212: Definitely nothing wrong with your approach but I would find this a little bit cleaner!

- Assign to `name` only if `attr.Namespace != ""`
- Use `fmt.Sprintf()` to concat namespace and key.

```suggestion
				name := attr.Key
				if attr.Namespace != "" {
					name = fmt.Sprintf("%s:%s", attr.Namespace, attr.Key)
				}
```

natemoo-re (approved):
This is a clean fix, thank you!

PR #419, opened 2022-06-04
Fix(tokenizer): reckon left bracket which previous char is not $ as raw left bracket in template literal in attribute

Shinyaigeek:
## Changes

Fixes: https://github.com/withastro/astro/issues/3497

- What does this change?
  - fix https://github.com/withastro/astro/issues/3497 issue.
  - in the current implementation, `{` in template literal in the attribute is marked as the start of Expression even if that’s previous character is not `$`. This causes the above issue
  - I fixed this to check whether previous charactor is `$` or not.

## Testing

I wrote a unit test to check astro/compiler can handle below code properly

```
<main id={`{`}>
</main>
```

## Docs

documentation update is not needed because this pr only includes bug fix


natemoo-re:
Thanks for the contribution @Shinyaigeek! Sorry it took so long to get back to you.

I moved your test into the Go codebase since it's a bit easier to reason about. This is a great fix! Really appreciate it. ❤️

PR #426, opened 2022-06-13
Expose diagnostics

natemoo-re:
## Changes

- Adds `errors` and `warnings` to output of `transform()` function
- Adds `errors` and `warnings` to output of `convertToTSX()` function
- Adds `errors` and `warnings` to output of `parse()` function
- Replace `panic`s in the compiler

Will require a follow-up in core to surface these errors.

## Testing

Tests added for new behavior

## Docs

N/A


matthewp (approved):
Nice refactor.

natemoo-re:
Just waiting on #542 before releasing a minor.

.changeset/curly-beers-raise.md
```diff
@@ -0,0 +1,19 @@
...
+const { errors } = await transform(file, opts);
+for (const error of errors) {
+  console.error(error.text);
+}
+// or
+const { code, warnings } = await transform(file, opts);
+for (const warning of warnings) {
+  console.warn(warning.text);
+}
+```
```
Princesseuh, line 19: Not a big deal, but the changeset is outdated for the new return shape

natemoo-re: Great catch, thank you!

Princesseuh (approved):
Looks good to me, I haven't tested it in the AST context, but in the TSX output there's no apparent issues!

Very nice work, as always

PR #433, opened 2022-06-17
Include component metadata in the transformResult

matthewp:
## Changes

- Includes component metadata as part of the transform result: `hydratedComponents` and `clientOnlyComponents`. This includes the specifier, resolvedPath, and exportName.
- This information will be used by the build to know what client JS needs to be built, without running the code in dev SSR mode as we do today.

## Testing

- Tests added

## Docs

N/A


natemoo-re (approved):
LGTM!

PR #444, opened 2022-06-27
Fix fallback content in slots

### LINKED ISSUE #441: 🐛 BUG: Named slot with fallback content causes a panic in the compiler

v1.0.0-beta.56

Deno

yarn

Windows

### Describe the Bug

With the named slot components, if you have a slot that has fallback content, it causes a compiler panic (citing an invalid memory address or nil pointer dereference).

May be related to https://github.com/withastro/astro/issues/3253

https://codesandbox.io/s/distracted-banzai-shciz3

### Participation

natemoo-re:
## Changes

- Closes #441
- Fixes edge case with fallback content inside of slots
- Edge case was hit when `slot` was a top-level child

## Testing

Tests added

## Docs

Bug fix only

### DIFF
--- a/.changeset/pretty-horses-cough.md
+++ b/.changeset/pretty-horses-cough.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fix issue with fallback content inside of slots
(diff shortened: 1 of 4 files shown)


FredKSchott (approved):
LGTM but can't say I know 100% what I'm looking at here

tony-sull (approved):
LGTM

PR #445, opened 2022-07-05
fix(#3823): scope html and body elements

natemoo-re:
## Changes

- Fixes https://github.com/withastro/astro/issues/3823
- For historical reasons, `body` was not scoped and `html` had inconsistent behavior.
- Now `html` and `body` are scoped like any element.

## Testing

Relevant tests updated

## Docs

Bug fix only


tony-sull (approved):
lgtm

Issue #352, opened 2022-04-03
🐛 BUG: Component slots in markdown pages throw an error


tony-sull:
### What version of `@astrojs/compiler` are you using?

0.14.1

### What package manager are you using?

pnpm

### What operating system are you using?

mac

### Describe the Bug

Ran into a build error when upgrading astro.build to astro@0.26.  Narrowed it down to component slots being used in two markdown pages

```html
<!-- src/pages/blog/introducing-astro.md -->
<Blockquote>
  <Fragment slot="quote">
    A well-designed system makes it easy to do the right things and annoying (but not impossible) to do the wrong things
  </Fragment>

  <Fragment slot="cite">Jeff Atwood, <a href="https://blog.codinghorror.com/falling-into-the-pit-of-success/">Falling Into The Pit of Success</a></Fragment>
</Blockquote>
```

**Error message**
```
panic: Element with a slot='...' attribute must be a child of a component or a descendant of a custom element

goroutine 124 [running]:
github.com/withastro/compiler/internal/printer.render1(0x5f9b08, 0x57f860, {0x0, 0x0, 0x3, 0x0, {{0x673e40, 0x8}, {0x4b05f0, 0x4f}, ...}})
	github.com/withastro/compiler/internal/printer/print-to-js.go:409 +0x533
```

### Link to Minimal Reproducible Example

https://github.com/withastro/astro.build/tree/chore/astro-0.26

JuanM04:
It also happend to me with the slot being inside a conditional:

```astro
<MainLayout {...props}>
  <slot />

  {mathStyles && <link rel="stylesheet" href={mathStyles} slot="head" />}
</MainLayout>
```

natemoo-re:
Ah that might be the culprit, @JuanM04! I probably won't be able to look at this for another week, but I'd really like to fix this ASAP.

umbrellasystem404:
> ### What version of `@astrojs/compiler` are you using?
> 
> 0.14.1
> 
> ### What package manager are you using?
> 
> pnpm
> 
> ### What operating system are you using?
> 
> mac
> 
> ### Describe the Bug
> 
> Ran into a build error when upgrading astro.build to astro@0.26.  Narrowed it down to component slots being used in two markdown pages
> 
> ```html
> 
> <Blockquote>
>   <Fragment slot="quote">
>     A well-designed system makes it easy to do the right things and annoying (but not impossible) to do the wrong things
>   </Fragment>
> 
[...]

PR #447, opened 2022-07-06
Support slot attribute inside expressions

### LINKED ISSUE #383: 🐛 BUG: Conditional named slots break the compiler

1.0.0-beta.19

None

yarn

Mac

### Describe the Bug

Any way of including named slots with JSX interpolation breaks the compiler.
```astro
<TestAstro>
	{isTrue && <p slot="name">Dan</p>}
</TestAstro>
```
Even though this is indeed the direct child of a component, the compiler enters `panic` mode with the message:
`panic: Element with a slot='...' attribute must be a child of a component or a descendant of a custom element`

https://stackblitz.com/edit/github-idj9xd-ghiwxt?file=src/pages/index.astro

### Participation

natemoo-re:
## Changes

- Closes #383, #352, #388
- Paired with `astro` PR https://github.com/withastro/astro/pull/3837
- Fixes support for the `slot` attribute inside expressions
- Previously, we treated `{}` as an opaque element, so any children with the `slot` attribute were not transformed.
- Now, we dive into the `{}` expression to identify any children with a `slot` attribute and bubble them up.
- Introduces `mergeSlots` util that recursively merges the results of function expressions into the `slots` object

## Testing

Tests added

## Docs

Bug fix only

### DIFF
--- a/.changeset/hungry-spoons-crash.md
+++ b/.changeset/hungry-spoons-crash.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fix handling of `slot` attribute used inside of expressions
--- a/.changeset/hungry-spoons-crash.md
+++ b/.changeset/hungry-spoons-crash.md
@@ -1,5 +1,5 @@
 ---
-'@astrojs/compiler': patch
+'@astrojs/compiler': minor
 ---
 
 Fix handling of `slot` attribute used inside of expressions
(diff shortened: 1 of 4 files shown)


tony-sull (approved):
Nice find!  I would have thought that would be a much larger fix to handle this properly 🚀

Jutanium:
Happy Dan ❤️

PR #448, opened 2022-07-06
Handle nested expressions edge case

natemoo-re:
## Changes

- Discovered a bug in the course of fixing https://github.com/withastro/compiler/pull/447
- Fixes edge case with multiple nested expressions
- The tokenizer has some rough heuristics around when braces start an expression vs when they act as raw text
- Previously, `StartTagToken` and `EndTagToken` would reset these heuristics
- This PR implements a simple stack for `StartTagToken` and `EndTagToken` that determines if we're inside elements (`{` === `StartExpressionToken`) or outside of them (`{` === `TextToken`)

## Testing

Multiple test cases added

## Docs

Bug fix only

### DIFF
--- a/.changeset/eight-teachers-sniff.md
+++ b/.changeset/eight-teachers-sniff.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Properly handle nested expressions that return multiple elements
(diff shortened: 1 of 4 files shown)


tony-sull (approved):
LGTM

PR #452, opened 2022-07-08
Warn on client: usage with scripts

matthewp:
## Changes

- Adds a warning when using a `client:` directive on a script.

## Testing

- Added a test, but unable to specifically capture the warning in a test. Tried overriding console.log/warn/error, I think Go might be pushing to `stdout` directly. However you can see that the warning does occur when running the test.

## Docs

N/A


natemoo-re (approved):
LGTM!

Issue #461, opened 2022-07-13
🐛 BUG: TSX Output does not handle missing ending semicolon in frontmatter


Princesseuh:
### What version of `@astrojs/compiler` are you using?

0.18.2

### What package manager are you using?

pnpm

### What operating system are you using?

macOS

### Describe the Bug

TSX requires a semicolon before elements when there's JS code before, for instance the following code will have errors all over (see reproducible example)

```
const hello = "Hello"

<div></div>
```

Simply adding a semicolon so it becomes `const hello = "Hello";` will fix all the issues. 

The TSX output should do this automatically, [here's how the previous output handled this](https://github.com/withastro/language-tools/blob/d817923d0319741971dc96337607165c11a151d7/packages/language-server/src/plugins/typescript/astro2tsx.ts#L34-L38)

### Link to Minimal Reproducible Example

https://www.typescriptlang.org/play?strict=false&noImplicitAny=false&jsx=1#code/MYewdgziA2CmB00QHMAUAidBKAUDgPACYCWAbgHz4D0JFQA

PR #462, opened 2022-07-13
Add tests for TSX sourcemaps

Princesseuh:
## Changes

This adds some tests for the source maps generated when doing the TSX output. This also add a test for https://github.com/withastro/compiler/issues/461, though the "solution" in the test is an idealist one, I think

The expected offsets should be correct, however building failing tests is not super convenient 😅 A more extensive test suite for this is to run the tests from [the language server's compiler output PR](https://github.com/withastro/language-tools/pull/335), if the sourcemaps are correct, all tests should pass

## Testing

It's tests all the way down

## Docs

N/A

### DIFF
--- a/packages/compiler/package.json
+++ b/packages/compiler/package.json
@@ -30,6 +30,7 @@
     "./types": "./types.d.ts"
   },
   "devDependencies": {
+    "@jridgewell/trace-mapping": "^0.3.14",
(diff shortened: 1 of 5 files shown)


natemoo-re:
Closing this in favor of https://github.com/withastro/compiler/pull/518

PR #468, opened 2022-07-15
Add `devcontainer` configuration

ran-dall:
This adds the configuration for VSCode `Devcontainer` and Codespaces.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -21,16 +21,12 @@ ENV GO111MODULE=auto
 RUN bash /tmp/library-scripts/go-debian.sh "none" "/usr/local/go" "${GOPATH}" "${USERNAME}" "false" \
   && apt-get clean -y && rm -rf /var/lib/apt/lists/*
 
-# Install JS tools
-SHELL ["/bin/bash", "-c"]
-ENV BASH_ENV $HOME/.bashrc
-ENV VOLTA_HOME $HOME/.volta
-ENV PATH $VOLTA_HOME/bin:$PATH
-
-RUN curl https://get.volta.sh | bash
-RUN volta install node
-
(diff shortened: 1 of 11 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.241.1/containers/go/.devcontainer/base.Dockerfile
+
+# [Choice] Go version (use -bullseye variants on local arm64/Apple Silicon): 1, 1.18, 1.17, 1-bullseye, 1.18-bullseye, 1.17-bullseye, 1-buster, 1.18-buster, 1.17-buster
+ARG VARIANT="1.18-bullseye"
```
natemoo-re, line 4: I think `1.17` would be the safer option until we update our CI scripts and CONTRIBUTING instructions

ran-dall: @natemoo-re Thanks for the review. Appreciate it. I'm pushing it now. 

FWIW, in the future, I'd like to engineer something a little cleaner. Just sticking to the VSCode recommendations for now though.

ran-dall:
@natemoo-re Before this is ready, I think I want to get rid of `nvm` in favor of `volta` that you're already using.

ran-dall:
@natemoo-re Restarting the work on this PR. So I replaced Volta, but you still need to run `volta install node && npm install --global pnpm` to get everything set up. Maybe after https://github.com/volta-cli/volta/pull/1273, things will get nicer, but till then, I think this is as good as it's going to get.

ran-dall:
@natemoo-re Actually, I fixed it. Now it will automatically set up `node` and `pnpm@7.0.1` with `volta` automatically. 🎉

PR #469, opened 2022-07-18
Remove Lerna

ran-dall:
Remove Lerna, in favor of integrating everything into PNPM.

Signed-off-by: Randall <ran.dall@icloud.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,10 @@
   },
   "scripts": {
     "build": "make wasm",
+    "build:compiler": "pnpm --filter @astrojs/compiler run build",
     "lint": "eslint \"packages/**/*.{cjs,js,jsx,mjs,ts,tsx}\"",
     "format": "prettier -w .",
-    "prerelease": "pnpm run build",
+    "prerelease": "pnpm run build:compiler",
     "release": "changeset publish",
     "test": "tsm node_modules/uvu/bin.js packages test -i utils -i stress",
     "test:stress": "tsm packages/compiler/test/stress/index.ts",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -7,11 +7,10 @@
     "url": "https://github.com/withastro/compiler.git"
   },
   "scripts": {
-    "build": "make wasm && lerna run build --scope @astrojs/compiler",
-    "build:compiler": "lerna run build --scope @astrojs/compiler",
+    "build": "make wasm",
```
natemoo-re, line 10:
```suggestion
    "build": "make wasm",
    "build:compiler": "pnpm --filter @astrojs/compiler run build",
```

package.json
```diff
@@ -7,11 +7,10 @@
...
     "lint": "eslint \"packages/**/*.{cjs,js,jsx,mjs,ts,tsx}\"",
     "format": "prettier -w .",
-    "prerelease": "pnpm run build:compiler",
+    "prerelease": "pnpm run build",
```
natemoo-re, line 13:
```suggestion
    "prerelease": "pnpm run build:compiler",
```

natemoo-re (commented):
Nice! I think this still needs handling for the `@astrojs/compiler` package.

Also would love to remove the `lerna` dependency and update the lockfile.

package.json
ran-dall: @natemoo-re You know, I looked for like an hour for that `--filter` command. LOL

natemoo-re (approved):
Thanks @ran-dall!

PR #472, opened 2022-07-19
Remove Markdown special-casing

matthewp:
## Changes

- Removes special Markdown component parsing, need to use `is:raw` in the future.

## Testing

- Tests updated

## Docs

N/A


natemoo-re (approved):
LGTM!

I also pushed a commit to remove `data-astro-raw` support, preferring the documented `is:raw` directive.

PR #474, opened 2022-07-21
Fixing missing semicolon on TSX Frontmatter last-entries

lostra01:
## Changes

TSX Output automatically adds a semicolon to the last entry, if there's none present, fixing the issue #461 which causes errors in typescript. Instead of appending `"";` like the previous version of the compiler did, the code checks the last Frontmatter element for a semicolon, and will add one if it's missing.

Thank you [ran-dall](https://github.com/ran-dall) for starting the work on this so soon, so we could finish it quick. 

## Testing

Removed semicolon from the last line of the input in `packages/compiler/test/tsx/basic.ts`

## Docs

No docs update, bug fix only.

Closes: #461

### DIFF
--- a/.changeset/chilly-beers-help.md
+++ b/.changeset/chilly-beers-help.md
@@ -1,5 +1,0 @@
----
-'@astrojs/compiler': patch
----
-
-Add support for HTML minification by passing `compact: true` to `transform`.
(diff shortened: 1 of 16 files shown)


natemoo-re (approved):
Thanks @lostra01 and @ran-dall!

PR #477, opened 2022-07-22
Stable scoped class

natemoo-re:
## Changes

- Updates hashing algorithm to **ignore** the contents of hoisted `<style>` blocks.
- This will make style HMR significantly easier, since we won't have to update all the `class` attributes in the DOM

## Testing

Tests added. Only happens in `transform` so these are JS tests.

## Docs

Internal change only


tony-sull (approved):
Oh that's fancy, lgtm!

PR #483, opened 2022-07-27
Prevent wrapping module script with scope

matthewp:
## Changes

- New `define:vars` logic wraps the entire script with `{ }` to create a scope to prevent global errors.
- Module scripts are already scoped, and adding `{ }` around them causes errors because `import` cannot be inside of a scope.
- This now makes it so that module scripts are not wrapped.

## Testing

- test added

## Docs

N/A, bug fix.

### DIFF
--- a/.changeset/wet-socks-glow.md
+++ b/.changeset/wet-socks-glow.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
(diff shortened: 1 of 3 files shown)


tony-sull (approved):
Ah yep great catch, this was needed for styles but not module scripts :+1:

PR #497, opened 2022-08-11
Better handling for namespaced components

natemoo-re:
## Changes

- Companion PR to https://github.com/withastro/astro/pull/4272
- Updates how we match namespaced `components.with.deep.members`.
- Properly match `*` exports and `import { components } from 'file'`
- Removes `$$metadata` matcher utils which are mostly dead code paths now. With this PR, they can probably be removed from the frontend!

## Testing

Tests updated

## Docs

Bug fix only

### DIFF
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -288,11 +288,6 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions) {
 func AddComponentProps(doc *astro.Node, n *astro.Node, opts *TransformOptions) {
 	if n.Type == astro.ElementNode && (n.Component || n.CustomElement) {
 		for _, attr := range n.Attr {
-			id := n.Data
-			if n.CustomElement {
-				id = fmt.Sprintf("'%s'", id)
-			}
-
 			if strings.HasPrefix(attr.Key, "client:") {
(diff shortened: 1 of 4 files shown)


internal/transform/transform.go
```diff
@@ -288,11 +288,6 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions) {
 func AddComponentProps(doc *astro.Node, n *astro.Node, opts *TransformOptions) {
 	if n.Type == astro.ElementNode && (n.Component || n.CustomElement) {
 		for _, attr := range n.Attr {
-			id := n.Data
-			if n.CustomElement {
```
matthewp, line 292: this doesn't look related to this change, are you sure it's safe to remove this? I know we don't support using CE tag names any more but I think we still print them to go through renderComponent, right?

natemoo-re: Yep, thanks for checking! After the refactoring below, `id` is unused. We still handle CE tags here, but we're also handling Component tags the same way now (as a string)

matthewp: Oh got it, since we are not using metadata.getPath we don't need the id here at all. Thanks.

matthewp (approved):

PR #503, opened 2022-08-15
Remove css print escape for experimentalStaticExtraction

### LINKED ISSUE #502: CSS url() with space not escaped correctly with experimentalStaticExtraction

1.0.0-rc.6

none

npm

Linux

### Describe the Bug

Not sure if this can be considered a bug or deserves further investigation, but when using
`background-image: url();`

if the filename contains a non escaped space character, for example:
`background-image: url('/filename spaced.jpg');`
AND the file exists with that same filename or it is starting with that same filename until the space,
astro gives no error in dev but gives an error in build

```
11:47:22 [build] Building static entrypoints...
[@astrojs/rollup-plugin-build-css-minify] Transform failed with 1 error:
<stdin>:1:202: ERROR: Expected ")" to end URL token
 error   Transform failed with 1 error:
  <stdin>:1:202: ERROR: Expected ")" to end URL token
  File:
    <stdin>
  Code:
      1623 |   }).join("");
    > 1624 |   let error = new Error(`${text}${summary}`);
           |               ^
      1625 |   error.errors = errors;
      1626 |   error.warnings = warnings;
      1627 |   return error;
```

this can be reproduced adding the above background-image in the index.astro file in the astro basic template, adding also an image file in /public and running build. Not sure if the codesandbox example attached helps.

The error in build can be obviously avoided be removing spaces in filenames (and path i guess) or properly escaping that.

https://codesandbox.io/s/sharp-babycat-v01ttb?file=/src/pages/index.astro

### Participation

bluwy:
## Changes

Fix #502 

I may be missing something big, but it doesn't seem like CSS needs to be escaped as it's not JS 🤔 I've removed it for CSS printing for static extraction

## Testing

Added test to `css.ts`

## Docs

N/A

### DIFF
--- a/.changeset/gorgeous-zoos-think.md
+++ b/.changeset/gorgeous-zoos-think.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+Remove css print escape for experimentalStaticExtraction
--- a/internal/printer/print-css.go
+++ b/internal/printer/print-css.go
@@ -27,7 +27,7 @@ func PrintCSS(sourcetext string, doc *Node, opts transform.TransformOptions) Pri
 		for _, style := range doc.Styles {
 			if style.FirstChild != nil && strings.TrimSpace(style.FirstChild.Data) != "" {
 				p.addSourceMapping(style.Loc[0])
-				p.print(escapeText(strings.TrimSpace(style.FirstChild.Data)))
+				p.print(strings.TrimSpace(style.FirstChild.Data))
 				result.Output = append(result.Output, p.output)
 				p.output = []byte{}
 				p.addNilSourceMapping()
--- a/packages/compiler/test/static-extraction/css.ts
+++ b/packages/compiler/test/static-extraction/css.ts
@@ -8,6 +8,7 @@ const FIXTURE = `
 <style>
     .thing { color: green; }
     .url-space { background: url('/white space.png'); }
+    .escape:not(#\\#) { color: red; }
 </style>
 `;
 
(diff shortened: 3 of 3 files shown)


matthewp:
Needs a changeset.

bluwy:
Note to future self: Test this locally with Astro, if not possible, cut a separate publish tag.

bluwy:
@matthewp I linked this locally to the Astro tests, and only found one failing test. It seems to be incorrectly tested, so this PR should still be good.

```
  1) CSS
       build
         Astro Styles
           HTML and CSS scoped correctly:
     AssertionError: expected '.linked-sass:where(.astro-ON5EH4HS){c…' to include '.blue:where(.astro-CXSJVMWR){color:#b…'
      at Context.<anonymous> (file:///Users/bjorn/Work/oss/astro/packages/astro/test/0-css.test.js:50:27)
      at processImmediate (node:internal/timers:466:21)
```

```
const expected = `.blue:where(.${scopedClass}){color:#b0e0e6}.color\\\\:blue:where(.${scopedClass}){color:#b0e0e6}.visible:where(.${scopedClass}){display:block}`;
expect(bundledCSS).to.include(expected);
```

The `expected` should check for `\\` instead of `\\\\`. Can also confirm that before the bundled CSS isn't correct:

```css
.color\\:blue:where(.astro-CXSJVMWR)
```

With this PR, it's:

```css
.color\:blue:where(.astro-CXSJVMWR)
```

matthewp (approved):

PR #513, opened 2022-09-07
Include filename when calculating scope

matthewp:
## Changes

- When calculating the hash for the scope, use the `filename` and not just the source. This prevents us from having the same hash for two components that have the same scope.
- For https://github.com/withastro/astro/issues/4383

## Testing

- Wasm test added

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/lucky-steaks-doubt.md
+++ b/.changeset/lucky-steaks-doubt.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Include filename when calculating scope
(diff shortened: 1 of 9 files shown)


natemoo-re (commented):
Woudn't this make the hash non-deterministic across file systems? Can we use the root-relative URL of the component?

matthewp:
Ah yeah, that's a good point. Can we assume that root-relative will always work? npm packages are inside of root presumably... People might create virtual modules to which will not be root relative.

Perhaps the compiler should take an optional `id` property and the Vite plugin is responsible for making it unique (which root-relative working in most cases).

matthewp:
I updated this to receive a moduleId. A follow up in core will be needed to pass this through. It's optional here and will produce the same thing as before until that change in core occurs.

natemoo-re (approved):
LGTM!

PR #515, opened 2022-09-07
Handle bad CSS better

matthewp:
## Changes

- Currently the compiler expects preprocessStyle to return valid CSS. However if there is an error preprocessing it might now.
- This makes it so that preprocessStyle can return an error. The compiler uses that to know the CSS is bad.
- The styleError is returned in the transform result and the caller can use that error to do whatever it needs.
- Part of https://github.com/withastro/astro/issues/4615, still need core update to completely fix.

## Testing

Test added

## Docs

N/A


natemoo-re (approved):
LGTM! Not sure what would happen with one-to-many styles, but if you're not worried about that, this is a good enough solution for now.

matthewp:
Ok, I made it return an error of style errors instead.

PR #518, opened 2022-09-22
TSX Sourcemaps

natemoo-re:
## Changes

- Gets `convertToTSX` sourcemaps working
- Adds test coverage for TSX sourcemaps
- Lays the foundation for better JS sourcemaps/test coverage
- Supercedes https://github.com/withastro/compiler/pull/462

## Testing

Lots of tests added

## Docs

N/A


Princesseuh:
Living feedback document here:

## Source map issues
- Everything seems to be off by one

## Output issues
- It'd be great if the types were exported (`TSXResult`). Also it seems like the type is wrong, `map` is an object not a string
- Filenames don't seems to be processed correctly, no matter what I pass, I get the same result: `object__AstroComponent_`

FredKSchott:
I'll leave this for @Princesseuh and you to review, just saying how exciting it is to see this PR!

cmd/astro-wasm/astro-wasm.go
```diff
@@ -202,15 +202,22 @@ func ConvertToTSX() interface{} {
 		transformOptions := makeTransformOptions(js.Value(args[1]), "XXXXXX")
 
 		var doc *astro.Node
-		doc, err := astro.Parse(strings.NewReader(source))
+		doc, err := astro.Parse(strings.NewReader(strings.TrimRightFunc(source, unicode.IsSpace)))
```
matthewp, line 205: what is this doing (just curious, not questioning)

matthewp: if this is trimming spaces could that mess with locations?

internal/printer/print-to-tsx.go
```diff
@@ -60,86 +63,95 @@ func getTextType(n *astro.Node) TextType {
 func renderTsx(p *printer, n *Node) {
 	// Root of the document, print all children
 	if n.Type == DocumentNode {
+		hasChildren := false
 		for c := n.FirstChild; c != nil; c = c.NextSibling {
+			if c.PrevSibling != nil && c.PrevSibling.Type == FrontmatterNode {
```
matthewp, line 68: It's unclear to me what this block of code is doing, could you add a comment?

matthewp (approved):
lgtm, nothing blocking but a couple of comments. nice tests!

cmd/astro-wasm/astro-wasm.go
natemoo-re: This just trims any trailing newlines from the source code before we process it. This avoid an extra `TextNode` at the end of the file. Doesn't break any mapping AFAICT.

internal/printer/print-to-tsx.go
natemoo-re: Sure thing!

natemoo-re: Added in [3432c8e](https://github.com/withastro/compiler/pull/518/commits/3432c8e4223ab5b86cdc605983ae01e0d1d6b81f)

jasikpark:
should `packages/compiler/deno/astro.wasm` be `.gitignore`'d?

natemoo-re:
Yeah good call @jasikpark, we needed it for deploying the [`astro_compiler`](https://deno.land/x/astro_compiler) Deno package, but it's not maintained so we can probably drop it. Will do that in another PR.

jasikpark:
oh, I just meant that build artifacts should be git-ignored :p

do you mean that the wasm had to be commited to the git repo?

natemoo-re:
> do you mean that the wasm had to be commited to the git repo?
Yep, to be included in the Deno package, that particular build artifact had to be committed. My guess is that the publishing workflow is more flexible than last time I tried, but we're not maintaining it, so no need to try to get it working.

Princesseuh (approved):
Passes the language-server's test suite! 🥳

image

Remaining fixes (if there's any) can definitely be tackled in separate PRs. Awesome work Nate!

matthewp:
Wow, what a big effort, congrats @natemoo-re @Princesseuh

PR #525, opened 2022-09-28
Fix: CSS order for imported CSS vs. Astro styles

matthewp:
## Changes

- Brings back https://github.com/withastro/compiler/pull/520
  - was reverted because it caused components with no frontmatter to not have their imports printed.
- This fixes that

## Testing

- Tests added,
  - testing that components without frontmatter have their css import.
  - testing that components with frontmatter imports come before the component style import.

## Docs

N/A, bug fix


natemoo-re (approved):
LGTM!

PR #532, opened 2022-09-29
Ensure that head injection occurs if there is only a hoisted script

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/4770
- If there is only a `<script>` and nothing else, we need to do head injection.

## Testing

- Printer test added

## Docs

N/A, bug fix


natemoo-re (approved):
LGTM!

PR #534, opened 2022-09-30
Remove experimental Deno package from compiler (unmaintained)

natemoo-re:
## Changes

- We last published [`astro_compiler`](https://deno.land/x/astro_compiler) to Deno 6 months ago. It's not maintained and nobody is using it (to the best of my knowledge).
- We had to commit the `astro.wasm` file as an artifact so the Deno package could use it. This was causing lots of headaches while rebasing https://github.com/withastro/compiler/pull/518. Better to just remove the package since it's not maintained

## Testing

N/A

## Docs

N/A, QoL improvement


jasikpark:
could not find it on deno.land 🤷‍♀️

jasikpark (commented):
lgtm

PR #540, opened 2022-10-04
Handle `slot` before <head>

### LINKED ISSUE #531: 🐛 BUG: [AST] `head` element inside slot is removed

0.23.4

pnpm

macOS

### Describe the Bug

The following code AST does not include the `head` element:

```
<slot>
   <head>hello</head>
</slot>
```

This cause our Prettier plugin to create the following result:

```
<slot />hello
```

See bug description

natemoo-re:
## Changes

- Closes #531
- Fixes edge case where `slot` content was not picked up

## Testing

Go test added

## Docs

N/A, bug fix only

### DIFF
--- a/.changeset/nervous-ducks-fetch.md
+++ b/.changeset/nervous-ducks-fetch.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+handle slots that contains the head element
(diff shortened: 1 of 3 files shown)


matthewp (approved):
kind of a weird thing to want to do but i guess it's ok if the parser allows it

PR #546, opened 2022-10-05
Support multi-line export hoisting

### LINKED ISSUE #536: Transform fails when the opening bracket of an interface is not on the same line as the `interface` keyword

1.1.3

None

npm

Linux

### Describe the Bug

I encountered this bug while declaring a `Props` interface that extended multiple interfaces. Because their names were long, prettier formatted it so the interface declaration spanned multiple lines, kind of like this:
```ts
interface Props
    extends Interface1, Interface2 {
    // ...interface properties
}
```

But this caused the following error:

> Transform failed with 1 error:
/home/projects/github-2nzrgw/src/components/Card.astro:29:0: ERROR: Expected "{" but found "const"

I thought it had to do with `extends` or the fact that I was using generics in my code, but after playing with it for a bit in StackBlitz I realized that it has to do with the opening bracket (`{`) not being on the same line as the `interface` keyword. 

```ts
// if we do this transform fails
interface Props extends Interface1, Interface2 
{
    // ...interface properties
}

// but if we do this then everything works fine
interface Props extends Interface1, Interface2 {
    // ...interface properties
}
```

https://stackblitz.com/edit/github-2nzrgw?file=src/components/Card.astro

### Participation

natemoo-re:
## Changes

- Closes #536
- Handles edge case with multi-line export hoisting
- Ensures types and interfaces are handled similarly to functions

## Testing

Test added

## Docs

Bug fix only

### DIFF
--- a/internal/js_scanner/js_scanner.go
+++ b/internal/js_scanner/js_scanner.go
@@ -86,6 +86,10 @@ outer:
 					if isKeyword(nextValue) && next != js.FromToken {
 						continue
 					}
+					// ensure type declarations are defined or exported from package
+					if flags["type"] && !(flags["="] || flags["from"]) {
+						continue
+					}
 					if !foundIdent {
 						foundIdent = true
 					}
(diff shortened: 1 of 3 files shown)


internal/js_scanner/js_scanner.go
```diff
@@ -86,14 +86,18 @@ outer:
 					if isKeyword(nextValue) && next != js.FromToken {
 						continue
 					}
+					// ensure type declarations are defined or exported from package
+					if flags["type"] && !(flags["="] || flags["from"]) {
```
matthewp, line 90: what is the `from` for?

natemoo-re: This is valid TypeScript, apparently
```
export type Value from "package"
```

matthewp (approved):

Issue #499, opened 2022-08-12
🐛 BUG: TSX output could type `Astro.props` automatically


Princesseuh:
### What version of `@astrojs/compiler` are you using?

N/A

### What package manager are you using?

N/A

### What operating system are you using?

N/A

### Describe the Bug

It'd be great if the TSX output would automatically type `Astro.props` with the types from the user's Props interface. This is a feature we always wanted to do and it's finally possible, so why not

Something like 

`declare const Astro: { props: Props } & Omit<import('astro').AstroGlobal, 'props'>`

In the file should do the trick. It does create a weird Go To Definition behaviour, but GTD's behaviour on the Astro global is already weird because it goes into `env.d.ts` and not `astro.d.ts` like you'd expect. That's something we could fix in the language server manually or make it point to the Props interface etc

(Mainly making an issue for myself to remember, I can implement this)

### Link to Minimal Reproducible Example

N/A

natemoo-re:
Note: #518 did NOT implement this, I was mistaken. Definitely something we can add, though.

PR #547, opened 2022-10-05
Auto type `Astro.props` interface

natemoo-re:
## Changes

- Closes https://github.com/withastro/compiler/issues/499
- Implements automatic typing for `Astro.props`
- Does not support generics, but that's okay

## Testing

Tests updated

## Docs

N/A, this is how it currently works

### DIFF
--- a/internal/printer/print-to-tsx.go
+++ b/internal/printer/print-to-tsx.go
@@ -64,6 +64,10 @@ func getTextType(n *astro.Node) TextType {
 func renderTsx(p *printer, n *Node) {
 	// Root of the document, print all children
 	if n.Type == DocumentNode {
+		props := js_scanner.GetPropsType([]byte(p.sourcetext))
+		if props.Ident != "Record<string, any>" {
+			p.printf(`declare const Astro: { props: %s } & Omit<import('astro').AstroGlobal, 'props'>`, props.Ident)
+		}
 		hasChildren := false
 		for c := n.FirstChild; c != nil; c = c.NextSibling {
 			// This checks for the first node that comes *after* the frontmatter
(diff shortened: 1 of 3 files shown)


internal/printer/print-to-tsx.go
```diff
@@ -64,6 +64,10 @@ func getTextType(n *astro.Node) TextType {
 func renderTsx(p *printer, n *Node) {
 	// Root of the document, print all children
 	if n.Type == DocumentNode {
+		props := js_scanner.GetPropsType([]byte(p.sourcetext))
+		if props.Ident != "Record<string, any>" {
+			p.printf(`declare const Astro: { props: %s } & Omit<import('astro').AstroGlobal, 'props'>`, props.Ident)
```
Princesseuh, line 69: It's a super unfortunate behavior, but the JSDoc comment for both the Astro const and the props attributes needs to be included or they get lost

What I'm thinking we could do on the Astro side is a type for the Astro global that takes a generic to type the props, that way we wouldn't need the Omit or to include the JSDoc, thoughts?

natemoo-re: Yep, that makes a lot of sense to me!

Princesseuh: Made a PR for it: https://github.com/withastro/astro/pull/5000

internal/printer/print-to-tsx.go
```diff
@@ -64,6 +64,10 @@ func getTextType(n *astro.Node) TextType {
 func renderTsx(p *printer, n *Node) {
 	// Root of the document, print all children
 	if n.Type == DocumentNode {
+		props := js_scanner.GetPropsType([]byte(p.sourcetext))
+		if props.Ident != "Record<string, any>" {
+			p.printf(`declare const Astro: import('astro').AstroGlobal<%s>`, props.Ident)
```
Princesseuh, line 69:
```suggestion
			p.printf(`declare const Astro: Readonly<import('astro').AstroGlobal<%s>>`, props.Ident)
```

Princesseuh, line 69: Not doing a suggestion since I figure you don't necessarily want it in the string itself, but the JSDoc comment for the Astro global itself needs to be there, that one we can't work around unfortunately (as far as I know, at least):

```
/**
 * Astro global available in all contexts in .astro files
 *
 * [Astro documentation](https://docs.astro.build/reference/api-reference/#astro-global)
*/
declare const Astro ...
 ```

Princesseuh (approved):
Looks absolutely wonderful to me!

PR #548, opened 2022-10-06
Copies `snapshot-release.yml` from withastro/astro

jasikpark:
Hopefully this can just work out of the box! I think there might be a couple of necessary secrets to set up but otherwise we might be good :thumbsup:

## Changes

- Adds the `snapshot-release.yml` workflow, which will publish a snapshot release of a branch when `!preview snapshot-name` is posted in a PR comment. This requires that the comment was posted by an admin of the repo

## Testing

Not sure how to test except for live 🤔

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -1,0 +1,101 @@
+name: Create a Snapshot Release
+
+on:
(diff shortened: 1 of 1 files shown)


natemoo-re (approved):
Thanks so much for doing this!

PR #558, opened 2022-10-10
chore: upgraded tooling and dependencies

JuanM04:
## Changes

- Upgraded pnpm and go dependencies. Also, pnpm was upgraded to a newer version.
- Node.js wasn't upgraded because the minumum version tested in Astro is v14
- Go wasn't upgraded because upgrading breaks everything 🙃
- Simplified the Devcontainer: it now uses as the images given from Microsoft instead of using our own
- Replaced `tsm` with `tsx` because the former doesn't work in some enviroments and the latter is better[*](https://discord.com/channels/830184174198718474/845430950191038464/1028431736788492369)


matthewp:
Just a heads up that there might be a reason why we aren't on Go 1.18 yet. @natemoo-re is off for the next couple of days but then will be back, he can probably confirm that it's ok to upgrade.

JuanM04:
@matthewp Yeah, I was testing that. At least, pnpm dependencies can be updated

natemoo-re (approved):

PR #561, opened 2022-10-12
Fix parse crashing when a component with a client directive was unfound

Princesseuh:
## Changes

Since 0.27.0, `parse` would crash on the following input:

```astro
{
	// For best cross-browser support of sticky or fixed elements, they must not be nested
	// inside elements that hide any overflow axis. The article content hides `overflow-x`,
	// so we must place the mobile TOC here.
	headers && (
		<nav class="mobile-toc">
			<TableOfContents
				client:media="(max-width: 72em)"
				headers={headers}
				labels={{ onThisPage: t('rightSidebar.onThisPage'), overview: t('rightSidebar.overview') }}
				isMobile={true}
			/>
		</nav>
	)
}
```

I eventually tracked this down to [this line](https://github.com/withastro/compiler/blob/38b0ea446bcbdcb1b69be5f8108136d9e23e7b49/internal/transform/transform.go#L446), `doc.FirstChild.FirstChild` was `nil`. Adding a if check fixed the issue and made tests pass. 

Not sure if there's a deeper issue or not (it seems weird that we even go in that code on this input considering the frontmatter check?), but it works

## Testing

Added a test

## Docs

N/A

### DIFF
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -442,7 +442,7 @@ func resolveIdForMatch(match *ImportMatch, opts *TransformOptions) string {
 }
 
 func eachImportStatement(doc *astro.Node, cb func(stmt js_scanner.ImportStatement) bool) {
-	if doc.FirstChild.Type == astro.FrontmatterNode {
+	if doc.FirstChild.Type == astro.FrontmatterNode && doc.FirstChild.FirstChild != nil {
 		source := []byte(doc.FirstChild.FirstChild.Data)
 		loc, statement := js_scanner.NextImportStatement(source, 0)
 		for loc != -1 {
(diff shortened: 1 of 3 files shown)


internal/transform/transform.go
```diff
@@ -442,7 +442,7 @@ func resolveIdForMatch(match *ImportMatch, opts *TransformOptions) string {
 }
 
 func eachImportStatement(doc *astro.Node, cb func(stmt js_scanner.ImportStatement) bool) {
-	if doc.FirstChild.Type == astro.FrontmatterNode {
+	if doc.FirstChild.Type == astro.FrontmatterNode && doc.FirstChild.FirstChild != nil {
```
natemoo-re, line 445: Yep, this is the correct fix! Empty frontmatter is always generated, so we need to check if it has a child or not before proceeding.

natemoo-re (approved):
Nice fix!

Issue #563, opened 2022-10-12
🐛 BUG: [TSX] Sourcemap and output issues


Princesseuh:
### What version of `@astrojs/compiler` are you using?

0.27.0

### What package manager are you using?

pnpm

### What operating system are you using?

macOS

### Describe the Bug

The TSX output is very close, but it still has a few errors:

# Output

## Unclosed tags generate a weird output
```astro
---
const myMarkdown = await import('../content/post.md');
---

<myMarkdown.
```

```js
const myMarkdown = await import('../content/post.md');

<Fragment>
<
></
>
</Fragment>
export default function NpmImports__AstroComponent_(_props: Record<string, any>): any {}
```

I'd expect it to generate something like:

```html
<Fragment>
<myMarkdown.
</Fragment>
```

## Semicolons

The semi-colon handling is really good, but it turns out that it fails in a very common case:

```astro
---
const { hello } = Astro.props
---

<div class={hello}></div>
```

I am not sure at this time what the best heuristic is for this, I think for now maybe doing again what the previous implement did (so `"";` after the frontmatter) is best?

# Sourcemap

## Windows 

Mapping seems to be off on Windows, this cause [the `language-server`'s test suite to fail](https://github.com/withastro/language-tools/actions/runs/3200193024/jobs/5226809047)

## Attributes are weirdly mapped

Seems like the `"` is considered part of the ending tag? Also the first `"` maps to both quotes

```astro
---
---

<div class=""></div>
```

<img width="920" alt="image" src="https://user-images.githubusercontent.com/3019731/195395484-b310f320-7d51-4029-9664-c05317c662e8.png">

[...]

PR #564, opened 2022-10-12
Fix TSX Issues

natemoo-re:
## Changes

Fixes issues outlined in https://github.com/withastro/compiler/issues/563.

This isn't everything, but it's definitely NWTWWHB!

- [x] Unclosed tag issue is confirmed, only happens with trailing whitespace in file. Added a test!
- [x] Semicolon issue is confirmed, fixed
- [x] Attribute mapping issue is confirmed, fixed
- Windows issue is confirmed, will need to follow-up with a fix

## Testing

Wasm tests added

## Docs

Bug fixes only

### DIFF
--- a/.changeset/tough-bananas-bathe.md
+++ b/.changeset/tough-bananas-bathe.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Improvements to TSX output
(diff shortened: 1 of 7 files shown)


Princesseuh (approved):
Can confirm that this is not worst than what we had before, still pass the language server and fixed the reported issues! Awesome work

PR #568, opened 2022-10-14
Remove non-existent sourcemap source

natemoo-re:
## Changes

- `astro:runtime` was introduced in the TSX sourcemap PR, but this was causing issues at runtime since Vite would log warnings about a missing sourcefile
- This PR removes the `astro:runtime` source (which did not exist)

## Testing

Manually, fixes the warning

## Docs

N/A


bluwy (approved):
I forgot to `git pull` that's why I can't find it 😅 Thanks for the fix!

PR #574, opened 2022-10-18
Add `resolvePath` option

bluwy:
## Changes

Add `resolvePath` option so we can control how hydration paths are resolved. This removes hacks made previously to undo the compiler path changes.

The Astro repo has [a branch](https://github.com/withastro/astro/tree/resolve-path-option) I'm working on integrating this PR, which is working nicely too.

## Testing

Added a simple test for the new option.

## Docs

It doesn't look like we have documentation for the optional options. I could send a PR to update the readme was this new option stabilizes, documenting along the other options.

### DIFF
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -4,14 +4,12 @@ import (
 	"fmt"
 	"net/url"
 	"strings"
-	"syscall/js"
 	"unicode"
 
 	astro "github.com/withastro/compiler/internal"
 	"github.com/withastro/compiler/internal/handler"
 	"github.com/withastro/compiler/internal/js_scanner"
 	"github.com/withastro/compiler/internal/loc"
-	wasm_utils "github.com/withastro/compiler/internal_wasm/utils"
 	a "golang.org/x/net/html/atom"
 )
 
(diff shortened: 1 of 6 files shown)


matthewp:
Assuming this is the last place we're using the Metadata object, we can remove the logic for it in the compiler. But we can do that in a later PR. This looks good to me.

matthewp (approved):

internal/transform/transform.go
```diff
@@ -425,20 +426,25 @@ func trimExtension(pathname string) string {
...
 }
 
-func resolveIdForMatch(match *ImportMatch, opts *TransformOptions) string {
-	if strings.HasPrefix(match.Specifier, ".") && len(opts.Pathname) > 0 {
+func ResolveIdForMatch(id string, opts *TransformOptions) string {
+	// Try custom resolvePath if provided
+	if opts.ResolvePath != nil {
+		return opts.ResolvePath(id)
+	}
+	// Else use default resolvePath
```
matthewp, line 434: Comment on how we can remove the rest of this code in the future.

bluwy: Added in 4ccccec

PR #577, opened 2022-10-20
Remove metadata export for `resolvePath`

bluwy:
## Changes

Remove unused code, e.g.

```js
import * as $$module1 from './foo/bar'
export const $$metadata = $$createMetadata(...)
```

...when `resolvePath` option is provided. Metadata isn't used for anything else other than `$$metadata.resolvePath` which is only called if no `resolvePath` is provided.

The main Astro repo also isn't using the `$$metadata` export for anything else. I've tested this on [this branch](https://github.com/withastro/astro/tree/resolve-path-option).

The removal of `import * as` helps to remove [this patch](https://github.com/withastro/astro/blob/9d195b0921b446a80e0c4c893b6cf29bef14312d/packages/astro/src/core/build/vite-plugin-css.ts#L227-L257) which fixes a bug with `.css?raw` imports. 

---

This PR sprinkles if conditions around 😬 but once `resolvePath` is solidified, we can remove the rest of unused code per [this comment](https://github.com/withastro/compiler/blob/307c83dd038e3ae5ce7adaf0818d2defb3ec76fd/internal/transform/transform.go#L435-L438).

## Testing

Updated `preserve.ts` test

## Docs

N/A

### DIFF
--- a/.changeset/thin-trains-remember.md
+++ b/.changeset/thin-trains-remember.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': minor
+---
+
+Remove metadata export if `resolvePath` option provided
--- a/internal/printer/print-to-js.go
+++ b/internal/printer/print-to-js.go
@@ -102,7 +102,7 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 	// Root of the document, print all children
 	if n.Type == DocumentNode {
 		p.addNilSourceMapping()
-		p.printInternalImports(p.opts.InternalURL)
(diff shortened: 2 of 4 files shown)


natemoo-re (approved):
Nice refactor, this looks great!

matthewp:
This does remove all of those `import *`s that we create as well, right? That has caused bugs in the past so it would be nice to remove those assuming they are not needed (which they were only needed for the metadata).

bluwy:
Yeah this removes all those `import *` too 👍

PR #580, opened 2022-10-21
TSX: fix spread attribute printing/mapping

natemoo-re:
## Changes

- Fixes issue with spread attribute printing and mappings

## Testing

Test added

## Docs

N/A, bug fix


Princesseuh (approved):
Looks good! 
image

PR #581, opened 2022-10-21
TSX: improve self-closing tag handling

natemoo-re:
## Changes

- Self-closing tags were mapped to an end-tag that didn't exist.
- Now self-closing tags are maintained.
- Small mapping tweaks as well

## Testing

New test added, old tests updated

## Docs

Bug fix


Princesseuh (approved):
Everything is in the right place, wonderful!

image

PR #583, opened 2022-10-21
Fix utf8 compat

natemoo-re:
## Changes

We had some UTF8 compatibility issues with our output. The problem was that we were appending `[]rune` (`utf8` characters) to our `output` buffer but it was treated as if the values were `[]byte` instead. This fixes that by always converting to `[]byte` first before appending characters to `output`. There are a few other refactors for clarity.

## Testing
Adds [this wonderful code](https://mathiasbynens.be/notes/javascript-identifiers) as a test case for our output

## Docs

N/A

### DIFF
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -23,7 +23,6 @@ type printer struct {
 	sourcetext         string
 	opts               transform.TransformOptions
 	output             []byte
-	runes              []rune
 	builder            sourcemap.ChunkBuilder
 	handler            *handler.Handler
 	hasFuncPrelude     bool
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -23,6 +23,7 @@ type printer struct {
 	sourcetext         string
 	opts               transform.TransformOptions
 	output             []byte
+	runes              []rune
 	builder            sourcemap.ChunkBuilder
 	handler            *handler.Handler
 	hasFuncPrelude     bool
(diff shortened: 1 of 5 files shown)


internal/printer/printer.go
```diff
@@ -52,31 +52,30 @@ var BACKTICK = "`"
 var styleModuleSpecExp = regexp.MustCompile(`(\.css|\.pcss|\.postcss|\.sass|\.scss|\.styl|\.stylus|\.less)$`)
 
 func (p *printer) print(text string) {
-	p.output = append(p.output, text...)
-}
-
-func (p *printer) printRune(c rune) {
-	p.output = append(p.output, byte(c))
+	p.output = append(p.output, []byte(text)...)
```
natemoo-re, line 55: Converting to `[]byte(text)` first will handle `rune` properly (which is a utf8 string)

internal/printer/printer.go
```diff
@@ -52,31 +52,30 @@ var BACKTICK = "`"
...
 
 func (p *printer) printf(format string, a ...interface{}) {
-	p.output = append(p.output, []byte(fmt.Sprintf(format, a...))...)
+	p.print(fmt.Sprintf(format, a...))
 }
 
 func (p *printer) println(text string) {
-	p.output = append(p.output, (text + "\n")...)
+	p.print(text + "\n")
 }
```
natemoo-re, line 64: Consolidate all these functions to call `print` internally.

internal/printer/printer.go
```diff
@@ -52,31 +52,30 @@ var BACKTICK = "`"
...
 
 func (p *printer) printTextWithSourcemap(text string, l loc.Loc) {
 	start := l.Start
-	for _, c := range text {
+	lastPos := -1
+	for pos, c := range text {
+		diff := pos - lastPos
 		if c == '\r' {
-			start++
+			start += diff
```
natemoo-re, line 72: Fixes sourcemap issue by maintaining `rune` length

internal/printer/printer.go
natemoo-re: Basically `text...` will append `rune`, but we always need `byte` here.

Princesseuh (approved):
Things are in the right place and all the language-server's tests pass!

image

Code makes sense to me too!

Issue #575, opened 2022-10-19
🐛 BUG: Missing `position.end` for `type: element` AST node since 0.26.0


bluwy:
### What version of `@astrojs/compiler` are you using?

0.27.2

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

For `type: 'element'` AST nodes, the `position` property used to have `start` and `end` props. Since 0.26.0, the `end` prop is missing.

The stackblitz repro has JSON files comparing the AST before and after 0.26.0.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/node-847xhk?file=index.js,result-0.25.2.json

Princesseuh:
This is currently blocking the Prettier plugin. While we generally don't care about positions in AST, this blocks supporting `prettier-ignore` comments as we need to know the original position of the node to print it raw. 

Since it seems like this is related to end tags, perhaps it's related to [the issues we're having in the TSX output](https://github.com/withastro/compiler/issues/571)?

PR #588, opened 2022-10-24
Include end location in literal mode

natemoo-re:
## Changes

- Closes https://github.com/withastro/compiler/issues/575
- Includes end location when `inLiteralIM`, which happens for `parse()`

## Testing

Added a test

## Docs

Bug fix only

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -2678,6 +2678,7 @@ func inLiteralIM(p *parser) bool {
 		}
 		return true
 	case EndTagToken:
+		p.addLoc()
 		p.oe.pop()
 		return true
 	case StartExpressionToken:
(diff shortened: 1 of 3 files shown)


internal/parser.go
```diff
@@ -2678,6 +2678,7 @@ func inLiteralIM(p *parser) bool {
 		}
 		return true
 	case EndTagToken:
+		p.addLoc()
```
natemoo-re, line 2681: This was the problem. Needed to attach the location!

packages/compiler/test/parse/position.ts
```diff
@@ -0,0 +1,15 @@
+import { test } from 'uvu';
+import * as assert from 'uvu/assert';
+import { parse } from '@astrojs/compiler';
+
+test('include start and end positions', async () => {
```
Princesseuh, line 5: This test doesn't actually test for a start, it probably should!

Princesseuh (approved):
Awesome!!

PR #589, opened 2022-10-24
Ensure text/comment nodes have end positions

natemoo-re:
## Changes

- Fixes `parse` AST to include end positions for `text` and `comment` nodes.
- Unblocks https://github.com/withastro/prettier-plugin-astro/pull/297

## Testing

Test added

## Docs

N/A, bug fix only


Princesseuh (approved):
Good catch!

PR #591, opened 2022-10-25
Move `end` position for elements

natemoo-re:
## Changes

- Internally, the end position of elements is the beginning of the end tag.
- For our public AST, it should be the final index of the end tag.
- This updates our AST to use the final index of the end tag for element nodes.

## Testing

Tested manually

## Docs

Bug fix only


Princesseuh (approved):
Definitely makes more sense to me too! Would be curious to see if certain projects would prefer a different form

Issue #595, opened 2022-10-27
🐛 BUG: Wrong severity is returned for warnings


Princesseuh:
### What version of `@astrojs/compiler` are you using?

^0.28.0

### What package manager are you using?

pnpm

### What operating system are you using?

macOS

### Describe the Bug

Warnings are currently returned with the Error severity, this cause every warnings to show up as an error in the editor:

![image](https://user-images.githubusercontent.com/3019731/198305346-badae25e-bedd-4236-884c-dbf7cdb27786.png)

### Link to Minimal Reproducible Example

N/A

PR #596, opened 2022-10-27
fix(diagnostics): return proper diagnostic level

natemoo-re:
## Changes

- Closes https://github.com/withastro/compiler/issues/595
- Fix issue with diagnostic severity levels, would always return as errors (`1`)

## Testing

Test updated

## Docs

Bug fix only


Princesseuh (approved):
Awesome!

PR #599, opened 2022-10-28
Use an IIFE to scope define:vars scripts

matthewp:
## Changes

- Safari hoists function declarations out of a block scope, https://bugs.webkit.org/show_bug.cgi?id=179698
- This means that our use of block scope breaks any define:vars script (in Safari) that uses a function declaration.
- Part of https://github.com/withastro/astro/issues/5064
- Fix is to use an IIFE instead.

## Testing

Tests updated.

## Docs

N/A

### DIFF
--- a/.changeset/seven-nails-compete.md
+++ b/.changeset/seven-nails-compete.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Use an IIFE for define:vars scripts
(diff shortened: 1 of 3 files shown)


tony-sull (commented):
Nice, that fix is a heck of a lot cleaner than I expected given the bug it fixes!

The fix and tests look good to me, leaving it open for Nate's approval though in case there are compiler edge cases hidden here

natemoo-re (approved):
I was worried about `import`/`export` but we're already skipping `module` scripts, so this looks good!

PR #601, opened 2022-10-31
Improve TSX sourcemaps on Windows

natemoo-re:
## Changes

Removes some inherited logic that was specific to Windows, since it was actually causing this bug 🙃 

## Testing

Adds Windows-style `\r\n` tests

## Docs

Bug fix only

### DIFF
--- a/internal/sourcemap/sourcemap.go
+++ b/internal/sourcemap/sourcemap.go
@@ -209,7 +209,6 @@ func (offset *LineColumnOffset) AdvanceBytes(bytes []byte) {
 		case '\r', '\n', '\u2028', '\u2029':
 			// Handle Windows-specific "\r\n" newlines
 			if c == '\r' && len(bytes) > 0 && bytes[0] == '\n' {
-				columns++
 				continue
 			}
 
@@ -235,7 +234,6 @@ func (offset *LineColumnOffset) AdvanceString(text string) {
 		case '\r', '\n', '\u2028', '\u2029':
 			// Handle Windows-specific "\r\n" newlines
 			if c == '\r' && i+1 < len(text) && text[i+1] == '\n' {
-				columns++
 				continue
 			}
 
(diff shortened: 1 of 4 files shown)


packages/compiler/test/tsx-sourcemaps/template-windows.ts
```diff
@@ -0,0 +1,123 @@
...
+  assert.equal(output, {
+    source: 'index.astro',
+    line: 2,
+    column: 1,
+    name: null,
+  });
+});
+
+test('template expression has dot', async () => {
+  const input = `<div>{\nconsole.log(hey)\n}</div>`;
```
Princesseuh, line 18: This test still uses `\n`, is this intended?

packages/compiler/test/tsx-sourcemaps/template-windows.ts
```diff
@@ -0,0 +1,123 @@
...
+  assert.equal(output, {
+    source: 'index.astro',
+    line: 2,
+    column: 9,
+    name: null,
+  });
+});
+
+test('template expression with addition', async () => {
+  const input = `{"hello" + \nhey}`;
```
Princesseuh, line 40: Same comment as higher here

internal/sourcemap/sourcemap.go
```diff
@@ -209,7 +209,6 @@ func (offset *LineColumnOffset) AdvanceBytes(bytes []byte) {
 		case '\r', '\n', '\u2028', '\u2029':
 			// Handle Windows-specific "\r\n" newlines
 			if c == '\r' && len(bytes) > 0 && bytes[0] == '\n' {
```
Princesseuh, line 211: I'm not 100% sure I follow the logic of this, is special behaviour needed for Windows because the generated code always uses `\n`? Otherwise, couldn't it be mapped 1:1?

packages/compiler/test/tsx-sourcemaps/template-windows.ts
natemoo-re: Yes! Checking that the outputs are the same.

internal/sourcemap/sourcemap.go
natemoo-re: Yes, I'm pretty sure the previous logic was only correct if the whole file was printed with `\r\n`, but as you correctly point out, we only print `\n`.

Princesseuh (approved):

PR #610, opened 2022-11-04
Hack Windows sourcemap offset

natemoo-re:
## Changes

- Hack sourcemap logic to confirm if Windows is just an off-by-one issue (somewhere else)

## Testing

Prerelease

## Docs

N/A


natemoo-re:
!preview windows

PR #612, opened 2022-11-07
Add warning when unable to inject `define:vars`

natemoo-re:
## Changes

- Closes https://github.com/withastro/astro/issues/4400
- Adds a warning when `<style define:vars>` doesn't work because of limitations of static analysis.

## Testing

Test added

## Docs

Bug fix only


matthewp (approved):

Princesseuh (approved):
Very excited to start seeing those show inside people's editors!

PR #620, opened 2022-11-11
fix: infinite loop regression in literal (parse/ast) mode

natemoo-re:
## Changes

- [Literal Insertion Mode handling](https://github.com/withastro/compiler/pull/617) introduced a regression for consumers using `parse`. Only caught this when `format` on Astro's main repo was hanging.
- This refactors `inLiteralIM` to ensure that the mode is exited properly when tokens are unhandled.
- Blocks https://github.com/withastro/prettier-plugin-astro/pull/305, https://github.com/withastro/astro/pull/5369

## Testing

Tested locally in the `prettier-plugin-astro` repo, no more hanging

## Docs

Bug fix only

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -2672,16 +2672,22 @@ func frontmatterIM(p *parser) bool {
 func inLiteralIM(p *parser) bool {
 	shouldExit := p.exitLiteralIM()
 	switch p.tok.Type {
-	case ErrorToken:
-		// Stop parsing.
-	case TextToken:
-		p.addText(p.tok.Data)
 	case StartTagToken:
 		p.addElement()
 		if p.hasSelfClosingToken {
 			p.oe.pop()
 			p.acknowledgeSelfClosingTag()
 		}
+		// always continue `inLiteralIM`
+		return true
+	case StartExpressionToken:
+		p.addExpression()
+		// always continue `inLiteralIM`
+		return true
+	case ErrorToken:
+		// Stop parsing.
+	case TextToken:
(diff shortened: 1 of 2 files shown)


internal/parser.go
```diff
@@ -2670,36 +2670,38 @@ func frontmatterIM(p *parser) bool {
...
-		case EndTagToken:
-			p.addLoc()
-			p.oe.pop()
-			return true
-		case StartExpressionToken:
-			p.addExpression()
-			return true
-		case EndExpressionToken:
-			p.addLoc()
+	shouldExit := p.exitLiteralIM()
```
natemoo-re, line 2673: Immediately check if we should exit after this token is handled

internal/parser.go
```diff
@@ -2670,36 +2670,38 @@ func frontmatterIM(p *parser) bool {
...
+		p.addLoc()
+		p.oe.pop()
 	}
-	p.im = p.originalIM
-	p.originalIM = nil
-	return p.tok.Type == EndTagToken
+	if shouldExit {
+		p.im = p.originalIM
+		p.originalIM = nil
+	}
```
natemoo-re, line 2703: If we should exit, do it! Resets to the previous insertion mode

internal/parser.go
```diff
@@ -2670,36 +2670,38 @@ func frontmatterIM(p *parser) bool {
...
+	return true
```
natemoo-re, line 2704: Otherwise, consider this token handled and continue `inLiteralIM`

matthewp (approved):

PR #625, opened 2022-11-14
Self-closing tags now generate end position

crutchcorn:
## Changes

- What does this change?
- Be short and concise. Bullet points can help!
- Before/after screenshots can be helpful as well.

While attempting to use `prettier-plugin-astro` and self-closing HTML tags, I ran into this issue:

https://github.com/withastro/prettier-plugin-astro/issues/306

Here, we can see that our self-closing tag is not receiving an `end` positional data, unlike other tags.

To do this, we simply had to `push` a `loc` data into the `SelfClosingTag` token.

We can see a similar issue that occurred in:

https://github.com/withastro/compiler/pull/588

Which was fixed in a similar manner.

## Testing

There are tests added in `position.ts` and I've verified this behavior in a manual Go debugger instance.

## Docs

No docs are need, this was only a bug fix.

### DIFF
--- a/.changeset/heavy-mice-begin.md
+++ b/.changeset/heavy-mice-begin.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Self-closing tags will now retreive "end" positional data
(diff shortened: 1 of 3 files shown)


crutchcorn:
CC: @aFuzzyBear, per his request

natemoo-re (commented):
Looks good! I'd love to double check this change against the `language-tools` repo to make sure there aren't any unintended side-effects from this.

Will keep you posted!

natemoo-re (approved):
LGTM! Thanks for the PR!

PR #627, opened 2022-11-14
fix regression for inLiteralIM handling

natemoo-re:
## Changes

- Will address https://github.com/withastro/astro/issues/5379
- `inLiteralIM` was being handled properly, but we forgot to set `p.originalIM` to `inHeadIM`, which was causing `body` tags to be disregarded

## Testing

Test added

## Docs

Bug fix only

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -686,6 +686,7 @@ func beforeHTMLIM(p *parser) bool {
 		}
 		if p.literal {
 			p.im = inLiteralIM
+			p.originalIM = beforeHTMLIM
 			return false
 		}
 	case EndTagToken:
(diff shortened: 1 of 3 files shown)


internal/parser.go
```diff
@@ -784,10 +785,8 @@ func inHeadIM(p *parser) bool {
 		// Allow components in Head
 		if isComponent(p.tok.Data) || isFragment(p.tok.Data) {
 			p.im = inLiteralIM
-			oe := len(p.oe)
-			p.exitLiteralIM = func() bool {
-				return len(p.oe) == oe
-			}
+			p.originalIM = inHeadIM
```
natemoo-re, line 788: This is the main fix

matthewp (approved):

PR #640, opened 2022-11-16
fix out of bounds error

jasikpark:
this was found via the fuzzer in #619 

## Changes

-  Fixes an out of bounds error accessing a slice

## Testing

Not sure how to test it

I want
```json
{
	name:   "unknown",
	source: `<A { 0>`,
	want:   "error, malformed template tag",
},
```
to be a valid "errors" test case for TestScopeHTML, but currently the compiler doesn't error for that input.. should I just check in the output as "expected"?


natemoo-re (commented):
> to be a valid "errors" test case for TestScopeHTML, but currently the compiler doesn't error for that input.. should I just check in the output as "expected"?

Yep, I would just add a test with the expected output. Great catch, fuzzing FTW!

jasikpark:
ok, I've pushed a test - could probably give it a more effective name

natemoo-re (approved):

PR #642, opened 2022-11-20
Remove `astro.ParseFragment` / Require an error handler for `astro.ParseFragmentWithOptions`

jasikpark:
## Changes

- Removes `astro.ParseFragment` in favor of `astro.ParseFragmentWithOptions`
- Updates `astro.ParseFragmentWithOptions` to return an error when called w/o an error handler. Should it panic instead?

## Testing

Might be worth adding a "expect error" test for `astro.ParseFragmentWithOptions` though idk where this function is tested, other than transitively via `scope_html_test.go`

### DIFF
--- a/.changeset/nine-ladybugs-exist.md
+++ b/.changeset/nine-ladybugs-exist.md
@@ -1,0 +1,7 @@
+---
+'@astrojs/compiler': patch
+---
+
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):
Nice! Thanks for the refactor.

jasikpark:
am i able to merge this w/ the approve?

PR #645, opened 2022-11-23
Include the id in createComponent

matthewp:
## Changes

- Passes the module's `moduleId` as the second argument to createComponent.
- This is used to enable head propagation. The Astro runtime creates a dependency map so that it can trace modules with head propagation to what pages they belong to, so that we wait for those components to run before sending out the head.
- This is an optional argument and is backwards compatible.

## Testing

- Test added to confirm that the passed in moduleId is include as the second argument.
- Head propagation tests are in core.

## Docs

N/A

### DIFF
--- a/.changeset/sharp-tools-knock.md
+++ b/.changeset/sharp-tools-knock.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': minor
+---
+
+Provide the moduleId of the astro component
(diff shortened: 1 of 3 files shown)


matthewp:
!preview bring-id

matthewp:
!preview module-id

matthewp:
!preview module-id

natemoo-re (approved):

bholmesdev (approved):
`Mod`efinitely a good change ✅

PR #646, opened 2022-11-29
fix: avoid nil pointer dereference in table parsing

wtchnm:
## Changes

- What does this change? 
Avoid nil pointer dereference in `inTableIM`, fixes https://github.com/withastro/astro/issues/5488.

## Testing

I couldn't create a failing test, the bug only happens when the table is inside a slot so I tested it manually.

## Docs

Bug fix.

### DIFF
--- a/.changeset/giant-chairs-rest.md
+++ b/.changeset/giant-chairs-rest.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+fix: avoid nil pointer dereference in table parsing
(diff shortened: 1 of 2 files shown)


wtchnm:
!preview table-nil

jasikpark (commented):
looks solid, i imagine we'd want to add a test fixture tho, maybe @natemoo-re knows where that'd fit in?

natemoo-re (approved):
I also wasn't able to get a failing test for some reason, but this looks like a straightforward change! 

Thanks for the PR.

PR #651, opened 2022-12-02
Replace internal logs with properly handled warnings

### LINKED ISSUE #647: 🐛 BUG: Unclosed style tag crash the compiler

0.29.17

pnpm

macOS

### Describe the Bug

When encountering an unclosed style tag, the compiler crashes with the following stacktrace:

image

This is easy to reproduce in the TSX output in the editor, as writing `<style>` will instantly crash the extension. This is not specific to the TSX output though, as an error also shows up in Astro when doing that:
image

File with just `<style>` will do it

natemoo-re:
## Changes

- Closes #647
- Replaces `fmt.Printf` statements in the tokenizer with proper warnings for unexpected characters. 
- Gracefully handles EOF now

## Testing

Test added

## Docs

Bug fix only

### DIFF
--- a/packages/compiler/test/bad-styles/unclosed-style.ts
+++ b/packages/compiler/test/bad-styles/unclosed-style.ts
@@ -4,12 +4,17 @@ import { parse } from '@astrojs/compiler';
 
 test('can compile unfinished style', async () => {
   let error = 0;
+  let result;
   try {
-    await parse(`<style>`);
+    result = await parse(`<style>`);
   } catch (e) {
     error = 1;
   }
+
+  const style = result.ast.children[0];
   assert.equal(error, 0, 'Expected to compile with unfinished style.');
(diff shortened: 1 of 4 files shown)


Princesseuh (approved):

packages/compiler/test/bad-styles/unclosed-style.ts
```diff
@@ -0,0 +1,15 @@
...
+import { parse } from '@astrojs/compiler';
+
+test('can compile unfinished style', async () => {
+  let error = 0;
+  try {
+    await parse(`<style>`);
+  } catch (e) {
+    error = 1;
+  }
+  assert.equal(error, 0, 'Expected to compile with unfinished style.');
```
jasikpark, line 12: wouldn't it make sense to also assert on the output if this is expected to not error?

PR #652, opened 2022-12-02
Fix combo of props and getStaticPaths

natemoo-re:
## Changes

- Fixes edge case where Props generic detection was getting tripped up.
- Runs a few tests that I'm not sure were running previously

## Testing

Adds a new test

## Docs

Bug fix only

### DIFF
--- a/internal/js_scanner/js_scanner.go
+++ b/internal/js_scanner/js_scanner.go
@@ -339,7 +339,7 @@ outer:
 		// Track our current position
 		i += len(value)
 	}
-	if len(genericsIdents) > 0 && ident != defaultPropType {
+	if start > 0 && len(genericsIdents) > 0 && ident != defaultPropType {
 		generics = fmt.Sprintf("<%s>", strings.Join(genericsIdents, ", "))
 		statement = strings.TrimSpace(string(source[start:end]))
 	}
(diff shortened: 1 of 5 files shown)


packages/compiler/test/tsx/non-latin.ts
```diff
@@ -66,3 +66,5 @@ export default function __AstroComponent_(_props: Record<string, any>): any {}`;
   const { code } = await convertToTSX(input, { sourcemap: 'external' });
   assert.snapshot(code, output, `expected code to match snapshot`);
 });
+
+test.run();
```
jasikpark, line 70: is this a requirement of uvu tests ig?

internal/js_scanner/js_scanner.go
```diff
@@ -339,7 +339,7 @@ outer:
 		// Track our current position
 		i += len(value)
 	}
-	if len(genericsIdents) > 0 && ident != defaultPropType {
+	if start > 0 && len(genericsIdents) > 0 && ident != defaultPropType {
```
jasikpark, line 342: heh? is this just saying that `Props` is a magic name

natemoo-re: Yep, `Props` is a magic name. In this case our lexing was getting tripped up because of the `Record<string, never>` and it was injecting the whole `---
type Props = Record<string, never>` string in the wrong place.

packages/compiler/test/tsx/non-latin.ts
natemoo-re: Yeah, I'm not sure if these were still running, but better safe than sorry!

Princesseuh (approved):

jasikpark (approved):

PR #665, opened 2022-12-08
Add join-base directive

matthewp:
Note that this is **just an experiment** and is not ready to be merged. If successful it will turn into an RFC.

## Changes

- Adds a new directive `join-base` that can be used on any element to join a path to your base configuration. For example:

```astro
<img join-base:src="/images/penguin.png" />
```

Will get transformed to this HTML assuming your `base` config is `"/docs/"`:

```html
<img src="/docs/images/penguin.png" />
```

## Testing

- Test case added


matthewp:
!preview join-base

matthewp:
This was an experiment we didn't go with.

Issue #660, opened 2022-11-24
Client side script inside layout file makes html lang attribute disappear


riccardolardi:
### What version of `astro` are you using?

1.6.11

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

yarn

### What operating system are you using?

Mac

### Describe the Bug

It seems using a `<script>` inside the layout which declares the `html` element will make the `lang` attribute disappear.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-xuenvq?file=src/layouts/Layout.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

Jarmos-san:
I've recently stumbled across this bug on [my website](https://jarmos.dev) as well and here's the [source code](https://github.com/Jarmos-san/jarmos.dev/blob/4da0e78353ace8a055f49c5225dc348c24c6e9c8/web/src/layouts/BaseLayout.astro#L40) particularly responsible for generating the base template. Is there any way I can debug and perhaps identify what the root cause of the issue is?

skilar:
Looks like this is caused not just by the `script` tag but also including `styles` before the `html` tag. That should help your issue, @Jarmos-san.

Jarmos-san:
@skilar that fixed the issue! Thanks for it. I'm curious, what was wrong with the code there?

skilar:
I'm not sure what the issue is, @Jarmos-san, seems like there's still a bug in Astro.

PR #666, opened 2022-12-09
Allow script/style before html

natemoo-re:
## Changes

- Allows `script` nodes before the `html` node.
- Closes https://github.com/withastro/compiler/issues/660
- This was an edge case in our parser that was causing the actual `html` node to be ignored

## Testing

Wasm test added

## Docs

Bug fix only

### DIFF
--- a/packages/compiler/test/basic/script-before-html.ts
+++ b/packages/compiler/test/basic/script-before-html.ts
@@ -1,0 +1,41 @@
+import { test } from 'uvu';
+import * as assert from 'uvu/assert';
+import { transform } from '@astrojs/compiler';
+
+const FIXTURE = `
+<script>
+  // anything ...
+</script>
+
+<html lang="de">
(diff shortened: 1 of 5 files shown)


packages/compiler/test/basic/script-before-html.ts
```diff
@@ -0,0 +1,41 @@
+import { test } from 'uvu';
+import * as assert from 'uvu/assert';
+import { transform } from '@astrojs/compiler';
+
+const FIXTURE = `
+<script>
+  // anything ...
+</script>
+
+<html lang="de">
```
jasikpark, line 10: maybe worth adding custom DOCTYPE in the test here as well?

jasikpark (commented):
don't fully understand the parser code, but the test looks like it validates the bugfix 👍

jasikpark (approved):

PR #671, opened 2022-12-15
fix: corner case with slot expression in head will cause body tag missing

yuhang-dong:
## Changes

- Solve issue https://github.com/withastro/astro/issues/5557
- `Slot` in `head` tag will cause a situation that paser cannot exit from `inExpressionIM`

For file:
```js
---
const isProd = true;
---
<!DOCTYPE html>
<html lang="en">
  <head>
    <TestHead />
    <title>document</title>
    {isProd && <slot />}
  </head>
  <body style="color: red;">
    <main>
       <h1>Welcome to <span class="text-gradient">Astro</span></h1>
    </main>
  </body>
</html>
```

Before, compiler has the below output, it omits the `body` tag.
```js
const $$stdin = $$createComponent(async ($$result, $$props, $$slots) => {
const Astro = $$result.createAstro($$Astro, $$props, $$slots);
Astro.self = $$stdin;
const isProd = true;
return $$render`<html lang="en">
  <head>
    ${$$renderComponent($$result,'TestHead',TestHead,{})}
    <title>document</title>
    ${isProd && $$render`${$$renderSlot($$result,$$slots["default"])}`}
  ${$$renderHead($$result)}</head>
  
    <main>
       <h1>Welcome to <span class="text-gradient">Astro</span></h1>
    </main>
  

</html>`;
}, '<stdin>');
```

After this pr, it will like this:
```js
const $$stdin = $$createComponent(async ($$result, $$props, $$slots) => {
const Astro = $$result.createAstro($$Astro, $$props, $$slots);
Astro.self = $$stdin;
const isProd = true;
return $$render`<html lang="en">
  <head>
    ${$$renderComponent($$result,'TestHead',TestHead,{})}
    <title>document</title>
    ${isProd && $$render`${$$renderSlot($$result,$$slots["default"])}`}
  ${$$renderHead($$result)}</head>
  <body style="color: red;">
    <main>
       <h1>Welcome to <span class="text-gradient">Astro</span></h1>
    </main>
  </body></html>`;
}, '<stdin>');
```

## Testing

Adding test

## Docs

bug fix only.

PR #672, opened 2022-12-17
fix: corner case that component in head expression will case `body` tag missing

yuhang-dong:
## Changes

- The previous pr https://github.com/withastro/compiler/pull/671 has fixed the problem that `Slot` in `head` expression will cause `body` tag missing
- And I found that `Component` in `head` expression will cause the same problem. 
- Original Issue: https://github.com/withastro/astro/issues/5557 Minimal Reproducible Example: https://stackblitz.com/edit/github-uwb7te-xzq2mg?file=src%2Fcomponents%2FDocument.astro
- So, I think we should add `Component` Support in `inExpressionIM` function when it's in `head` tag.

## Testing

Added
## Docs

bug fix only

### DIFF
--- a/.changeset/ninety-emus-swim.md
+++ b/.changeset/ninety-emus-swim.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+fix: corner case that component in head expression will case body tag missing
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):
This makes sense to me! Thanks for the fix!

PR #674, opened 2022-12-19
Improve JS Sourcemaps

natemoo-re:
## Changes

- Vastly improves our JavaScript sourcemaps
- Previously, frontmatter was chopped up in a few different places and we were discarding location information
- Now, we are tracking the location of all hoisted imports, exports, and other content when restructuring frontmatter. This allows us to properly construct a sourcemap while printing.

## Testing

Tested manually using https://evanw.github.io/source-map-visualization/

https://user-images.githubusercontent.com/7118177/208511623-07a22d96-be57-496d-bb1a-6fccd72eb1dc.mp4

## Docs

N/A


packages/compiler/test/js-sourcemaps/complex-frontmatter.ts
```diff
@@ -0,0 +1,46 @@
+import { test } from 'uvu';
+import * as assert from 'uvu/assert';
+import { testJSSourcemap } from '../utils';
+
+const input = `---
+// TODO: Due to this issue: https://github.com/withastro/astro/issues/1438, this route can't be in the same folder
```
jasikpark, line 6: an aside, for <https://defined.net> we get around this by generating both paths via the same `getStaticPaths` function && adding an `isPost: true` identifier to the post entries, then in the astro body:

```astro
{props.content.isPost ?
  <BlogPost content={content} />
  : <BlogIndex content={content} />}
```

natemoo-re: Ah! I wonder if we can fix that...

This was just a copy + paste from an existing test fixture. Can't remember where it came from. https://github.com/withastro/compiler/blob/main/packages/compiler/test/tsx-sourcemaps/complex-generics.ts#L6-L8

jasikpark:
```astro
---
import BlogIndexLayout from "@layouts/BlogIndex.astro";
import BlogPostLayout from "@layouts/BlogPost.astro";
import { assert } from "assert-ts";
import type { Page, PaginateFunction } from "astro";
import type { AstroComponentFactory } from "astro/dist/runtime/server";
import type { MarkdownFrontmatter } from "@data/posts/MarkdownFrontmatter";
import type { MarkdownOrMDXInstance } from "./_getAllPublishablePosts";
import BlogMeta from "./meta.json";

export async function getStaticPaths({ paginate }: { paginate: PaginateFunction }) {
  const { getAllPublishablePosts } = await import("./_getAllPublishablePosts");
  const allPages = getAllPublishablePosts();

  // generate paths for the blog posts
  const postPaths = allPages.map(({ Content, frontmatter }: MarkdownOrMDXInstance) => {
    return {
      params: {
        page: frontmatter.slug,
      },
      props: {
        page: {
          ...frontmatter,
          Content,
          isPost: true,
        },
      },
    };
  });

  // generate paths for the blog index pages
  // @ts-expect-error the return type of `paginate` is too generic here, so we use `unknown` to shut up ts
  const paginationPaths = paginate(allPages, { pageSize: 10 }) as Page<MarkdownOrMDXInstance>[];

  return [...paginationPaths, ...postPaths];
}

type Post = MarkdownFrontmatter & { Content: AstroComponentFactory; isPost: true };
type PaginationPage = Page<MarkdownOrMDXInstance> & { isPost: undefined };
type Props = {
  page: Post | PaginationPage;
};
const { page } = Astro.props as Props;
---

{
  page.isPost ? (
    <BlogPostLayout content={page}>
      <page.Content />
    </BlogPostLayout>
  ) : (
    <BlogIndexLayout
      meta={{ title: assert(BlogMeta.title), description: assert(BlogMeta.description) }}
      page={page}
    />
  )
}
```

:D

internal/js_scanner/js_scanner.go
```diff
@@ -13,24 +13,33 @@ import (
 )
 
 type HoistedScripts struct {
-	Hoisted [][]byte
-	Body    []byte
+	Hoisted     [][]byte
+	HoistedLocs []loc.Loc
+	Body        [][]byte
+	BodyLocs    []loc.Loc
```
natemoo-re, line 19: Previously we were throwing away all location info when hoisting imports/exports, which was obviously unhelpful!

internal/printer/print-to-js.go
```diff
@@ -131,45 +130,89 @@ func render1(p *printer, n *Node, opts RenderOptions) {
...
+							}
+						}
+
+						if len(preprocessed.Body) > 0 {
+							for j, body := range preprocessed.Body {
+								bodyLoc := preprocessed.BodyLocs[j]
+								bodiesLocs = append(bodiesLocs, loc.Loc{Start: start + innerStart + bodyLoc.Start})
+								bodies = append(bodies, body)
+							}
+						}
```
natemoo-re, line 190: This is interleaving import hoisting with export hoisting, which gets kinda hard to follow, but the main point is that we're preserving the location of each statement.

internal/printer/print-to-js.go
```diff
@@ -131,45 +130,89 @@ func render1(p *printer, n *Node, opts RenderOptions) {
...
 
+				p.printFuncPrelude(opts.opts)
+				// PRINT BODY
+				if len(bodies) > 0 {
+					for i, body := range bodies {
+						bodyLoc := bodiesLocs[i]
+						if len(bytes.TrimSpace(body)) == 0 {
+							continue
+						}
+						p.printTextWithSourcemap(string(body), bodyLoc)
```
natemoo-re, line 213: Now we can print these with intact location info!

jasikpark:
woot 🥳

PR #676, opened 2022-12-20
Add internal/js_scanner/js_scanner_test.go#FuzzHoistExport

jasikpark:
## Testing

- Adds internal/js_scanner/js_scanner_test.go#FuzzHoistExport
Runnable via `go test ./internal/js_scanner -fuzz=FuzzHoistExports`

### DIFF
--- a/internal/js_scanner/js_scanner_test.go
+++ b/internal/js_scanner/js_scanner_test.go
@@ -535,7 +535,7 @@ export const foo = 0
 		{
 			name:   "gobbely-gook",
 			source: "\"\"export=",
-			want:   "\"\"export=",
+			want:   "",
 		},
 	}
 }
--- a/internal/js_scanner/js_scanner_test.go
+++ b/internal/js_scanner/js_scanner_test.go
@@ -534,8 +534,8 @@ export const foo = 0
 		},
 		{
 			name:   "gobbely-gook",
-			source: "i t n m\"\"export c n s t t o\xc6t''t b=tf",
-			want:   "i t n m\"\"export c n s t t o\xc6t''t b=tf",
+			source: "\"\"export =",
+			want:   "\"\"export =",
 		},
 	}
 }
--- a/internal/js_scanner/js_scanner_test.go
+++ b/internal/js_scanner/js_scanner_test.go
@@ -530,7 +530,7 @@ export const foo = 0
 		{
 			name:   "incomplete interface",
 			source: "import { a } from \"a\";\nexport interface ",
-			want:   "import { a } from \"a\";\nexport interface ",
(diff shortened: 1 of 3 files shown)


internal/js_scanner/js_scanner_test.go
```diff
@@ -527,7 +527,21 @@ export const foo = 0
 */`,
 			want: `export const foo = 0`,
 		},
+		{
+			name:   "incomplete interface",
+			source: "import { a } from \"a\";\nexport interface ",
```
jasikpark, line 532: not sure what the expected output would be for this

internal/js_scanner/js_scanner_test.go
```diff
@@ -527,7 +527,21 @@ export const foo = 0
...
+			want:   "",
+		},
+		{
+			name:   "gobbely-gook",
+			source: "i t n m\"\"export c n s t t o\xc6t''t b=tf",
```
jasikpark, line 537: this input causes HoistExport to hang

internal/js_scanner/js_scanner_test.go
```diff
@@ -552,6 +566,24 @@ export const foo = 0
...
+		f.Add(tt.source) // Use f.Add to provide a seed corpus
+	}
+	f.Fuzz(func(t *testing.T, source string) {
+		result := HoistExports([]byte(source))
+		got := []byte{}
+		for _, imp := range result.Hoisted {
+			got = append(got, bytes.TrimSpace(imp)...)
+			got = append(got, '\n')
+		}
+		if utf8.ValidString(source) && !utf8.ValidString(string(got)) {
```
jasikpark, line 581: not sure what additional properties I can assert on here and in `FuzzHoistExport`. Currently I'm basically just testing that no errors are thrown plus that we get unicode out

internal/js_scanner/js_scanner_test.go
```diff
@@ -552,6 +583,36 @@ export const foo = 0
...
+		got2 := []byte{}
+		for _, imp := range result2.Hoisted {
+			got2 = append(got2, bytes.TrimSpace(imp)...)
+			got2 = append(got2, '\n')
+		}
+		if utf8.ValidString(string(got)) && !utf8.ValidString(string(got2)) {
+			t.Errorf("Export hoisting produced an invalid string: %q", got2)
+		}
+		if string(got) != string(got2) {
+			t.Errorf("Hoisting exports twice should return the same result.\nsource:\n%s\nfirst pass:\n%s\nsecond pass:\n%s\n", source, got, got2)
```
jasikpark, line 611: we should get the same result if we hoist twice

PR #677, opened 2022-12-20
chore: set github action timeout to 3min / job

jasikpark:
## Changes

- All jobs consistently complete in under 3min - this should bring the generous default down to something that we can detect. CI sorta PR!


natemoo-re (approved):
Nice!

PR #681, opened 2022-12-22
Fix all the typos :D

jasikpark:
## Changes

- Fixes some typos in code and comments, I'm not sure this will pass immediately 😅

### DIFF
--- a/packages/compiler/browser/utils.ts
+++ b/packages/compiler/browser/utils.ts
@@ -102,11 +102,11 @@ function serializeAttributes(node: TagLikeNode): string {
   return output;
 }
 
-export interface SerializeOtions {
+export interface SerializeOptions {
   selfClose: boolean;
 }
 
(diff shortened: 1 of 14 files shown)


packages/compiler/browser/utils.ts
```diff
@@ -102,11 +102,11 @@ function serializeAttributes(node: TagLikeNode): string {
   return output;
 }
 
-export interface SerializeOtions {
```
jasikpark, line 105: I believe this should be the only externally facing change?

Princesseuh:
Was this done using automated tools? There was a PR recently to Astro core that fixed some typos using a tool, maybe it'd be interesting here too?

jasikpark:
uhuh! https://github.com/marketplace/actions/typos-action

great tool

natemoo-re (approved):
Thanks for this!

PR #683, opened 2022-12-27
fix: handle module IDs containing quotes

43081j:
## Changes

* Escape single quotes in module identifiers

## Testing

* Added a test to the printer's tests

## Docs

* no docs updated

---

its possible you can have a file named `foo'bar.astro` which currently will trip up the compiler, since it'd produce something along the lines of `createComponent(..., 'foo'bar')` (ofc syntactically incorrect).

let me know if this is the wrong way to fix it and i'll update

i also added an `editorconfig`, let me know if its wrong. it'd be useful to have one in this repo so contributors automatically use the right indentation etc

### DIFF
--- a/.changeset/sixty-dragons-perform.md
+++ b/.changeset/sixty-dragons-perform.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Handle module IDs containing quotes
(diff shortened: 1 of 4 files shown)


bluwy (approved):
I don't have an opinion on the `.editorconfig` addition, but putting an approval for the code changes.

natemoo-re (approved):
This change looks good to me!

PR #684, opened 2023-01-02
Fix OSS-Fuzz #54632 (failed to scope input)

jasikpark:
## Changes

Currently trying to sort out how to fix this. I assume that it should throw an error w/ this input?

The real weirdness is that `astro.ParseFragmentWithOptions` parsed it fine?

## Testing

Adds failing input to `TestScopeHTML` from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=54636&q=label:Proj-astro-compiler

### DIFF
--- a/internal/transform/scope-html_test.go
+++ b/internal/transform/scope-html_test.go
@@ -95,6 +95,11 @@ func tests() []struct {
 			source: `<A { 0>`,
 			want:   `<A  0>={0>} class="astro-XXXXXX"></A>`,
 		},
+		{
+			name:   "incomplete props spread",
+			source: "<a {...>",
+			want:   "<a {...}></a>",
+		},
 	}
 
 }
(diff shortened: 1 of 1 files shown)


jasikpark:
```go
	fmt.Printf("source: %s, nodes: %+v\n", tt.source, nodes[0])
```

```
source: <a {...>, nodes: &{Fragment:false CustomElement:false Component:false Expression:false Parent:<nil> FirstChild:<nil> LastChild:<nil> PrevSibling:<nil> NextSibling:<nil> Styles:[] Scripts:[] HydratedComponentNodes:[] HydratedComponents:[] ClientOnlyComponentNodes:[] ClientOnlyComponents:[] HydrationDirectives:map[] Type:element DataAtom:a Data:a Namespace: Attr:[{Namespace: Key:> KeyLoc:{Start:7} Val: ValLoc:{Start:8} Tokenizer:<nil> Type:spread}] Loc:[{Start:1}]}
```

seems like it just parses strange...

PR #686, opened 2023-01-03
Add test/fuzztest for ParseFragmentWithOptions

jasikpark:
## Changes

- Adds testing for [ParseFragmentWithOptions](https://github.com/withastro/compiler/blob/5bbd15991581307e9308d6b259ac7b35007de115/internal/parser.go#L3063)
- Adds <https://pkg.go.dev/github.com/stretchr/testify/assert> which may be an unwanted dep, but it's a widely liked and simple assertion library that provides nice diffs in this instance

## Testing

This sets up `TestParseFragmentWithOptions` and `FuzzParseFragmentWithOptions`. `astro.ParseFragmentWithOptions` is tested elsewhere, but I wanted to check in these parse decisions, like returning `<nil>` when provided `\x00</F></a>` as input.

Also sets up a test that parsing then printing then parsing again returns the same values


jasikpark:
This currently fails, because the parser accepts invalid syntax instead of returning an error

internal/parser.go
```diff
@@ -1553,7 +1553,7 @@ func (p *parser) inBodyEndTagFormatting(tagAtom a.Atom, tagName string) {
 	// refactor this code to be more idiomatic.
 
 	// Steps 1-2
-	if current := p.oe.top(); current.Data == tagName && p.afe.index(current) == -1 {
+	if current := p.oe.top(); current != nil && current.Data == tagName && p.afe.index(current) == -1 {
```
jasikpark, line 1556: this is the fix for the "weird control characters" test. i feel like we should error instead of returning a nil node here though?

jasikpark:
Is there a good way to detect malformed input and return an error here?

internal/parser_test.go
```diff
@@ -126,3 +129,100 @@ func findTargetNode(doc *Node) *Node {
...
+	}{
+		{
+			name:   "none",
+			source: "<div />",
+			want:   []*Node{{Type: ElementNode, DataAtom: atom.Div, Data: atom.Div.String(), Namespace: "", Loc: []loc.Loc{{Start: 1}}}},
+		},
+		{
+			name:   "fault input currently accepted",
+			source: `<A { 0>`,
+			want:   []*Node{{Component: true, Type: ElementNode, DataAtom: 0, Data: "A", Namespace: "", Attr: []Attribute{{Namespace: "", Key: " 0>", KeyLoc: loc.Loc{Start: 4}, Val: "", ValLoc: loc.Loc{Start: 7}, Tokenizer: nil, Type: ShorthandAttribute}}, Loc: []loc.Loc{{Start: 1}}}},
```
jasikpark, line 153: @natemoo-re determined that this is a faulty parse / tokenization, this should be resolved probably as an element w/ two empty attributes with names `{` and `0`?? like `<A {={true} 0={true}>`?

to give context to the failing test

jasikpark: hmm but from the test fail we can see that it gets parsed as `<A {0>}/>` as if `0>` was a valid identifier in JS, resulting in the printed output of `<A 0>={0>}/>` similar to `<A {x}/>` == `<A x={x}/>`

jasikpark: I can actually make this test pass by altering

https://github.com/withastro/compiler/blob/2dc9f533b0afbdb7a17d7e5a283b6545b7e53feb/internal/print-to-source.go#L60-L63

since currently we don't strip whitespace from the attribute name but we do from the attribute variable?

aFuzzyBear:
you cant test the fuzz....the fuzz tests you!
🐻💚

Issue #692, opened 2023-01-05
Remove createAstro unused params


bluwy:
https://github.com/withastro/astro/blob/a3a7fc9298e6d88abb4b7bee1e58f05fa9558cf1/packages/astro/src/runtime/server/astro-global.ts#L21

There are unused params with the removal of `Astro.resolve`. We should clean it up and accept `site` only.

Maybe we should also deprecate `Astro.site`? Users can access it via `import.meta.env.SITE`

PR #693, opened 2023-01-06
Remove site and projectRoot options

bluwy:
## Changes

Closes https://github.com/withastro/compiler/issues/692

Remove `site` and `projectRoot` options, in favour of `injectGlobals` option which is more generic, e.g.

```js
transform(source, {
  injectGlobals: JSON.stringify(astroConfig.site)
})
```

The compiler would output `$$createAstro("https://astro.build")`. This allows us to pass params dynamically to `createAstro` without leaking business logic that gets added to, e.g. `Astro.site`.

Further discussion on discord: https://discord.com/channels/830184174198718474/1060212452756815954/1060564993768169542

Note this change goes in hand to this Astro code: https://github.com/withastro/astro/blob/6a95790b45114e2e9fae65f513c6c9381ccdf376/packages/astro/src/runtime/server/astro-global.ts#L19-L32

## Testing

Updated the test to cover this new option.

## Docs

n/a

### DIFF
--- a/cmd/astro-wasm/astro-wasm.go
+++ b/cmd/astro-wasm/astro-wasm.go
@@ -89,15 +89,7 @@ func makeTransformOptions(options js.Value) transform.TransformOptions {
 		sourcemap = "both"
 	}
 
-	site := jsString(options.Get("site"))
-	if site == "" {
-		site = ""
-	}
-
-	projectRoot := jsString(options.Get("projectRoot"))
-	if projectRoot == "" {
-		projectRoot = "."
-	}
+	injectGlobals := jsString(options.Get("injectGlobals"))
 
 	compact := false
 	if jsBool(options.Get("compact")) {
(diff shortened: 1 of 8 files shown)


matthewp:
lgtm, you'll need to create a follow-up PR relatively soon to use this new option astro.

matthewp (approved):

bluwy:
I've created https://github.com/withastro/astro/pull/5803 👍  Running the tests locally it seems to work fine.

@natemoo-re I'd love to have your eyes on this too if you have thoughts on it.

cmd/astro-wasm/astro-wasm.go
```diff
@@ -84,15 +84,7 @@ func makeTransformOptions(options js.Value) transform.TransformOptions {
...
-	site := jsString(options.Get("site"))
-	if site == "" {
-		site = ""
-	}
-
-	projectRoot := jsString(options.Get("projectRoot"))
-	if projectRoot == "" {
-		projectRoot = "."
-	}
+	injectGlobals := jsString(options.Get("injectGlobals"))
```
natemoo-re, line 87: Only comment is around naming: Would something like `astroGlobalArgs` be more clear? `injectGlobals` sounds like it will do something with `globalThis`.

natemoo-re (approved):
LGTM! Only comment is around naming, but non-blocking and I leave the final decision up to you.

cmd/astro-wasm/astro-wasm.go
bluwy: I do like `astroGlobalArgs` more, will update to it 👍

PR #696, opened 2023-01-11
Refactor file name options

bluwy:
## Changes

- Merge `sourcefile` and `moduleId` options as a single `filename` option. (Possible since https://github.com/withastro/astro/pull/5811)
- Add a new `normalizedFilename` option to generate stable hashes instead.

---

The new scope hashes are derived like this:

1. If `normalizedFilename` is define, use it only to generate the hash.
2. Else use the entire source to generate the hash.

I don't think we need to combined both the file path and source to generate the hash, the `normalizedFilename` itself should be unique to generate off if it's define, which Astro core will always pass. The value would be like `/src/pages/index.astro` etc.

---

Should also pave the way for https://github.com/withastro/astro/issues/5136 once we are able to use the `normalizedFilename` option in core.

## Testing

- Updated the tests with the option name changes.
- Updated `printer_test` to also handle case where `filename` option is passed and used in the generated code.
- Remove "same file, same html but different style tag, but still same hash" test. I think the hash should derive from the source entirely (no excluding style) by default. This shouldn't affect us as if `normalizedFilename` is passed (which we will), the hash will derive from it instead.

## Docs

n/a

### DIFF
--- a/.changeset/green-avocados-dance.md
+++ b/.changeset/green-avocados-dance.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': minor
+---
+
+Merge `sourcefile` and `moduleId` options as a single `filename` option. Add a new `normalizedFilename` option to generate stable hashes instead.
--- a/cmd/astro-wasm/astro-wasm.go
+++ b/cmd/astro-wasm/astro-wasm.go
@@ -52,7 +52,7 @@ func makeParseOptions(options js.Value) t.ParseOptions {
 		position = pos.Bool()
 	}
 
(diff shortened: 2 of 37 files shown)


matthewp:
Regarding the hash, the biggest thing here is that `astro build` should produce the same result on different machines, so that a build doesn't unnecessarily invalidate caching. It sounds like we are achieving that by using a root relative `normalizedFilename`, is that right?

natemoo-re (approved):
This looks great! Much easier to reason about.

PR #697, opened 2023-01-12
Fix `export type` hang

natemoo-re:
## Changes

- Closes https://github.com/withastro/astro/issues/5818
- Adds better handling when exporting types
- Specifically, we might hit EOF when iterating over tokens when something isn't closed. If we do, we'll just hoist everything out.

## Testing

Test added and existing tests passing

## Docs

Bug fix only

### DIFF
--- a/internal/js_scanner/js_scanner.go
+++ b/internal/js_scanner/js_scanner.go
@@ -82,7 +82,6 @@ outer:
 			flags := make(map[string]bool)
 			foundIdent := false
 			foundSemicolonOrLineTerminator := false
-			isEOF := false
 			start := i
 			i += len(value)
 			for {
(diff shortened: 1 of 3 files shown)


internal/js_scanner/js_scanner.go
```diff
@@ -95,22 +96,31 @@ outer:
...
 						continue
 					}
 					if !foundIdent {
 						foundIdent = true
 					}
-				} else if next == js.LineTerminatorToken || next == js.SemicolonToken || (next == js.ErrorToken && l.Err() == io.EOF) {
+				} else if next == js.LineTerminatorToken || next == js.SemicolonToken || isEOF {
 					if (flags["function"] || flags["=>"] || flags["interface"]) && !flags["{"] {
-						continue
+						if !isEOF {
```
jasikpark, line 116: can these if blocks be simplified somehow or does this make the most sense?

matthewp (approved):

internal/js_scanner/js_scanner.go
natemoo-re: Oh yeah, not sure what I was thinking. Cleaned it up significantly!

jasikpark: nice job!

PR #698, opened 2023-01-12
Ignore trailing whitespace in files

natemoo-re:
## Changes

- Closes https://github.com/withastro/astro/issues/5737
- Trim trailing whitespace from file before parsing to avoid significant whitespace in the render function output

## Testing

Test added. It failed before this change, now it passes.

## Docs

N/A, bug fix only


bluwy (commented):
LGTM! But it looks like the `pnpm.lock` file is accidentally formatted.

jasikpark (approved):
lgtm

matthewp (approved):

Issue #699, opened 2023-01-13
🐛 BUG: [TSX] Interfaces with properties using generic arguments prevent component props from being typed


Princesseuh:
### What version of `@astrojs/compiler` are you using?

^0.31.3

### What package manager are you using?

pnpm

### What operating system are you using?

macOS

### Describe the Bug

As the title says, any interfaces using generic will prevent the proper types from being generated, as such the following component:
```astro
---
interface LocalImageProps {
  src: Promise<{ default: string }>;
}

type Props = LocalImageProps;
---
```
will generate the following signature: `export default function Image__AstroComponent_(_props: Record<string, any>): any {}`, instead of the proper one that I'd expect: `export default function Image__AstroComponent_(_props: Props): any {}`

### Link to Minimal Reproducible Example

See description

PR #702, opened 2023-01-13
Fix generics handling that occurs before `Props`

natemoo-re:
## Changes

- Closes https://github.com/withastro/compiler/issues/699
- Adds proper handling for generics when we haven't found `Props` yet

## Testing

All existing tests pass, one test added for edge case

## Docs

Bug fix only


jasikpark (approved):
looking great to me! fixes the extant bug

MoustaphaDev:
I guess [https://github.com/withastro/astro/pull/5788](this PR) won't be needed anymore?

natemoo-re:
> I guess [this PR](https://github.com/withastro/astro/pull/5788) won't be needed anymore?

Hmm I'm not sure! The original should work once this change makes its way into VS Code, but that might take a while

Princesseuh (approved):

PR #703, opened 2023-01-13
test(#486): add regression test

### LINKED ISSUE #486: 🐛 BUG: Compilation breaks when exporting Props interface without a line-break between {}

1.0.0-rc.1

None

npm

Mac

### Describe the Bug

Shown here, https://stackblitz.com/edit/github-hjpxzk?file=src%2Fcomponents%2FTest.astro,src%2Fpages%2Findex.astro&on=stackblitz

```astro
---
// eslint-disable-next-line no-undef
export interface Props extends astroHTML.JSX.HTMLAttributes {}

const props = { ...Astro.props } as Props;
---

<body class:list={props['class:list']}>
  <slot></slot>
</body>
```

Removing the comment or adding a newline between `{}` fixes it. So possibly 2 bugs here.

https://stackblitz.com/edit/github-hjpxzk?file=src%2Fcomponents%2FTest.astro,src%2Fpages%2Findex.astro&on=stackblitz

### Participation

natemoo-re:
## Changes

- Closes #486 
- Regression test only

## Testing

This is that

## Docs

N/A

### DIFF
--- a/packages/compiler/test/basic/props-interface.ts
+++ b/packages/compiler/test/basic/props-interface.ts
@@ -1,0 +1,27 @@
+import { test } from 'uvu';
+import * as assert from 'uvu/assert';
+import { transform } from '@astrojs/compiler';
+
+const FIXTURE = `---
+// eslint-disable-next-line no-undef
+export interface Props extends astroHTML.JSX.HTMLAttributes {}
+
+const props = { ...Astro.props } as Props;
+---
+
+<body class:list={props['class:list']}>
+  <slot></slot>
+</body>`;
+
+let result;
+test.before(async () => {
+  result = await transform(FIXTURE);
+});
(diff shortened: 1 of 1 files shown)


packages/compiler/test/basic/props-interface.ts
```diff
@@ -0,0 +1,27 @@
...
+<body class:list={props['class:list']}>
+  <slot></slot>
+</body>`;
+
+let result;
+test.before(async () => {
+  result = await transform(FIXTURE);
+});
+
+test('< and > as raw text', () => {
```
jasikpark, line 21: what does this test name mean?

jasikpark: looks like copy paste

PR #704, opened 2023-01-13
Fix accidentally failing test

natemoo-re:
## Changes

- Found a test that had a logic error and it turns out was actually failing!
- Fixed the underlying bug
- Test passes now

## Testing

This is that.

## Docs

Bug fix only

### DIFF
--- a/packages/compiler/test/basic/lt-gt-text.ts
+++ b/packages/compiler/test/basic/lt-gt-text.ts
@@ -40,7 +40,7 @@ test.before(async () => {
 
 test('< and > as raw text', () => {
   assert.ok(result.code, 'Expected to compile');
-  assert.ok(result.diagnostics.length, 0, 'Expected no diagnostics');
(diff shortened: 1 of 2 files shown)


packages/compiler/test/basic/lt-gt-text.ts
```diff
@@ -40,7 +40,7 @@ test.before(async () => {
 
 test('< and > as raw text', () => {
   assert.ok(result.code, 'Expected to compile');
-  assert.ok(result.diagnostics.length, 0, 'Expected no diagnostics');
```
jasikpark, line 43: wacky that uvu doesn't complain about extra args - https://github.com/lukeed/uvu/blob/master/docs/api.assert.md#okactual-t-msg-message

jasikpark (commented):
dunno about how the bugfix works, but the test assertion fix looks good!

Princesseuh (approved):

PR #717, opened 2023-02-06
Pass through result to slot

matthewp:
## Changes

- Adds the `resultScopedSlot` option. Using this option causes the slot to receive the `$$result` object. This is so that we can provide scoped result objects within the context of slots. This is used so that if head injection occurs we can conditionally on render head contents when we should; for example not to render head contents when a JSX component using an Astro component, but itself is used within a layout.
- Ideally we can remove this option in the future if we remove implicit head injection in a major version.

## Testing

- Test for the option added

## Docs

N/A

### DIFF
--- a/.changeset/pink-maps-sip.md
+++ b/.changeset/pink-maps-sip.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': minor
+---
+
+Allow passing through result to slot call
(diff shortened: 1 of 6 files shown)


matthewp:
!preview slot-pass

matthewp:
!preview slot-pass

natemoo-re (approved):
LGTM!

PR #721, opened 2023-02-07
Better handling for invalid attributes in TSX

Princesseuh:
## Changes

The fateful day came where writing an actual solution for this instead of doing some clumsy search was the better solution. Hopefully that covers everything now!

## Testing

Added a test!

## Docs

N/A


natemoo-re (approved):
🤯 Erika wrote Go!

Looks great!

PR #732, opened 2023-02-20
Preserve namespaced attributes when using expressions

MoustaphaDev:
## Changes

Closes https://github.com/withastro/astro/issues/6316

Updates namespaced attribute logic to ensure `namespace:` prefix is always included

## Testing

Regression test added, all existing tests pass

## Docs

N/A, bug fix

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -1669,7 +1669,7 @@ import { Container, Col, Row } from 'react-bootstrap';
 			name:   "Signal when using namespaced attributes with expression values",
 			source: `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><rect xlink:href={` + BACKTICK + `#${iconId}` + BACKTICK + `}></svg>`,
 			want: want{
-				code: `${$$maybeRenderHead($$result)}<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><rect xlink:${$$addAttribute(` + BACKTICK + `#${iconId}` + BACKTICK + `, "href", true)}></rect></svg>`,
+				code: `${$$maybeRenderHead($$result)}<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><rect xlink:${$$addAttribute(` + BACKTICK + `#${iconId}` + BACKTICK + `, "href", false)}></rect></svg>`,
 			},
 		},
 		{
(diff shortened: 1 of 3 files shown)


MoustaphaDev:
!preview namespaced-attribute-nospace

MoustaphaDev:
!preview namespaced-attribute-nospace

MoustaphaDev:
preview namespaced-attribute-nospace

internal/printer/printer_test.go
```diff
@@ -1665,6 +1665,13 @@ import { Container, Col, Row } from 'react-bootstrap';
 				code: `${$$maybeRenderHead($$result)}<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><rect xlink:href="#id"></rect></svg>`,
 			},
 		},
+		{
+			name:   "Signal when using namespaced attributes with expression values",
+			source: `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><rect xlink:href={` + BACKTICK + `#${iconId}` + BACKTICK + `}></svg>`,
+			want: want{
+				code: `${$$maybeRenderHead($$result)}<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><rect xlink:${$$addAttribute(` + BACKTICK + `#${iconId}` + BACKTICK + `, "href", undefined, false)}></rect></svg>`,
```
bluwy, line 1672:
```suggestion
				code: `${$$maybeRenderHead($$result)}<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><rect ${$$addAttribute(` + BACKTICK + `#${iconId}` + BACKTICK + `, "xlink:href")}></rect></svg>`,
```

Shouldn't the output be ideally something like this? Where the key is `xlink:href`, instead of `href`.

MoustaphaDev: Ideally yes, and that would avoid making a change to the `addAttribute` in the runtime.
But currently the compiler specificaly passes only the `href` when it sees a namespaced attribute belonging to a foreign element (svg,...), so there may be a reason behind it that I don't understand yet

bluwy: I'm not sure why too, maybe @natemoo-re knows checking the git blame 👀

natemoo-re: Yep, I'm looking into this now!

natemoo-re: OK it was just a bug in the namespace logic. Should be fixed by my commit with no change to `astro` itself!

matthewp (approved):

PR #734, opened 2023-02-22
fix: quoted `set` directives to not inlined as text node

MoustaphaDev:
## Changes

Fix https://github.com/withastro/compiler/issues/735

Makes a text node the value of `set:html` and `set:text` directives when using quoted or template literal attributes

## Testing

Added tests to check if `set` directives using quoted or template literal attributes compile to text nodes

## Docs

bug fix only

### DIFF
--- a/.changeset/tall-mails-impress.md
+++ b/.changeset/tall-mails-impress.md
@@ -2,4 +2,4 @@
 '@astrojs/compiler': patch
 ---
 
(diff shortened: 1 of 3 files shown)


MoustaphaDev:
!preview

MoustaphaDev:
!preview set-directive-fix

MoustaphaDev:
!preview set-directive-fix

natemoo-re (approved):
Thanks for the fix!

PR #738, opened 2023-02-25
Add `<noscript/>` to allowed head elements

jasikpark:
## Changes

- Adds `<noscript>` to the allowed head elements
- This prevents injecting Astro assets too early in the head

## Testing

unsure how to test, maybe a fixure like the one in https://discord.com/channels/830184174198718474/1078802894050054144?

## Docs

Bugfix, docs unneeded


jasikpark:
hmm i may have been naive, this should be considered based on its child content, right?

matthewp:
what's the test that's failing? It might just need to be updated

jasikpark:
you think? i wasn't sure since it seemed to test for the bug is was experiencing maybe

jasikpark:
closig this since it's pretty stale and i've forgotten what it was trying to solve

Issue #737, opened 2023-02-24
🐛 BUG: define:vars on style fails when the root element has a style attribute


alexpdraper:
### What version of `@astrojs/compiler` are you using?

1.1.2

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

Using `<style define:vars>` in an Astro component when the root element has a style attribute as well works in unexpected ways. It can sometimes cause an error, or it can result in a style attribute like `style="style --bgColor: tomato;"`.

```astro
---
const bgColor = 'tomato';
---

<section style="color: cornsilk;">
  <p>This text should be cornsilk coloured on a tomato background!</p>
</section>

<style define:vars={{ bgColor }}>
	section {
		background-color: var(--bgColor);
	}
</style>
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-hycdvy?file=src/pages/index.astro

PR #741, opened 2023-02-28
Fix `define:vars` when used with a `style` attribute

MoustaphaDev:
## Changes

Fix https://github.com/withastro/compiler/issues/737

## Testing

Added test for when the `style` attribute is:
- A shorthand
- An expression
- Empty
- A template literal
- Quoted

## Docs

We might need to document semicolon usage depending of the resolution taken in [this comment](https://github.com/withastro/compiler/pull/741#pullrequestreview-1318082743)

### DIFF
--- a/internal/transform/scope-html.go
+++ b/internal/transform/scope-html.go
@@ -54,7 +54,7 @@ func injectDefineVars(n *astro.Node, values []string) {
 			switch attr.Type {
 			case astro.ShorthandAttribute:
 				attr.Type = astro.ExpressionAttribute
-				attr.Val = fmt.Sprintf(`%s + %s`, attr.Key, definedVars)
+				attr.Val = fmt.Sprintf("`${%s}; ${%s}`", attr.Key, definedVars)
 				n.Attr[i] = attr
 				return
 			case astro.EmptyAttribute:
--- a/internal/transform/scope-html.go
+++ b/internal/transform/scope-html.go
@@ -64,17 +64,17 @@ func injectDefineVars(n *astro.Node, values []string) {
 				return
 			case astro.QuotedAttribute:
 				attr.Type = astro.ExpressionAttribute
-				attr.Val = fmt.Sprintf("`${\"%s\"} ${%s}`", attr.Val, definedVars)
+				attr.Val = fmt.Sprintf("`${\"%s\"}; ${%s}`", attr.Val, definedVars)
 				n.Attr[i] = attr
 				return
 			case astro.TemplateLiteralAttribute:
 				attr.Type = astro.ExpressionAttribute
-				attr.Val = fmt.Sprintf("`${`%s`} ${%s}`", attr.Val, definedVars)
+				attr.Val = fmt.Sprintf("`${`%s`}; ${%s}`", attr.Val, definedVars)
 				n.Attr[i] = attr
 				return
 			case astro.ExpressionAttribute:
 				attr.Type = astro.ExpressionAttribute
-				attr.Val = fmt.Sprintf("`${%s} ${%s}`", attr.Val, definedVars)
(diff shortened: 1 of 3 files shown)


internal/transform/scope-html.go
```diff
@@ -62,14 +62,19 @@ func injectDefineVars(n *astro.Node, values []string) {
...
-			case astro.QuotedAttribute, astro.TemplateLiteralAttribute:
+			case astro.QuotedAttribute:
+				attr.Type = astro.ExpressionAttribute
+				attr.Val = fmt.Sprintf("`${\"%s\"} ${%s}`", attr.Val, definedVars)
+				n.Attr[i] = attr
+				return
+			case astro.TemplateLiteralAttribute:
 				attr.Type = astro.ExpressionAttribute
-				attr.Val = fmt.Sprintf("`%s ${%s}`", attr.Key, definedVars)
+				attr.Val = fmt.Sprintf("`${`%s`} ${%s}`", attr.Val, definedVars)
```
MoustaphaDev, line 72:
```suggestion
				attr.Val = fmt.Sprintf("`${`%s`}; ${%s}`", attr.Val, definedVars)
```

internal/transform/scope-html.go
```diff
@@ -62,14 +62,19 @@ func injectDefineVars(n *astro.Node, values []string) {
...
 				n.Attr[i] = attr
 				return
 			case astro.ExpressionAttribute:
 				attr.Type = astro.ExpressionAttribute
-				attr.Val = fmt.Sprintf("(%s) + %s`", attr.Val, definedVars)
+				attr.Val = fmt.Sprintf("`${%s} ${%s}`", attr.Val, definedVars)
```
MoustaphaDev, line 77:
```suggestion
				attr.Val = fmt.Sprintf("`${%s}; ${%s}`", attr.Val, definedVars)
```

MoustaphaDev (commented):
When a trailing semicolon isn't added in the `style` attribute, the resulting attribute value is wrong.
To make it so that never happens, whether or not a trailing semicolon is added, we could just add one anyway, that'd be perfectly valid.

internal/transform/scope-html.go
MoustaphaDev, line 72:
```suggestion
				attr.Val = fmt.Sprintf("`${`%s`}; ${%s}`", attr.Val, definedVars)
```

internal/transform/scope-html.go
MoustaphaDev, line 77:
```suggestion
				attr.Val = fmt.Sprintf("`${%s}; ${%s}`", attr.Val, definedVars)
```

MoustaphaDev (commented):
Currently, when a trailing semicolon isn't added in the `style` attribute, the resulting value is wrong.
To make it valid when a trailing semicolon is added or not, we could just add one anyway, that is perfectly valid syntax.
I think we should go that route, otherwise we should at least document that

internal/transform/scope-html.go
```diff
@@ -62,14 +62,19 @@ func injectDefineVars(n *astro.Node, values []string) {
 				attr.Val = definedVars
 				n.Attr[i] = attr
 				return
-			case astro.QuotedAttribute, astro.TemplateLiteralAttribute:
+			case astro.QuotedAttribute:
+				attr.Type = astro.ExpressionAttribute
+				attr.Val = fmt.Sprintf("`${\"%s\"} ${%s}`", attr.Val, definedVars)
```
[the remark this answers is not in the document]
bluwy: This makes sense to me, I think it's good to safe guard this case with a semicolon always.

PS: I did a review and the rest of the changes lgtm!

bluwy (approved):

ematipico (approved):

PR #744, opened 2023-03-02
Fix ignored `form` elements after a `form` element containing an expression

MoustaphaDev:
## Changes

Fix https://github.com/withastro/compiler/issues/743

The form pointer element wasn't reset when going from `inExpressionIM` to `textIM` at the end of an expression inside a form element, which caused the following `form` elements to be ignored.

## Testing

- Added a tokenizer test for a sanity check
- Added a printer test

## Docs

bug fix only

### DIFF
--- a/.changeset/tame-tables-taste.md
+++ b/.changeset/tame-tables-taste.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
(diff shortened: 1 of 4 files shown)


natemoo-re (approved):
Nice catch, thank you! 🎉

PR #753, opened 2023-03-16
Add metadata for head propagation

matthewp:
## Changes

- Adds metadata on whether the tree includes a head. This is used on the Astro side to know if the implicit head injection algorithm can be ignored. It should be in most cases.

## Testing

- Added wasm tests

## Docs

N/A, bug fix


matthewp:
!preview head-metadata

natemoo-re (approved):

Issue #755, opened 2023-03-20
🐛 BUG: Self-closing tag breaks transformation in nested expressions


space-otter:
### What version of `@astrojs/compiler` are you using?

1.2.1

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

The self-closing tag optimization in 1.2.1 breaks expression after self-closing tags.

For example: 

```astro
{() => {
  return <div><div />{"expression"}</div>
}}
```

Results in: 

```html
<div>
  "{"expression"}"
</div>
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ff1z1d?file=src/pages/index.astro

PR #740, opened 2023-02-27
Improve handling of self-closing tags returned from expression

MoustaphaDev:
## Changes

Fix https://github.com/withastro/compiler/issues/739

## Testing

Added a tokenizer and a printer test

## Docs

bug fix only

PR #756, opened 2023-03-20
Add test for self-closing tags before expressions

space-otter:
## Changes

Add test for tokenizer evaluating nested expressions after self-closing tags.
See #755 for related bug.

Closes https://github.com/withastro/astro/issues/6589

## Testing

`go test`

## Docs

Just added test.

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -2750,6 +2750,14 @@ const items = ["Dog", "Cat", "Platipus"];
 				code: `${$$renderComponent($$result,'Layout',Layout,{})}`,
 			},
 		},
+		{
+			name:     "complex recursive component",
+			source:   `{(<Fragment><Fragment set:html={` + BACKTICK + `<${Node.tag} ${stringifyAttributes(Node.attributes)}>` + BACKTICK + `} />{Node.children.map((child) => (<Astro.self node={child} />))}<Fragment set:html={` + BACKTICK + `</${Node.tag}>` + BACKTICK + `} /></Fragment>)}`,
+			filename: "/projects/app/src/components/RenderNode.astro",
+			want: want{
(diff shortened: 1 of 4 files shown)


space-otter:
As in issue #755 described the test should fail.

internal/printer/printer_test.go
```diff
@@ -2750,6 +2750,14 @@ const items = ["Dog", "Cat", "Platipus"];
...
 			},
 		},
+		{
+			name:     "complex recursive component",
+			source:   `{(<Fragment><Fragment set:html={` + BACKTICK + `<${Node.tag} ${stringifyAttributes(Node.attributes)}>` + BACKTICK + `} />{Node.children.map((child) => (<Astro.self node={child} />))}<Fragment set:html={` + BACKTICK + `</${Node.tag}>` + BACKTICK + `} /></Fragment>)}`,
+			filename: "/projects/app/src/components/RenderNode.astro",
+			want: want{
+				code: `${($$render` + BACKTICK + `${$$renderComponent($$result,'Fragment',Fragment,{},{"default": () => $$render` + BACKTICK + `${$$renderComponent($$result,'Fragment',Fragment,{},{"default": () => $$render` + BACKTICK + `${$$unescapeHTML(` + BACKTICK + `<${Node.tag} ${stringifyAttributes(Node.attributes)}>` + BACKTICK + `)}` + BACKTICK + `,})}${Node.children.map((child) => ($$render` + BACKTICK + `${$$renderComponent($$result,'Astro.self',Astro.self,{"node":(child)})}` + BACKTICK + `))}${$$renderComponent($$result,'Fragment',Fragment,{},{"default": () => $$render` + BACKTICK + `${$$unescapeHTML(` + BACKTICK + `</${Node.tag}>` + BACKTICK + `)}` + BACKTICK + `,})}` + BACKTICK + `,})})` + BACKTICK + `}`,
```
bholmesdev, line 2758: Trusting the number of parens and closing braces are correct here! This was causing the parse error originally. Do the compiler tests ensure this is valid JS?

natemoo-re: Nope, not validating that it's valid JS, just that the output is correct. This was the output prior to https://github.com/withastro/compiler/pull/740. I verified that this was failing prior to this PR and passing with it.

PR #768, opened 2023-04-03
Implement scopedStyleStrategy

matthewp:
## Changes

- Implements compiler changes needed for https://github.com/withastro/roadmap/pull/543
- Mostly this is just threading a config option down to the CSS printer so that it doesn't include `:where` when the `class` option is provided.

## Testing

- New css printing test.

## Docs

N/A in the compiler, to be documented on the docs site.

### DIFF
--- a/internal/transform/scope-css.go
+++ b/internal/transform/scope-css.go
@@ -41,10 +41,15 @@ func ScopeStyle(styles []*astro.Node, opts TransformOptions) bool {
 		if n.FirstChild == nil || strings.TrimSpace(n.FirstChild.Data) == "" {
 			continue
 		}
+		scopeStrategy := 1
+		if opts.ScopedStyleStrategy == "class" {
+			scopeStrategy = 2
+		}
+
 		// Use vendored version of esbuild internals to parse AST
 		tree := css_parser.Parse(logger.Log{AddMsg: func(msg logger.Msg) {}}, logger.Source{Contents: n.FirstChild.Data}, css_parser.Options{MinifySyntax: false, MinifyWhitespace: true})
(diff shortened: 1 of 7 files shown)


matthewp:
!preview scopedStyleStrategy

internal/transform/scope-css.go
```diff
@@ -41,10 +41,15 @@ func ScopeStyle(styles []*astro.Node, opts TransformOptions) bool {
 		if n.FirstChild == nil || strings.TrimSpace(n.FirstChild.Data) == "" {
 			continue
 		}
+		scopeStrategy := 1
+		if opts.ScopedStyleStrategy == "class" {
+			scopeStrategy = 2
+		}
```
natemoo-re, line 47: Should these use an actual `type` instead of integers?

```go
// A NodeType is the type of a Node.
type ScopeStrategy uint32

const (
	ScopeStrategyWhere ScopeStrategy = iota
	ScopeStrategyClass
)
```

lib/esbuild/css_printer/css_printer.go
```diff
@@ -37,6 +37,7 @@ type Options struct {
 	AddSourceMappings bool
 	LegalComments     config.LegalComments
 	Scope             string
+	ScopeStrategy     int
```
natemoo-re, line 40: Could also update this type to use the enum

```suggestion
	ScopeStrategy     ScopeStrategy
```

natemoo-re (commented):
Looks good, would prefer an actual enum to an int, but otherwise good!

matthewp:
!preview scopedStyleStrategy

natemoo-re (approved):

PR #770, opened 2023-04-06
Preserve whitespace in slots

matthewp:
## Changes

- Preserves whitespace in slots
- Takes care to only add whitespace when there are other elements, as the intended behavior is that a component with no children should not have whitespace added.
- This is part of (but does not fix) https://github.com/withastro/astro/issues/6561

## Testing

- Had to update a bunch of them

## Docs

N/A

### DIFF
--- a/.changeset/smart-ties-punch.md
+++ b/.changeset/smart-ties-punch.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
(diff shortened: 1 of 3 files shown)


matthewp:
!release slot-stream

matthewp:
!preview slot-stream

natemoo-re (approved):
Looks great!

PR #772, opened 2023-04-09
feat(sync): Add a sync and CJS entrypoint for editor tooling

Princesseuh:
## Changes

TypeScript unfortunately does not support running async methods inside its LanguageService API, so our language server needs a sync version of the compiler. Since WASM instantiation is async only, you still need at least one async init call, but then you can use all the methods sync. It's not perfect, but it at leasts allows us to not rely on `synckit`, which makes things slow

Why the separate file? `index.ts` has a call to `import.meta.url`, which prevented my CJS-based application to even import the file. Annoying. ESM in VS Code extensions when 😠

## Testing

Tested manually

## Docs

N/A

### DIFF
--- a/packages/compiler/node/sync.cts
+++ b/packages/compiler/node/sync.cts
@@ -1,7 +1,7 @@
 import { readFileSync } from 'node:fs';
-import type * as types from '../shared/types';
-import Go from './wasm_exec';
 import { join } from 'node:path';
+import type * as types from '../shared/types';
+import Go from './wasm_exec.js';
 
 type UnwrappedPromise<T> = T extends (...params: any) => Promise<infer Return> ? (...params: Parameters<T>) => Return : T;
 
--- a/packages/compiler/node/sync.cts
+++ b/packages/compiler/node/sync.cts
@@ -1,6 +1,7 @@
-import { promises as fs } from 'fs';
+import { readFileSync } from 'node:fs';
 import type * as types from '../shared/types';
 import Go from './wasm_exec';
+import { join } from 'node:path';
 
 type UnwrappedPromise<T> = T extends (...params: any) => Promise<infer Return> ? (...params: Parameters<T>) => Return : T;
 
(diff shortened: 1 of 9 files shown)


Princesseuh:
!preview sync

Princesseuh:
!preview sync

bluwy:
I believe `es-module-lexer` has this pattern too, where you can call `await init` so that the rest of the functions are sync. If `init` isn't called before, the first function call would return a promise as fallback. Maybe we can also have this same flow for both ESM and CJS.

Princesseuh:
!preview sync

Princesseuh:
!preview sync

packages/compiler/node/sync.cts
bluwy, line 1: When building this file, the output is ESM with a `.cjs` extension. Is that fine?

packages/compiler/node/sync.cts
```diff
@@ -0,0 +1,62 @@
+import { readFileSync } from 'node:fs';
+import type * as types from '../shared/types';
+import Go from './wasm_exec';
```
bluwy, line 3: Do we need a `.js` here?

```suggestion
import Go from './wasm_exec.js';
```

packages/compiler/node/sync.cts
Princesseuh: Hmm, that's true. It works though... It was mostly about avoiding `import.meta.url`, it being ESM does not really matter

bluwy: Maybe we can stick with `.js` then if we don't need to support the entry being imported from CJS.

Princesseuh: I think that won't work, because then all the `.js` files will be treated as ESM and `__dirname` will cause issues, won't it?

bluwy (approved):

Princesseuh:
!preview sync

PR #773, opened 2023-04-18
TSX: Map default export to 0:0

Princesseuh:
## Changes

Map the default export to 0:0 so things mapped to the default export at least go to 0:0 by default (ex: Go to Definition)

## Testing

Added a test

## Docs

N/A


Princesseuh:
!preview map-export

Princesseuh:
Unnecessary after some tooling changes, it's better for the mapping themselves to be accurate and for us to fix this downstream

Issue #782, opened 2023-05-02
🐛 BUG: [TSX] Fragment using `<>` syntax crashes compiler


Princesseuh:
### What version of `@astrojs/compiler` are you using?

1.3.2

### What package manager are you using?

pnpm

### What operating system are you using?

macOS

### Describe the Bug

The following code:
```astro
<>+0123456789</>
```
(nothing else in the file, this is the only content)

crashes the compiler when running `convertToTSX` with the following error:
```
panic: runtime error: slice bounds out of range [-1:]

goroutine 6 [running]:
github.com/withastro/compiler/internal/printer.renderTsx(0x75ba98, 0x6d6b40)
	github.com/withastro/compiler/internal/printer/print-to-tsx.go:427 +0x5a8
github.com/withastro/compiler/internal/printer.renderTsx(0x75ba98, 0x6d68c0)
	github.com/withastro/compiler/internal/printer/print-to-tsx.go:134 +0xdb
github.com/withastro/compiler/internal/printer.PrintToTSX({0x6b8510, 0x10}, 0x6d68c0, {{0x3913f, 0x6}, {0x6c20a0, 0x41}, {0x6c20a0, 0x41}, {0x438d4, ...}, ...}, ...)
	github.com/withastro/compiler/internal/printer/print-to-tsx.go:25 +0xc
main.ConvertToTSX.func1({{}, 0x7ff800010000000d, 0x6b84e0}, {0x42c820, 0x2, 0x2})
	./astro-wasm.go:237 +0x2a
syscall/js.handleEvent()
	syscall/js/func.go:94 +0x26
exit code: 2
```

Causing https://github.com/withastro/language-tools/issues/518 downstream

### Link to Minimal Reproducible Example

See description

PR #783, opened 2023-05-02
Fix usage of fragments in TSX

Princesseuh:
## Changes

Fix https://github.com/withastro/compiler/issues/782

## Testing

Added a test

## Docs

N/A


internal/printer/print-to-tsx.go
```diff
@@ -421,12 +421,15 @@ declare const Astro: Readonly<import('astro').AstroGlobal<%s>>`, props.Ident)
 	if len(n.Attr) == 0 {
 		endLoc = n.Loc[0].Start + len(n.Data) - 1
 	}
+	if endLoc == -1 {
+		endLoc = 0
+	}
```
natemoo-re, line 426: This was the problem!

natemoo-re (approved):
Thanks for adding the test, found the issue and fixed!

PR #787, opened 2023-05-03
Restart service on crashes when using `convertToTSX` and `parse` like for `transform`

Princesseuh:
## Changes

Whenever `transform` panic, the compiler force a restart of the service to avoid the "Go Program already Exited" error. This PR adds that mechanism to `parse` and `convertToTSX` as well. Otherwise it crashes our entire language-server whenever the compiler fails to convert a file to TSX

## Testing

Tested manually, service isn't actually supposed to crash, ha!

## Docs

N/A

### DIFF
--- a/.changeset/dry-knives-glow.md
+++ b/.changeset/dry-knives-glow.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Handle crashes when using parse and convertToTSX by restarting the service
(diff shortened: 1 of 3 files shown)


.changeset/dry-knives-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Handle crashes when using parse and convertToTSX by restarting the service
```
ematipico, line 5:
```suggestion
Handle crashes when using `parse` and `convertToTSX` by restarting the service
```

ematipico (approved):

matthewp (approved):

PR #790, opened 2023-05-05
Polyfill the entire crypto object

andremralves:
## Changes

Fixes https://github.com/withastro/astro/issues/6870

- Polyfill the entire crypto object in `wasm_exec.ts`

First I tried to add the `crypto` object to be polyfilled in `@astrojs/webapi`, but as it was already being polyfilled here before, the `crypto` polyfill there was being ignored.

So I think the better solution would be to complete the polyfill here.

I mentioned some alternatives in this PR: https://github.com/withastro/astro/pull/6981

## Testing

I can add if necessary.

## Docs

n/a


bluwy:
I think the CI fails are actual fails because we still support node12 and 14 (we should probably remove support for that in the next major). Maybe we could check if `crypto.webcrypto` exist before polyfilling? And if not, it'll fallback to the old semi-polyfilled trick.

The reason also being that Astro supports node v16.12.0, while Node's webcrypto is only added in node v16.17.0.

andremralves:
> I think the CI fails are actual fails because we still support node12 and 14 (we should probably remove support for that in the next major). 
I noticed the error. I thought about updating the versions, but I imagined that there was a reason for that.

> Maybe we could check if `crypto.webcrypto` exist before polyfilling? And if not, it'll fallback to the old semi-polyfilled trick.

It's a good solution. I will do that. Thanks.

andremralves:
I made the changes. Should I add a version bump for that or not?

bluwy (approved):

PR #800, opened 2023-05-31
feat: export compiler as dual CJS / ESM package

Princesseuh:
## Changes

Use `tsup` to distribute the compiler as both CJS and ESM, so people (me) downstream can use it fully

## Testing

Tests should still pass + tested manually in prettier plugin in this PR: https://github.com/withastro/prettier-plugin-astro/pull/349

## Docs

N/A


Princesseuh:
!preview dual

packages/compiler/package.json
```diff
@@ -7,38 +7,51 @@
...
+      "types": "./dist/node/utils.d.ts",
+      "browser": "./dist/browser/utils.js",
+      "import": "./dist/node/utils.js",
+      "require": "./dist/node/utils.cjs",
+      "default": "./dist/browser/utils.js"
     },
-    "./astro.wasm": "./astro.wasm",
-    "./types": "./types.d.ts"
+    "./astro.wasm": "./dist/astro.wasm",
+    "./types": "./dist/shared/types.d.ts"
```
natemoo-re, line 46: Not strictly required but I like to make sure this is always exposed. Surprised we didn't already have that!

```suggestion
    "./package.json": "./package.json"
```

.changeset/kind-icons-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/compiler': minor
```
natemoo-re, line 2: I think `minor` is "nicer," but... this is technically a `major` breaking change, right? Version numbers are free!

natemoo-re (commented):
Looks great! Just had a few comments.

.changeset/kind-icons-attack.md
natemoo-re: Discussed on Discord and we decided to go with `minor`.

natemoo-re (approved):
Approved. Super clean, thanks for tackling this!

PR #816, opened 2023-07-04
Preserve whitespace in text surrounded by newlines

### LINKED ISSUE #815: 🐛 BUG: HTML minification removes whitespace that should not be removed

1.5.3

pnpm

Linux (NixOS)

### Describe the Bug

Here's an example:

```html
<p>
    A
    <span>B</span>
    C
</p>
```

| Value of `compressHTML` | Screenshot of result |
|---|---|
|`false`|image|
|`true`|image|

The whitespace is not removed if "B" is also text. I have several more examples in the repo I provided.

https://gitlab.com/wackbyte/astro-compress-html-removes-whitespace

wackbyte:
## Changes

Closes #815. This issue came about as withastro/astro#7401 reported the previous whitespace-preserving behavior as incorrect, leading to #809. See [my comment](https://github.com/withastro/astro/issues/7401#issuecomment-1619327198) for why the behavior introduced by that PR is incorrect.

TLDR: revert #809 (& more)

## Testing

I added a new test based on #815 and fixed a few problems with old tests.

## Docs

Documentation not changed as this is purely a bugfix.

### DIFF
--- a/.changeset/soft-eggs-exercise.md
+++ b/.changeset/soft-eggs-exercise.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Do not remove surrounding whitespace from text surrounded by newlines when `compressHTML` is enabled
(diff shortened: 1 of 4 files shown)


wackbyte:
(It would also be possible to drop `hasNewline` and instead append/prepend only spaces)

natemoo-re (approved):
Thanks for the fix and for updating those tests!

Totally agree with [your comment](https://github.com/withastro/astro/issues/7401#issuecomment-1619327198) about the intended behavior. Sorry for the regression!

PR #820, opened 2023-07-06
Add raw attribute values to AST

### LINKED ISSUE #796: 🐛 BUG: [AST] Serialize does not conserve quote type used for attributes

1.3.2

pnpm

macOS

### Describe the Bug

The AST for the following code:

```ts
<span class='after:content-["this"] after:mx-4'></span>
```

Does not include that `class` is using single quote, so `serialize` always uses double quotes, but in this case, it generates invalid code.

See description

natemoo-re:
## Changes

- Resolves #796
- In the AST, quoted attributes now contain their raw values
- This allows `serialize` to maintain the originally authored format of the attribute (double, single, or no quotes)

## Testing

Test added

## Docs

Need to update docs still!

### DIFF
--- a/.changeset/orange-kangaroos-dress.md
+++ b/.changeset/orange-kangaroos-dress.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+[AST] add raw attribute values to AST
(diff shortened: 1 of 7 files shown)


ematipico (approved):

ematipico:
Looks like some test needs to be updated

PR #822, opened 2023-07-07
Maintain whitespace before tag close

### LINKED ISSUE #801: 🐛 BUG: [TSX] Spaces in tag names are not preserved unless there's an attribute

1.5.0

pnpm

macOS

### Describe the Bug

`<Button ></Button>` gets transformed into `<Button></Button>` (note the missing space), as such, when asking completions here: `<Button |>`, you get unexpected things because it's trying to complete the component's tag, and not a value at the requested position 
image

image

Cause of https://github.com/withastro/language-tools/issues/566

See description

natemoo-re:
## Changes

- Resolves #801
- Maintains trailing whitespace in TSX output, which fixes TS completions in some cases

## Testing

Two tests added

## Docs

N/A, bug fix only

### DIFF
--- a/internal/printer/print-to-tsx.go
+++ b/internal/printer/print-to-tsx.go
@@ -264,6 +264,7 @@ declare const Astro: Readonly<import('astro').AstroGlobal<%s>>`, props.Ident)
 	p.print("<")
 	p.addSourceMapping(loc.Loc{Start: n.Loc[0].Start})
 	p.print(n.Data)
+	p.addSourceMapping(loc.Loc{Start: n.Loc[0].Start + len(n.Data)})
 
 	invalidTSXAttributes := make([]Attribute, 0)
 	endLoc := n.Loc[0].Start + len(n.Data)
@@ -422,6 +423,7 @@ declare const Astro: Readonly<import('astro').AstroGlobal<%s>>`, props.Ident)
 	isSelfClosing := false
 	hasLeadingSpace := false
 	tmpLoc := endLoc
+	leadingSpaceLoc := endLoc
 	if len(p.sourcetext) > tmpLoc {
 		for i := 0; i < len(p.sourcetext[tmpLoc:]); i++ {
 			c := p.sourcetext[endLoc : endLoc+1][0]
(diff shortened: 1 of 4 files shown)


internal/printer/print-to-tsx.go
```diff
@@ -375,15 +375,13 @@ declare const Astro: Readonly<import('astro').AstroGlobal<%s>>`, props.Ident)
 			p.print(`:`)
 			p.addSourceMapping(loc.Loc{Start: eqStart + 1})
 			p.print(`"` + encodeDoubleQuote(a.Val) + `"`)
-			endLoc = eqStart + 1 + len(a.Val) + 2
```
natemoo-re, line 378: Turns out these would mess up the `endLoc` because they could be out of order

internal/printer/print-to-tsx.go
```diff
@@ -436,6 +432,9 @@ declare const Astro: Readonly<import('astro').AstroGlobal<%s>>`, props.Ident)
 				p.addSourceMapping(loc.Loc{Start: endLoc})
 				endLoc++
 				break
+			} else if unicode.IsSpace(rune(c)) {
+				hasLeadingSpace = true
+				endLoc++
```
natemoo-re, line 437: If we find a space before the `>`, we flag it

internal/printer/print-to-tsx.go
```diff
@@ -444,13 +443,16 @@ declare const Astro: Readonly<import('astro').AstroGlobal<%s>>`, props.Ident)
 		endLoc++
 	}
 
+	if hasLeadingSpace {
+		p.addSourceMapping(loc.Loc{Start: endLoc - 1})
+		p.print(" ")
+	}
```
natemoo-re, line 449: And if we have the flag, we print an extra `" "` character

Princesseuh (approved):
Logic seems fine to me, but I have one question, what happens if there's multiple spaces?

```ts
<Button      >
```

Will they all get mapped to the first space that was created, since only one space is created?

natemoo-re:
Great question @Princesseuh! I just pushed a change to make the mapping a bit more clear for that case—that one space technically gets mapped back to the whole block of spaces in the source file. I think TypeScript should handle this just fine, but we can adjust if that's not the case.

PR #832, opened 2023-07-17
Support for view transition directives

matthewp:
## Changes

- `transition:animate` on an element is turned into `data-astro-transition-scope={renderTransition()}` which is used to create a scope name for the particular element and set up the CSS.
- Automatically includes the View transition CSS if one of the directives are used.

## Testing

- Testing all done in core for now

## Docs

N/A

### DIFF
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -118,8 +118,10 @@ func (p *printer) printInternalImports(importSpecifier string, opts *RenderOptio
 	p.print("defineStyleVars as " + DEFINE_STYLE_VARS + ",\n  ")
 	p.addNilSourceMapping()
 	p.print("defineScriptVars as " + DEFINE_SCRIPT_VARS + ",\n  ")
-	p.addNilSourceMapping()
-	p.print("renderTransition as " + RENDER_TRANSITION + ",\n  ")
+	if opts.opts.ExperimentalTransitions {
+		p.addNilSourceMapping()
+		p.print("renderTransition as " + RENDER_TRANSITION + ",\n  ")
+	}
 
 	// Only needed if using fallback `resolvePath` as it calls `$$metadata.resolvePath`
 	if opts.opts.ResolvePath == nil {
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -28,6 +28,7 @@ type printer struct {
 	hasFuncPrelude     bool
 	hasInternalImports bool
 	hasCSSImports      bool
+	needsTransitionCSS bool
 }
 
 var TEMPLATE_TAG = "$$render"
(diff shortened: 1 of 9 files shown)


internal/printer/printer.go
```diff
@@ -141,6 +145,10 @@ func (p *printer) printCSSImports(cssLen int) {
 		p.print(fmt.Sprintf("import \"%s?astro&type=style&index=%v&lang.css\";", p.opts.Filename, i))
 		i++
 	}
+	if p.needsTransitionCSS {
+		p.addNilSourceMapping()
+		p.print(`import "astro/components/viewtransitions.css";`)
```
bluwy, line 150: Should we add a new experimental option for this instead? It would also help with conditionally importing `renderTransition` so we don't break older versions of Astro that uses the latest compiler version, as they don't export `renderTransition`

internal/printer/printer.go
```diff
@@ -250,15 +258,19 @@ func (p *printer) printFuncPrelude(opts transform.TransformOptions) {
...
 		escapedFilename := strings.ReplaceAll(opts.Filename, "'", "\\'")
-		p.println(fmt.Sprintf("}, '%s');", escapedFilename))
-	} else {
-		p.println("});")
+		filenameArg = fmt.Sprintf("'%s'", escapedFilename)
+	}
+	if n.Transition {
+		propagationArg = "'self'"
 	}
+	p.println(fmt.Sprintf("}, %s, %s);", filenameArg, propagationArg))
```
bluwy, line 273: I wonder if the propagation handling could be moved to core somehow 🤔 maybe `renderTransition` could update the propagation type? It would be nice to reduce the params needed to create an Astro component.

internal/printer/printer.go
matthewp: Ah yes, I didn't think about that, but good point.

internal/printer/printer.go
bluwy: Hoping to reduce the API surface for the runtime if we ever need another implementation of it, but if it's not possible now I'm also fine with this.

internal/printer/printer.go
```diff
@@ -141,6 +147,10 @@ func (p *printer) printCSSImports(cssLen int) {
 		p.print(fmt.Sprintf("import \"%s?astro&type=style&index=%v&lang.css\";", p.opts.Filename, i))
 		i++
 	}
+	if p.needsTransitionCSS {
+		p.addNilSourceMapping()
+		p.print(`import "astro/components/viewtransitions.css";`)
```
bluwy, line 152: I didn't clearly comment this yesterday, I was also wondering if we should create a new option for this import path, similar to how we have the `internalURL` option.

Not blocking as this is experimental and we can change again if we want.

bluwy (approved):

internal/printer/printer.go
matthewp: Ah yeah, I can add that, good call.

bluwy: Awesome. Do we also want to prefix with "experimental"? (just in case)

PR #834, opened 2023-07-20
Handle expressions for transition:name

matthewp:
## Changes

- Fixes this: `<div transition:name={post.data.slug}>`. Previously only worked correctly with string values.

## Testing

- Test added

## Docs

Bug fix

### DIFF
--- a/internal/printer/print-to-js.go
+++ b/internal/printer/print-to-js.go
@@ -447,6 +447,8 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 					transitionExpr = fmt.Sprintf(`"%s"`, attr.Val)
 				case astro.ExpressionAttribute:
 					transitionExpr = fmt.Sprintf(`(%s)`, attr.Val)
+				case astro.TemplateLiteralAttribute:
+					transitionExpr = fmt.Sprintf("`%s`", attr.Val)
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


internal/printer/print-to-js.go
```diff
@@ -439,14 +439,20 @@ func render1(p *printer, n *Node, opts RenderOptions) {
...
-			transitionName := ""
+			transitionExpr := ""
 			if transform.HasAttr(n, transform.TRANSITION_NAME) {
-				transitionName = transform.GetAttr(n, transform.TRANSITION_NAME).Val
+				attr := transform.GetAttr(n, transform.TRANSITION_NAME)
+				switch attr.Type {
+				case astro.QuotedAttribute:
+					transitionExpr = fmt.Sprintf(`"%s"`, attr.Val)
+				case astro.ExpressionAttribute:
+					transitionExpr = fmt.Sprintf(`(%s)`, attr.Val)
```
natemoo-re, line 449: Probably want to handle `TemplateLiteralAttribute` as well?

natemoo-re (approved):

internal/printer/print-to-js.go
matthewp: Ah ok, will do that. didn't even realize we supported that.

matthewp: Added.

PR #838, opened 2023-07-26
Pass transition directives onto components

matthewp:
## Changes

- Adds the `data-astro-transition-scope` var into components as well. This is needed so they can pass the result down to children, if they want.

## Testing

- New test added

## Docs

- N/A

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -2815,6 +2815,15 @@ const items = ["Dog", "Cat", "Platipus"];
 				code: `${$$maybeRenderHead($$result)}<div${$$addAttribute($$renderTransition($$result, "", (slide({duration:15})), ""), "data-astro-transition-scope")}></div>`,
 			},
 		},
+		{
+			name:     "transition:animate on Component",
+			only:     true,
+			source:   `<Component class="bar" transition:animate="morph"></Component>`,
+			filename: "/projects/app/src/pages/page.astro",
+			want: want{
+				code: `${$$renderComponent($$result,'Component',Component,{"class":"bar","data-astro-transition-scope":($$renderTransition($$result, "", "morph", ""))})}`,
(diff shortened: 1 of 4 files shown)


internal/printer/printer_test.go
```diff
@@ -2815,6 +2815,15 @@ const items = ["Dog", "Cat", "Platipus"];
 				code: `${$$maybeRenderHead($$result)}<div${$$addAttribute($$renderTransition($$result, "", (slide({duration:15})), ""), "data-astro-transition-scope")}></div>`,
 			},
 		},
+		{
+			name:     "transition:animate on Component",
+			only:     true,
+			source:   `<Component class="bar" transition:animate="morph"></Component>`,
+			filename: "/projects/app/src/pages/page.astro",
+			want: want{
+				code: `${$$renderComponent($$result,'Component',Component,{"class":"bar","data-astro-transition-scope":($$renderTransition($$result, "", "morph", ""))})}`,
```
ematipico, line 2824: It's tough to understand the expected result. Are the parentheses needed in `($$renderTransition($$result, "", "morph", ""))})`? 

```diff
- ($$renderTransition($$result, "", "morph", ""))
+ $$renderTransition($$result, "", "morph", "")
```

It should yield the same result

ematipico (approved):

internal/printer/printer_test.go
matthewp: I think our attribution expression code always wraps expressions with parens. Maybe we can change it? I don't know. I assume it's to protect against something... but what alludes me. Perhaps in case someone puts a statement inside of an expression, the parens make it into an expression? idk.

43081j:
@matthewp you left an `only: true` in FYI so the tests in main are now skipping all of the other printer test cases 👀

PR #840, opened 2023-07-28
Persist

matthewp:
Just a draft in order to demo the change.

## Changes

- What does this change?
- Be short and concise. Bullet points can help!
- Before/after screenshots can be helpful as well.

### DIFF
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -50,15 +50,9 @@ func Transform(doc *astro.Node, opts TransformOptions, h *handler.Handler) *astr
 			n.TransitionScope = astro.HashString(fmt.Sprintf("%s-%v", opts.Scope, i))
 		}
 
-		if opts.ExperimentalTransitions {
-			transitionPersistIndex := AttrIndex(n, TRANSITION_PERSIST)
-			if transitionPersistIndex != -1 {
-				n.Attr = remove(n.Attr, transitionPersistIndex)
-				n.Attr = append(n.Attr, astro.Attribute{
-					Key: "data-astro-transition-persist",
-					Val: "",
-				})
-			}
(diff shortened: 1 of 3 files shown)


matthewp:
!preview view-transitions

matthewp:
!preview view-transitions

internal/printer/printer.go
```diff
@@ -272,18 +277,38 @@ func (p *printer) printFuncSuffix(opts transform.TransformOptions, n *astro.Node
 	if n.Transition {
 		propagationArg = "'self'"
 	}
+	fmt.Printf("PROPAGATION ARG %v\n", propagationArg)
 	p.println(fmt.Sprintf("}, %s, %s);", filenameArg, propagationArg))
 	p.println(fmt.Sprintf("export default %s;", componentName))
 }
 
+var skippedAttributes = map[string]bool{
+	"define:vars":        true,
```
matthewp, line 286: These are exactly the same except that the object form doesn't skip `define:vars`.

internal/printer/printer.go
```diff
@@ -458,6 +483,27 @@ func maybeConvertTransition(n *astro.Node) {
...
 		})
 	}
+	if transform.HasAttr(n, transform.TRANSITION_PERSIST) {
+		transitionPersistIndex := transform.AttrIndex(n, transform.TRANSITION_PERSIST)
+		// If there no value, create a transition scope for this element
+		if n.Attr[transitionPersistIndex].Val != "" {
+			// Just rename the attribute
+			n.Attr[transitionPersistIndex].Key = "data-astro-transition-persist"
+
+		} else if transform.HasAttr(n, transform.TRANSITION_NAME) {
```
matthewp, line 493: If there is a transition:name on the element, use that as the persist value as well.

`<div transition:name="media-player" transition:persist>` This just avoids repeating the same thing.

PR #842, opened 2023-07-31
feat: new  style strategy

ematipico:
## Changes

Implements the new `scopedStyleStrategy: 'attribute'`.

We had three alternatives to implement the strategy:
- `<div data-astro-hash="1234">`
- `<div data-astro-hash-1234>`
- `<div astro-hash-1234>`

I want for the second:
- we use the data attributes, easy to query
- the hash is something that we create during compilation, and we should expose it as value and allow users to "easily" obtain it

## Testing

Added a new test case

## Docs

N/A


bluwy (commented):
Does this add the attribute to the html? I'm assuming a tweak needs to be added somewhere https://github.com/withastro/compiler/blob/002fd34fd82f824862ffb2999a1a1162575cd6d8/internal/transform/scope-html.go#L96

ematipico:
Thank you @bluwy for pointing that out. I updated the code and added new test cases.

internal/transform/scope-html.go
```diff
@@ -95,72 +95,81 @@ func injectDefineVars(n *astro.Node, values []string) {
 
 func injectScopedClass(n *astro.Node, opts TransformOptions) {
 	hasSpreadAttr := false
-	scopedClass := fmt.Sprintf(`astro-%s`, opts.Scope)
-	for i, attr := range n.Attr {
-		if !hasSpreadAttr && attr.Type == astro.SpreadAttribute {
-			// We only handle this special case on built-in elements
-			hasSpreadAttr = !n.Component
-		}
+	if opts.ScopedStyleStrategy != "attribute" {
```
ematipico, line 98: I would like to run the equality check using the enum `ScopeStrategy.ScopeStrategyAttribute`, but I didn't find a good way to import this enum inside this file. Any help would be appreciated

matthewp (approved):
Looks great!

internal/transform/scope-html.go
```diff
@@ -95,72 +95,81 @@ func injectDefineVars(n *astro.Node, values []string) {
...
+			Key:  fmt.Sprintf(`data-astro-hash-%s`, opts.Scope),
+			Type: astro.EmptyAttribute,
+		})
 	}
-	// If there's a spread attribute, `class` might be there, so do not inject `class` here
-	// `class` will be injected by the `spreadAttributes` helper
-	if hasSpreadAttr {
-		return
-	}
-	// If we didn't find an existing class attribute, let's add one
```
bluwy, line 160: Not sure if it's a mistake this comment was removed, but it looks like a nice one to preserve.

bluwy (approved):

internal/transform/scope-html.go
ematipico: Ah shoot, definitely not my intention. Great catch! Thank you!

PR #856, opened 2023-08-21
feat: lower case hash

ematipico:
## Changes

This PR changes the hash emitted by the compiler, by making it lower-case.

## Testing

I updated all the test cases. The current CI should pass.

## Docs

Internal change, is not needed

### DIFF
--- a/.changeset/popular-wombats-promise.md
+++ b/.changeset/popular-wombats-promise.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Make astro hash lower case
--- a/.changeset/popular-wombats-promise.md
+++ b/.changeset/popular-wombats-promise.md
@@ -1,5 +1,9 @@
 ---
-'@astrojs/compiler': patch
+'@astrojs/compiler': major
 ---
 
-Make astro hash lower case
+The scope hash created by the compiler is now **lowercase**.
+
+This aligns with the HTML spec of the attribute names, where they are lowercase by spec.
+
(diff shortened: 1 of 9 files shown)


.changeset/popular-wombats-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/compiler': patch
```
matthewp, line 2: probably should be `major` I guess, right?

matthewp: idk, always have a hard time judging  these grey-area changes. Happy with whatever @natemoo-re thinks.

matthewp (approved):
just to figure out patch vs. minor vs. major

natemoo-re (approved):
I'm thinking it's a major just since all of our tests needed to be updated. Seems silly, but it is technically breaking.

.changeset/popular-wombats-promise.md
ematipico, line 2:
```suggestion
'@astrojs/compiler': major
```

.changeset/popular-wombats-promise.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Make astro hash lower case
```
ematipico, line 5:
```suggestion
The scope hash created by the compiler is now **lowercase**.

This aligns with the HTML spec of the attribute names, where they are lowercase by spec.

This change is needed because the compiler now creates data attributes that contain the hash in their name.
```

ematipico:
Thank you. I updated the changelog, to give a better explanation.

PR #858, opened 2023-08-23
Remove experimental transition flags

natemoo-re:
## Changes

- Removes `experimental` flags related to ViewTransitions (stable in 3.0)
- Will fix an error that throws in the current RC, since `astro` doesn't pass those flags anymore

## Testing

Manually

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/giant-boats-mate.md
+++ b/.changeset/giant-boats-mate.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Remove experimental flags from `transition:` directives. They are now enabled by default.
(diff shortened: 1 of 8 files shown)


.changeset/giant-boats-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/compiler': patch
```
matthewp, line 2: This will have to be a major because we are depending on runtime APIs that won't exist for older versions of Astro.

matthewp (approved):

.changeset/giant-boats-mate.md
natemoo-re: Will it? `astro@2.10.13` has `"@astrojs/compiler": "^1.8.0"` as a dependency. `astro@3.0.0-rc.5` has `"@astrojs/compiler": "^2.0.0"` as a dependency.

The compiler version in 3.0.0-rc.5 is currently broken, so this is a bug fix for that release line.

matthewp: Ok, didn't realize we already did a major for the compiler, nm then.

PR #873, opened 2023-10-03
[TSX] infer props and params from getStaticPaths

natemoo-re:
## Changes

- This PR is relevant for consumers of our TSX output (namely language tools)
- Adds new generated output if an Astro component has a `getStaticPaths` export
- `Astro.props` and `Astro.params` can be inferred from the return value of `getStaticPaths`
  - Existing `Props` definitions take precedence over inferred definitions, so this should be compatible with the [existing recommended pattern](https://docs.astro.build/en/guides/typescript/#infer-getstaticpaths-types) below:
  ```ts
  type Params = InferGetStaticParamsType<typeof getStaticPaths>;
  type Props = InferGetStaticPropsType<typeof getStaticPaths>;
  ```
- Ideally we'd also be able to inject a `satisfies` clause for `getStaticPaths` itself, but that can be left for a follow-up PR

## Testing

Tests added

## Docs

We should update the docs when this lands upstream in the language tools!


internal/printer/print-to-tsx.go
```diff
@@ -134,15 +136,34 @@ func renderTsx(p *printer, n *Node) {
...
+			if propsIdent == "Record<string, any>" {
+				propsIdent = "Get<InferredGetStaticPath, 'props'>"
+			}
+		}
 
-		p.print(fmt.Sprintf("export default function %s%s(_props: %s%s): any {}\n", componentName, props.Statement, props.Ident, props.Generics))
-		if props.Ident != "Record<string, any>" {
+		p.print(fmt.Sprintf("export default function %s%s(_props: %s%s): any {}\n", componentName, props.Statement, propsIdent, props.Generics))
+		if hasGetStaticPaths {
+			p.printf(`type ArrayElement<ArrayType extends readonly unknown[]> = ArrayType extends readonly (infer ElementType)[] ? ElementType : never;
```
Princesseuh, line 150: All those types should probably have some sort of marker on them (something like `ASTRO__Flattened`) because they end up in the global scope and could conflict with user code

natemoo-re: Yep, great catch! I've added `ASTRO__` as a prefix

Princesseuh (approved):
Awesome work! People will be very happy about this.

PR #881, opened 2023-10-30
fix(tsx): add `MergeUnion` util to automatically inferred Props

natemoo-re:
## Changes

- Automatically inferred `Props` has shipped!
- There was an edge case with union `Props` where they couldn't be destructured. 
- This PR adds a `MergeUnion` utility to flatten unions into each other so that divergent props don't need to be manually discriminated before destructuring.

## Testing

Test updated, type util manually tested

## Docs

Bug fix only

### DIFF
--- a/.changeset/large-geese-kick.md
+++ b/.changeset/large-geese-kick.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+[TSX] Add ASTRO\_\_MergeUnion util to allow destructuring from automatically inferred union Prop types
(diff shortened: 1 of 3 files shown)


.changeset/large-geese-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+[TSX] Add ASTRO\_\_MergeUnion util to allow destructuring from automatically inferred union Prop types
```
ematipico, line 5:
```suggestion
[TSX] Add `ASTRO__MergeUnion` util to allow destructuring from automatically inferred union Prop types
```

ematipico: Should this be a minor? We're adding a new API/functionality right?

ematipico (approved):

bluwy (approved):

Issue #870, opened 2023-09-27
Astro misparses table HTML


mb21:
### Astro Info

```block
Astro                    v3.1.4
Node                     v16.20.0
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

The Astro template

```jsx
<table>
    <tr><td>{'foo'}</td></tr>
</table>
<h2>Chats</h2>
```

is misparsed and emits the HTML:

```jsx
<table>
  <tr><td>foo</td></tr>
  <h2>Chats</h2>
</table>
````

which is clearly invalid and not intended.

Browsers parse this as "Chats" appearing above the table containing "foo", instead of below.

### What's the expected result?

Output HTML corresponds to the input HTML.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ewybuh?file=src%2Fpages%2Findex.astro

mb21:
Interestingly, closing the table twice (`</table></table>`) works 🤷

bluwy:
NOTE: https://github.com/withastro/astro/issues/8765 reported a similar issue where the `table` markup at the top of the file is affecting the lower `form` markup generated.

bluwy:
Re-opening since it was reverted in https://github.com/withastro/compiler/pull/889. Going to continue finding a fix

MoustaphaDev:
From your comment in the other issue, I'd assume it's okay if I take this one too @bluwy

PR #885, opened 2023-11-09
Fix expressions within td exitLiteralIM tracking

bluwy:
## Changes

Fix https://github.com/withastro/compiler/issues/870

For starting expressions within a table cell (`td`), `p.exitLiteralIM` (used to check when we can exit the insertion mode while parsing literals) was not assigned so the exit check at

https://github.com/withastro/compiler/blob/f5cff0127d2b2ea59201f6f84d308c57540bde68/internal/parser.go#L2697

Always returned false (default behaviour). Causing any tokens after the first `{` expression to be considered a literal incorrectly, since it'll never exit the insertion mode.

This PR sets `p.exitLiteralIM` to fix it, and also fix a slightly unrelated issue where exiting the literal IM in a table returned to a body IM, which I think is incorrect (?).

## Testing

Added a new test in go

## Docs

n/a. bug fix

### DIFF
--- a/.changeset/odd-dryers-flow.md
+++ b/.changeset/odd-dryers-flow.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fix generated code for expressions within `td` elements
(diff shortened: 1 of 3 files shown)


Princesseuh (approved):
This *looks* fine to me, but I definitely do not have extensive knowledge of this area of the compiler

ematipico (approved):
The test passes and there aren't regressions. Looks good to me!

Issue #879, opened 2023-10-27
🐛 BUG: compact transform removes all spaces between inline elements


cocoliliace:
### What version of `@astrojs/compiler` are you using?

2.1.0

### What package manager are you using?

pnpm

### What operating system are you using?

Alpine Linux

### Describe the Bug

The compact option for transform removes all spaces between two inline elements. This causes documents transformed with and without the compact option to be rendered differently. It should collapse them to one space instead.

For example, the following html should remain the same after transforming with the compact option:
```html
<p>Click <a>here</a> <span>👈</span></p>
```
but it is currently being transformed to
```html
<p>Click <a>here</a><span>👈</span></p>
```

The collapseWhitespace function is defined at
https://github.com/withastro/compiler/blob/main/internal/transform/transform.go#L275

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-mnt287?file=src%2Fpages%2Findex.astro

PR #886, opened 2023-11-10
Fix compact for empty text nodes in elements

bluwy:
## Changes

fix https://github.com/withastro/compiler/issues/879

We're previously collapsing empty text nodes (only contains whitespace) to `""`, but sometimes this is unsafe and should convey a `" "` instead (one space).

This PR checks for cases where we can collapse completely with no spaces, but otherwise would only always collapse to one space. I'm not sure if there's a safer way else to deal with it.

## Testing

Added new test in go

## Docs

n/a. bug fix.

### DIFF
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -277,10 +277,7 @@ func isRawElement(n *astro.Node) bool {
 }
 
 func isWhitespaceInsensitiveElement(n *astro.Node) bool {
-	if n.Data == "head" {
-		return true
-	}
-	return false
+	return n.Data == "head"
 }
 
 func collapseWhitespace(doc *astro.Node) {
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -262,15 +262,25 @@ func isRawElement(n *astro.Node) bool {
 	if n.Type == astro.FrontmatterNode {
 		return true
 	}
+	for _, attr := range n.Attr {
+		if attr.Key == "is:raw" {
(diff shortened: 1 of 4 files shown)


internal/transform/transform.go
```diff
@@ -262,15 +262,25 @@ func isRawElement(n *astro.Node) bool {
 	if n.Type == astro.FrontmatterNode {
 		return true
 	}
+	for _, attr := range n.Attr {
+		if attr.Key == "is:raw" {
+			return true
+		}
+	}
```
bluwy, line 269: It looks like we can extract this out of the `rawTags` loop as it checks against the `n` node only.

internal/transform/transform.go
```diff
@@ -262,15 +262,25 @@ func isRawElement(n *astro.Node) bool {
...
+	return false
+}
+
+func isWhitespaceInsensitiveElement(n *astro.Node) bool {
+	rawTags := []string{"head"}
+	for _, tag := range rawTags {
+		if n.Data == tag {
+			return true
 		}
 	}
```
natemoo-re, line 285: Could avoid the loop here if we're only checking for `head` elements?

natemoo-re (approved):
This looks good to me!

internal/transform/transform.go
bluwy: Hmm yeah I can make this an if for now and we can expand this to a loop if needed later.

PR #889, opened 2023-11-14
Revert "Fix expressions within td exitLiteralIM tracking (#885)"

natemoo-re:
## Changes

- This reverts commit e241f2d545e5753eaafe19fac8626f0191cca96c.
- Don't want to break people, going to release this quickly. Can circle back to actually fix the table issue.

## Testing

The following case regressed, throwing a parsing error for the user. Notably, even with the regression, this case is still broken, albeit not fatally. Tables are fun! 😭

```astro
---
const content = "lol";
---

<html>
  <body>
    <table>
      <tr>
        <td>{content}</td>
      </tr>
      {
        (
          <tr>
            <td>1</td>
          </tr>
        )
      }
    </table>
     1
  </body>
</html>
```

## Docs

N/A

### DIFF
--- a/.changeset/stupid-rats-cough.md
+++ b/.changeset/stupid-rats-cough.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Revert table related parsing change as it resulted in a regression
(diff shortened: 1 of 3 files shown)


natemoo-re:
YOLO'ing this since everyone else is offline but we don't want a regression out in the wild.

Issue #609, opened 2022-11-01
Error when comment is in JS expression


jasikpark:
### What version of `astro` are you using?

1.6.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm 

### What operating system are you using?

Mac

### Describe the Bug

When adding a comment to a mapped JS expression, the compiler errors:

```astro
{fontCSSURLs.map((fontCSSURL) => (
  <Fragment>
    <!-- See https://csswizardry.com/2020/05/the-fastest-google-fonts/#google-fonts-async-snippet -->
    <link rel="stylesheet" href={fontCSSURL} media="print" onload="this.media='all'" />
    <noscript>
      <link rel="stylesheet" href={fontCSSURL} />
    </noscript>
  </Fragment>
))}
```
^ this errors (fontCSSURLs is a list of string)

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-k999x5?file=src/pages/index.astro&view=editor

### Participation

- [X] I am willing to submit a pull request for this issue.

MoustaphaDev:
This was fixed in #895 and #897!

khanakia:
No, it's not fixed. i am still facing the issue. it adds `//` as comment that is wrong it should comment it `{/* */}` like this.

@MoustaphaDev any idea how to fix it?

```astro
{
posts.map((post) => (
	<li>
		<a href={`/blog/${post.slug}/`}>
			<h4 class="title">{post.data.title}</h4>
			<p class="date">
				// <FormattedDate date={post.data.pubDate} />
			</p>
		</a>
	</li>
))
}
```

khanakia:
Thanks, it needs to be fixed. Currently out code is so bad for developer experience

PR #896, opened 2023-11-23
Fix expression bug

MoustaphaDev:
## Changes

Second part of the fix for #609

## Testing

Added a printer test and fixed and incorrect test

## Docs

N/A bug fix


MoustaphaDev:
Had issues with git, I opened another PR #897

Issue #776, opened 2023-04-19
🐛 BUG:


Burleigh-Beach-Designs:
### What version of `@astrojs/compiler` are you using?

2.3.0

### What package manager are you using?

npm

### What operating system are you using?

11

### Describe the Bug

CompilerError
Unknown compiler error.
runtime error: index out of range [0] with length 0
[See Docs Reference](https://docs.astro.build/en/reference/errors/compiler-error/)
This is almost always a problem with the Astro compiler, not your code. Please open an issue at https://astro.build/issues/compiler.
layouts/MainLayout.astro

---
import MainHead from "./MainHead.astro";
---

<MainHead title={title} description={description} />
<html lang="en">
	<body>
		<slot />
	</body>
</html>

Stack Trace
goroutine 8 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:24 +0x6
github.com/withastro/compiler/internal_wasm/utils.ErrorToJSError(0x47a5a0, {0xb08898, 0x486150})
	github.com/withastro/compiler/internal_wasm/utils/utils.go:68 +0x2
main.Transform.func1.1.1.1()
	./astro-wasm.go:278 +0x5
panic({0x2daa0, 0x486150})
	runtime/panic.go:884 +0x29
github.com/withastro/compiler/internal.inBodyIM(0x47a6c0)
	github.com/withastro/compiler/internal/parser.go:1157 +0x658
github.com/withastro/compiler/internal.(*parser).parseCurrentToken(0x47a6c0)
	github.com/withastro/compiler/internal/parser.go:2966 +0x13
github.com/withastro/compiler/internal.(*parser).parse(0x47a6c0)
	github.com/withastro/compiler/internal/parser.go:2993 +0x3
github.com/withastro/compiler/internal.ParseWithOptions({0x7aaa8, 0x42c1e0}, {0x439af0, 0x1, 0x1})
	github.com/withastro/compiler/internal/parser.go:3057 +0x1c
main.Transform.func1.1.1()
	./astro-wasm.go:283 +0xe
created by main.Transform.func1.1
	./astro-wasm.go:274 +0x16

### Link to Minimal Reproducible Example

https:

Burleigh-Beach-Designs:
ss

Burleigh-Beach-Designs:
ss

MoustaphaDev:
Closing as this was fixed in #891 🥳

PR #901, opened 2023-11-29
Add regression test for #776

MoustaphaDev:
## Changes

Add a regression test for #776. The issue has been fixed in #891

## Testing

Added a transform test

## Docs

N/A bug fix, though ultimately, I think we want to discourages things like using elements before html.

### DIFF
--- a/internal/transform/transform_test.go
+++ b/internal/transform/transform_test.go
@@ -252,9 +252,14 @@ func TestFullTransform(t *testing.T) {
 			want: `<Component><h1>Hello world</h1></Component>`,
 		},
 		{
-			name:   "Component before html",
+			name:   "Component before html I",
 			source: `<Navigation /><html><body><h1>Astro</h1></body></html>`,
 			want:   `<Navigation></Navigation><h1>Astro</h1>`,
+		},
+		{
+			name:   "Component before html II",
+			source: `<MainHead title={title} description={description} /><html lang="en"><body><slot /></body></html>`,
+			want:   `<MainHead title={title} description={description}></MainHead><slot></slot>`,
 		},
 		{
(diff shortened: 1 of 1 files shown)


internal/transform/transform_test.go
```diff
@@ -252,10 +252,15 @@ func TestFullTransform(t *testing.T) {
...
 		{
-			name:   "Component before html",
+			name:   "Component before html I",
 			source: `<Navigation /><html><body><h1>Astro</h1></body></html>`,
 			want:   `<Navigation></Navigation><h1>Astro</h1>`,
 		},
+		{
+			name:   "Component before html II",
+			source: `<MainHead title={title} description={description} /><html lang="en"><body><slot /></body></html>`,
+			want:   `<MainHead title={title} description={description}></MainHead><slot></slot>`,
```
MoustaphaDev, line 262: The output looks incorrect but I believe it's intended?
(though this isn't related to this PR or #891. I'll open an issue so we can discuss about it)

natemoo-re: Hmm, both of these seem incorrect, to be honest. I'd expect the `<html>` and `<body>` to be preserved.

MoustaphaDev: Could we maybe merge this for now and close the issue as it wasn't actually about the incorrect markup, but the `index out of range` errors which have been fixed?

This bug apparently existed since at least v1.4.3 (which is the oldest version my playground supports) for normal elements.

I opened a new issue for this specific bug https://github.com/withastro/compiler/issues/913

natemoo-re: Yep that seems reasonable!

PR #902, opened 2023-11-29
fix: better sourcemaps handling with multibyte characters

### LINKED ISSUE #872: Compiler does not handle unusual line terminator
### Astro Info

```block
  <-- this char is an unusual line terminator when pasting in stackblitz we get a warning but if we don't use an editor that detects this our astro app crashes.

Go to this document https://www.figma.com/file/TyJtpYWrQE7PecTHzuRIaA/Astro.js-bug?type=design&node-id=0%3A1&mode=design&t=MZ9J3mWPSnaMCwVE-1

Right click the text and (copy as text)

Paste into your astro project and save.
```

_No response_

### Describe the Bug

  <-- this char is an unusual line terminator when pasting in stackblitz we get a warning but if we don't use an editor that detects this our astro app crashes.

Go to this document https://www.figma.com/file/TyJtpYWrQE7PecTHzuRIaA/Astro.js-bug?type=design&node-id=0%3A1&mode=design&t=MZ9J3mWPSnaMCwVE-1

Right click the text and (copy as text)

Paste into your astro project and save.

Give a better warning or handle the line terminator in a more graceful way.

https://stackblitz.com/edit/github-baxezw?file=src%2Fpages%2Findex.astro

https://www.fileformat.info/info/unicode/char/2028/index.htm

MoustaphaDev:
## Changes

Fix #872

The initial code inaccurately incremented `start` for each character iteration without considering variable byte sizes. The updated code utilizing `utf8.DecodeRuneInString`, correctly adjusts `start` based on the byte size of each character.

## Testing

Added a printer test

## Docs

N/A bug fix

### DIFF
--- a/.changeset/strange-beans-scream.md
+++ b/.changeset/strange-beans-scream.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fixes an issue causing `index out of range` errors when handling some multibyte characters like `\u2028`.
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -70,17 +70,17 @@ func (p *printer) println(text string) {
 func (p *printer) printTextWithSourcemap(text string, l loc.Loc) {
 	start := l.Start
 	for pos, c := range text {
-		// nextChar will eventually be `RuneError` when we reach the last iteration
-		// but it's ok, as we'll exit the loop anyway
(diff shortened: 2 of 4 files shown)


natemoo-re (approved):
Thanks for the PR!

MoustaphaDev:
I'll be watching closely if this causes any issue. Touching sourcemaps is scary 😆

Issue #733, opened 2023-02-21
🐛 BUG: Styling an emoji - runtime error: index out of range [11] with length 11


tysian:
### What version of `@astrojs/compiler` are you using?

2.0.14

### What package manager are you using?

npm, pnpm

### What operating system are you using?

Windows 10

### Describe the Bug

Hello!

When I'm using `<style></style>` with `lang=scss`, I'm getting an runtime error.
I've noticed that my div has an cookie emoji 🍪 in it, and this was a reason why I was gettin a compiler error. Reproduced bug in stackblitz example.

![image](https://user-images.githubusercontent.com/16374473/220410444-90fcbf48-ab47-4e05-b8c0-7727abdd8e5e.png)

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ze3pye?file=src/CookieBar.astro

jasikpark:
this bug gives me https://fasterthanli.me/articles/the-bottom-emoji-breaks-rust-analyzer energy 😅

tasuren:
As shown as below, not limited to emojis, this seems to happen if html tag has class and a few multibyte characters and also use style.
```astro
<style>a { font-size: 16px; }</style>
<a class="test">ツ</a>
```
It also seems to be possible to build, though not a radical solution, by assigning a character to a constant and using that constant to embed the character in an HTML tag:
```astro
---
const TSU = "ツ";
---
<style>a { font-size: 16px; }</style>
<a class="test">{TSU}</a>
```
Note that in some cases, it seems to work just by doing something like `{"ツ"}` even if you don't put it in a constant.

PR #906, opened 2023-12-13
Fix multibyte character handling

natemoo-re:
## Changes

- Closes https://github.com/withastro/astro/issues/9421
- Closes https://github.com/withastro/compiler/issues/733
- When generating sourcemaps, we were unconditionally accessing an index. For multi-byte characters, this index was sometimes out of range.
- This PR adds a `len()` check before access the index, which guards against the issue.
- There might be a more fundamental fix here, but I really don't think it's a great idea to dig into the sourcemapping logic to try to find it.

## Testing

Tests added for cases that have been reported to throw

## Docs

Bug fix only

### DIFF
--- a/.changeset/dull-bees-grab.md
+++ b/.changeset/dull-bees-grab.md
@@ -2,4 +2,4 @@
 '@astrojs/compiler': patch
 ---
 
-Fix a `index out of range` error when multibyte characters were rendered as markup
+Fix an `index out of range` error when multibyte characters were rendered as markup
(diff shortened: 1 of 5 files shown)


Princesseuh (approved):
Can be tackled separately just to fix the crash, however it'd be great to have a sourcemap test with multi bytes characters to make sure this didn't break stuff!

ematipico (approved):

natemoo-re:
@Princesseuh Added some sourcemap-specific tests in [525b6cc](https://github.com/withastro/compiler/pull/906/commits/525b6ccfdd8c663773ebbb52900f04547758bcf0)! Everything seems to be working well.

MoustaphaDev (approved):
LGTM!

PR #917, opened 2023-12-20
feat(tsx): Prefix tsx output with a JSX pragma

Princesseuh:
## Changes

Add `/** @jsxImportSource astro */` to our TSX output so it uses the newly exported types at https://github.com/withastro/astro/pull/9501, this is all to attempt to fix https://github.com/withastro/language-tools/issues/727

## Testing

Adjusted tests

## Docs

N/A

### DIFF
--- a/internal/printer/print-to-tsx.go
+++ b/internal/printer/print-to-tsx.go
@@ -16,7 +16,7 @@ import (
 )
 
 func getTSXPrefix() string {
-	return "/** @jsx astroHTML */\n\n"
+	return "/** @jsxImportSource astro */\n\n"
 }
 
 func PrintToTSX(sourcetext string, n *Node, opts transform.TransformOptions, h *handler.Handler) PrintResult {
--- a/internal/printer/print-to-tsx.go
+++ b/internal/printer/print-to-tsx.go
@@ -111,7 +111,7 @@ func renderTsx(p *printer, n *Node) {
 			// Without this, TypeScript can get tripped up by the body of our file.
 			if c.PrevSibling != nil && c.PrevSibling.Type == FrontmatterNode {
 				buf := strings.TrimSpace(string(p.output))
-				if len(buf) > 1 {
+				if len(buf)-len(getTSXPrefix()) > 1 {
 					char := rune(buf[len(buf)-1:][0])
 					// If the existing buffer ends with any character other than ;, we need to add a `;`
 					if char != ';' {
--- a/internal/printer/print-to-tsx.go
+++ b/internal/printer/print-to-tsx.go
@@ -15,13 +15,17 @@ import (
 	"golang.org/x/net/html/atom"
 )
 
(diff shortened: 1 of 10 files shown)


Princesseuh:
!preview tsx-prefix

internal/printer/print-to-tsx.go
```diff
@@ -15,13 +15,17 @@ import (
...
+}
+
 func PrintToTSX(sourcetext string, n *Node, opts transform.TransformOptions, h *handler.Handler) PrintResult {
 	p := &printer{
 		sourcetext: sourcetext,
 		opts:       opts,
 		builder:    sourcemap.MakeChunkBuilder(nil, sourcemap.GenerateLineOffsetTables(sourcetext, len(strings.Split(sourcetext, "\n")))),
 	}
-
+	p.print(getTSXPrefix())
```
natemoo-re, line 28: If you move this inside of `renderTsx()` > `if n.Type == DocumentNode`, right after `hasChildren := false`, do you still get the weird behavior?

natemoo-re (commented):
Looks fine to me!

internal/printer/print-to-tsx.go
Princesseuh: Same behaviour, yeah. That's actually where I put it originally, ha!

internal/printer/print-to-tsx.go
```diff
@@ -107,12 +111,12 @@ func renderTsx(p *printer, n *Node) {
...
 			if c.PrevSibling != nil && c.PrevSibling.Type == FrontmatterNode {
 				buf := strings.TrimSpace(string(p.output))
-				if len(buf) > 1 {
+				if len(buf)-len(getTSXPrefix()) > 1 {
 					char := rune(buf[len(buf)-1:][0])
 					// If the existing buffer ends with any character other than ;, we need to add a `;`
 					if char != ';' {
 						p.addNilSourceMapping()
-						p.print("\"\";")
+						p.print("{};")
```
Princesseuh, line 119: From my test, this seemed fine as a replacement of `"";`, but I feel like there must be a reason I went for `"";` back in the day. Can't remember it, though.

(When I first joined Astro, it was `;//`, from https://github.com/withastro/language-tools/pull/59 )

Princesseuh:
!preview tsx-prefix

MoustaphaDev (approved):
Not sure why `""` was replaced by `{}` after frontmatter closing, but the code looks good to me!

Princesseuh:
> The code looks good to me, but I wonder why `""` was replaced by `{}` after frontmatter closing?

Ah, I forgot to mention the justification, it's because it's seen as a string pragma (like use client, use strict) and so TypeScript will attempt to move it at the top of the file, which messes with imports

MoustaphaDev:
Ah ok, makes sense now

PR #918, opened 2023-12-21
Fix template literal attributes on components

### LINKED ISSUE #903: Class with template string breaks when HTML tag is variable
### Astro Info

```block
Astro                    v3.4.0
Node                     v18.17.1
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

I was using an a tag with multiple classes in a template string. 

```
<a
	class=`button regular-text--small--bold bg-${backgroundColor} text-${color} ${size}`
	href={href}
>
	<slot />
</a> 
```

This worked fine. I decided to make the tag variable according to props passed to the component. It stopped working. I noticed that if I only had the string part, without the ${} notation, it would work. To make it work with the variable classes, I needed to use class:list

```
<TagName
	class:list={[
		"button",
		"regular-text--small--bold",
		`bg-${backgroundColor}`,
		`text-${color}`,
		`${size}`,
	]}
	href={href}
>
	<slot />
</TagName>
```

Being able to use template strings with variable tag, without needing class:list

https://github.com/marcelluscaio/portfolioCaioCabral/blob/projectPage/src/components/Elements/Button.astro

### Participation

MoustaphaDev:
## Changes

Fix #903
Components with template literal attributes were printed with the attribute name as value.

## Testing

Added a printer test

## Docs

N/A bug fix

### DIFF
--- a/.changeset/perfect-falcons-build.md
+++ b/.changeset/perfect-falcons-build.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fixes an issue where components with template literal attributes were printed with the name of the attribute as value.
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):
🙌

PR #919, opened 2023-12-21
Fix edge case that caused odd panic error

natemoo-re:
## Changes

- User reported an inscrutable panic error
- I haven't been able to reproduce the specific situation that causes the panic, but I can see where the panic is happening.
- Doing a prerelease to confirm if this fixes the problem

## Testing

I haven't figured out how to repro yet

## Docs

N/A

### DIFF
--- a/.changeset/calm-parents-tell.md
+++ b/.changeset/calm-parents-tell.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fixes an edge case that caused a difficult-to-reproduce panic
--- a/internal/token.go
+++ b/internal/token.go
@@ -1251,10 +1251,11 @@ func (z *Tokenizer) readUnclosedTag() bool {
 		if close != -1 && len(buf) < close {
 			buf = buf[0:close]
 		}
-	}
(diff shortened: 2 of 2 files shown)


natemoo-re:
!preview no-panic

natemoo-re:
!preview no-panic

natemoo-re:
!preview no-panic

MoustaphaDev:
Could be helpful adding a test with the code that fails maybe? If it's a windows error, our CI could catch it (not sure how it could be in this case though)

natemoo-re:
@MoustaphaDev Well part of the problem here is that I'm trying to reproduce it and I'm unable to. CI might be able to catch it, though?

MoustaphaDev:
Probably, yes, if it's a platform specific issue

torchsmith:
It was not working on Mac, and it was also not working on linux server.

torchsmith:
If nate was testing on windows i wonder if it'd error out in ci or even in WSL

natemoo-re:
Nope, I was also trying this on macOS, but I was isolating the file to just run it through the compiler...

Princesseuh:

PR #925, opened 2023-12-23
Fix incorrect table parsing when it contains expressions in certain cases

### LINKED ISSUE #695: 🐛 BUG: Dynamic tables delete following closing tag

0.31.4

npm

MacOS

### Describe the Bug

When creating a table with dynamic values it just makes the following tag to the table closing tag disappear (I guess) if it's a closing tag or teleport to the top if it's not.

A way to fix this is creating a component with only the table inside of it.

 Wrapping the elements with tbody also fixes the issue
 
 Tested also on Ubuntu and Stackblitz, probably not a system related problem.

https://stackblitz.com/edit/github-rzo3qv?file=src/pages/index.astro

### LINKED ISSUE #870: Astro misparses table HTML

MoustaphaDev:
## Changes

- Fixes #870
- Fixes #695

Inside a `table` element, when there was an implied `tr` tag which had a `td` or `th` which contained an expression (that's a mouthful yeah 😅), it would cause any trailing element/component/text node to be swallowed by the table element (just before the `table` closing tag). This fixes that.

## Testing

- Edit the (deliberately) incorrect printer test case introduced in #889
- Brought back the test case reverted in that PR to makes sure that trailing elements aren't swallowed

## Docs

N/A bug fix

### DIFF
--- a/.changeset/perfect-fishes-cover.md
+++ b/.changeset/perfect-fishes-cover.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fixes an issue where a `tr` element which contained an expression would cause its parent table to swallow any trailing element inside said table
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -793,6 +793,7 @@ func inHeadIM(p *parser) bool {
 				return true
 			}
 			p.tok.Data = s
+			return textIM(p)
 		} else if p.oe.top() != nil && (isComponent(p.oe.top().Data) || isFragment((p.oe.top().Data))) {
 			p.addText(p.tok.Data)
 			return true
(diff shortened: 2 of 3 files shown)


MoustaphaDev:
!preview table-tr-expr-issue

natemoo-re (approved):
Amazing, thank you so much for tackling this!

PR #926, opened 2023-12-23
Fix fragment in `tr` printing incorrect output

### LINKED ISSUE #722: Incomplete rendering output with multiple generated tables on page

1.8.0

None

pnpm, npm

Linux

### Describe the Bug

Hey y'all,

happy new year!

I'm still not entirely sure about the title for this one, but here we go.

Based on the [page source code](https://github.com/mundry/astro-issue-incomplete-rendering-multiple-tables/blob/main/src/pages/index.astro) of the attached repo, I'm getting incomplete output for the table in the second `section`.
```html
 <section">
    <div class="row">



                    </div></section"></main></body><th>B</th>
                    <th>B</th>
                    <th>B</th>
                    <th>B</th>
                    <th>B</th>
                    <th>B</th>
                    <th>B</th>



                <td></td><td></td><td></td><td></td><td></td><td></td><td></td>
</html>
```
As you'll notice, the `table`, `thead`, `tbody`, and `tr` tags are missing and the order of the closing block tags is also mixed up.
Removing either one of the expressions (`{'a'}` or `{Array(7).fill...}`) in the `map` function resolves the issue. Removing the outer `map` function also resolves the issue.

Wrapping the `{Array(7).fill...}` expression in a fragment generates another kind of weird output in the tables in the first `section`. ([Source code](https://github.com/mundry/astro-issue-incomplete-rendering-multiple-tables/blob/main/src/pages/frag.astro))
```html
<section>
    <div>
        <div class="row">
            a
            <th>A</th><th>A</th><th>A</th><th>A</th><th>A</th><th>A</th><th>A</th><table>
                <thead>
                    <tr>

                    </tr>
                </thead>
                <tbody>
                    <tr><td></td><td></td><td></td><td></td><td></td><td></td><td></td></tr>
                </tbody>
            </table>
        </div>
    </div><div>
```
Suddenly the output of the `map` call of the `th` elements is placed right before the opening `table` tag, instead of inside the `thead` tag.

The output is identical in `dev` and `build`.

I've tested this with Node 19 and 18, npm and pnpm, as well as with the latest 1.6.x, 1.7.x, and 1.8.x Astro releases.
Until Astro 1.5.3, I even need to remove all expressions in the outer `map` call to get the proper rendering for the table in the second `section`.

So far I haven't had any problems using nested `map` functions with any other tags than tables.

Based off of experience with other frontend frameworks I've tried wrapping with HTML tags as well as with fragments but to no avail.

I'm not really sure what's going on here. What really throws me for a loop (no pun intended... maybe), is that the table in the second section is affected, even though it's completely outside of the scope of any of the JavaScript expressions.

https://github.com/mundry/astro-issue-incomplete-rendering-multiple-tables

### Participation

MoustaphaDev:
## Changes

- Closes #722
- Ensure parsing of fragments in table elements

## Testing

Added a printer test

## Docs

N/A bug fix

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -1789,7 +1789,7 @@ func inTableIM(p *parser) bool {
 		p.im = inExpressionIM
 		return true
 	case StartTagToken:
-		if isComponent(p.tok.Data) {
+		if isComponent(p.tok.Data) || isFragment(p.tok.Data) {
 			p.originalIM = inTableIM
 			p.im = inLiteralIM
 			p.exitLiteralIM = getExitLiteralFunc(p)
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -1792,7 +1792,6 @@ func inTableIM(p *parser) bool {
 		if isComponent(p.tok.Data) || isFragment(p.tok.Data) {
 			p.originalIM = inTableIM
 			p.im = inLiteralIM
-			p.exitLiteralIM = getExitLiteralFunc(p)
 			return false
 		}
 		switch p.tok.DataAtom {
--- a/.changeset/plenty-mice-beam.md
+++ b/.changeset/plenty-mice-beam.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fixes an issue where Astro fragments used inside a `table` element would cause lots of missing pieces of markup
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -2101,8 +2101,9 @@ const content = "lol";
                             <thead>
                                 <tr>
                                     ${$$renderComponent($$result,'Fragment',Fragment,{},{"default": () => $$render` + BACKTICK + `${Array(7).fill(false).map((entry, index) => $$render` + BACKTICK + `<th>A</th>` + BACKTICK + `)}` + BACKTICK + `,})}
-                            </tr>
-</thead><tbody>
+                                </tr>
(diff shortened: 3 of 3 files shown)


internal/parser.go
```diff
@@ -1789,10 +1789,9 @@ func inTableIM(p *parser) bool {
 		p.im = inExpressionIM
 		return true
 	case StartTagToken:
-		if isComponent(p.tok.Data) {
+		if isComponent(p.tok.Data) || isFragment(p.tok.Data) {
 			p.originalIM = inTableIM
 			p.im = inLiteralIM
-			p.exitLiteralIM = getExitLiteralFunc(p)
```
MoustaphaDev, line 1795: I'm not sure why but removing this live fixes the weird white space issue I had in [`ea2dce3` (#926)](https://github.com/withastro/compiler/pull/926/commits/ea2dce32c80e682faa7c4de61ee6e36b7d1ccb86#diff-d0dbf4b902747f2cdc45290c88eb59b04d03bebcb2bd8872a0b49284a9dac9d1R2105)

MoustaphaDev:
!preview fragment-in-tr-issue

internal/parser.go
natemoo-re: Hmm, this was definitely a hacky solution when I introduced it. I'm a little worried this will cause some other regressions, but I'm willing to try it.

natemoo-re (approved):

Issue #784, opened 2023-05-02
🐛 BUG: [TSX] Multiline comments are handled weirdly


Princesseuh:
### What version of `@astrojs/compiler` are you using?

1.3.2

### What package manager are you using?

pnpm

### What operating system are you using?

macOS

### Describe the Bug

Multilines comments seems to not be handled properly, they move around, disappear.. real magicians over there.

For instance, the following code:
```astro
{/* @ts-expect-error */}
<Component prop="value"></Component>

{
// @ts-expect-error
}
<Component prop="value"></Component>

{
/* @ts-expect-error */
<Component prop="value"></Component>
}

{
// @ts-expect-error
<Component prop="value"></Component>
}
```

Becomes this TSX:
```ts
<Fragment>

<Component prop="value"></Component>

{
// @ts-expect-error
}
<Component prop="value"></Component>

{
/* @ts-expect-error */<Fragment><Component prop="value"></Component></Fragment>}

{
// @ts-expect-error
<Fragment><Component prop="value"></Component></Fragment>}
</Fragment>
```

For the average comment, this doesn't really matter. However, it matters for TS pragmas, as their position is important.

### Link to Minimal Reproducible Example

See description

MoustaphaDev:
Would you expect this to be the correct output?
```jsx
<Fragment>

<Component prop="value"></Component>

<Component prop="value"></Component>

{
/* @ts-expect-error */
<Fragment><Component prop="value"></Component></Fragment>
}

{
// @ts-expect-error
<Fragment><Component prop="value"></Component></Fragment>
}
</Fragment>
```

Princesseuh:
I would expect all forms to maintain their shape, so the following would be the optimal result:

```ts
<Fragment>

{/* @ts-expect-error */}
<Component prop="value"></Component>

{
// @ts-expect-error
}
<Component prop="value"></Component>

{
/* @ts-expect-error */
<Fragment><Component prop="value"></Component></Fragment>
}

{
// @ts-expect-error
<Fragment><Component prop="value"></Component></Fragment>
}
</Fragment>
```

PR #930, opened 2023-12-28
Don't ignore whitespace in expressions

MoustaphaDev:
## Changes

- Fix https://github.com/withastro/compiler/issues/784
- The parser doesn't remove whitespace in expressions anymore
- Don't remove expressions that only contain comments in the TSX output
- Handle empty expressions containing only whitespace as before (see https://github.com/withastro/compiler/pull/930/files#r1437911849)

## Testing

- Adjusted whitespace in a few tests
- Added a printer test to make sure empty expressions that contain whitespace are handled like before.
- Added a TSX output test to ensure the whitespace in expressions is respected

## Docs

N/A bug fix

### DIFF
--- a/internal/printer/print-to-js.go
+++ b/internal/printer/print-to-js.go
@@ -316,7 +316,11 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 
 	// Tip! Comment this block out to debug expressions
 	if n.Expression {
-		if n.FirstChild == nil {
+		clean := ""
+		if n.FirstChild != nil {
+			clean = strings.TrimSpace(n.FirstChild.Data)
+		}
+		if n.FirstChild == nil || clean == "" {
 			p.print("${(void 0)")
 		} else if expressionOnlyHasComment(n) {
 			// we do not print expressions that only contain comment blocks
--- a/internal/printer/print-to-tsx.go
+++ b/internal/printer/print-to-tsx.go
@@ -230,9 +230,6 @@ declare const Astro: Readonly<import('astro').AstroGlobal<%s, typeof %s`, propsI
 		p.addSourceMapping(n.Loc[0])
 		if n.FirstChild == nil {
 			p.print("{(void 0)")
-		} else if expressionOnlyHasComment(n) {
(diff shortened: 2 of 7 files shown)


MoustaphaDev:
!preview tsx-comments-whitespace

internal/printer/print-to-js.go
```diff
@@ -316,7 +316,11 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 
 	// Tip! Comment this block out to debug expressions
 	if n.Expression {
-		if n.FirstChild == nil {
+		clean := ""
+		if n.FirstChild != nil {
+			clean = strings.TrimSpace(n.FirstChild.Data)
+		}
+		if n.FirstChild == nil || clean == "" {
```
MoustaphaDev, line 323: We now need to check for whitespace too as the parser doesn't remove them in expressions anymore.

internal/printer/print-to-tsx.go
```diff
@@ -230,9 +230,6 @@ declare const Astro: Readonly<import('astro').AstroGlobal<%s, typeof %s`, propsI
 		p.addSourceMapping(n.Loc[0])
 		if n.FirstChild == nil {
 			p.print("{(void 0)")
-		} else if expressionOnlyHasComment(n) {
-			// we do not print expressions that only contain comment blocks
-			return
```
MoustaphaDev, line 235: Don't remove comments in the TSX output

Princesseuh (approved):
Admittedly not super familiar with the logic around these parts, but this looks good to me!

internal/printer/print-to-tsx.go
MoustaphaDev: I'm not sure, I was wondering the same as that would unexpectedly remove TS directives like `@ts-ignore` in this case.
Pinging @natemoo-re, maybe he could provide some background

natemoo-re: I honestly don't remember the context! Guessing it was just a copy + paste from the JS output which has special handling for this case.

natemoo-re (approved):

PR #931, opened 2024-01-02
feat(tsx): Return interesting ranges for TSX output

Princesseuh:
## Changes

A common issue in the language server is it being hard to map things to the "frontmatter" of the TSX file because, well, we don't know where it is inside the TSX. This PR now returns the start and end of both the frontmatter and the body of the file, which the language server can use to map things.

This is for https://github.com/withastro/language-tools/pull/733

## Testing

Added a test

## Docs

N/A, I updated the types though

### DIFF
--- a/packages/compiler/src/shared/types.ts
+++ b/packages/compiler/src/shared/types.ts
@@ -106,6 +106,16 @@ export interface TSXResult {
   code: string;
   map: SourceMap;
   diagnostics: DiagnosticMessage[];
+  metaRanges: {
+    frontmatter: {
+      start: number;
+      end: number;
+    };
+    body: {
+      start: number;
+      end: number;
+    };
+  };
 }
 
 export interface ParseResult {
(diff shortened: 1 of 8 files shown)


packages/compiler/src/shared/types.ts
```diff
@@ -106,6 +106,16 @@ export interface TSXResult {
   code: string;
   map: SourceMap;
   diagnostics: DiagnosticMessage[];
+  metaRanges: {
+    frontmatter: {
+      start: number;
+      end: number;
+    };
```
ematipico, line 113: Are these always present even when we don't have the frontmatter?

ematipico (commented):
Can we add a new test where we assert the ranges in case we don't have the frontmatter?

matthewp (approved):

packages/compiler/src/shared/types.ts
Princesseuh: Yes, because there's still a location for where the frontmatter would be, after the prefix, but before the template, it just has the same value for both the start and the end (since it has no content)

Princesseuh:
> Can we add a new test where we assert the ranges in case we don't have the frontmatter?

Added a test!

ematipico (approved):

PR #932, opened 2024-01-02
Prevent printing tagged render function on whitespace only text nodes

MoustaphaDev:
## Changes

https://github.com/withastro/compiler/pull/930 introduced a regression which broke a test added in https://github.com/withastro/compiler/pull/928, but we didn't catch that probably because we didn't update the branch before merging. This fixes the regression and updates tests

## Testing

Update tests to account for recently merged changes

## Docs

N/A bug fix

### DIFF
--- a/.changeset/hot-clocks-judge.md
+++ b/.changeset/hot-clocks-judge.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Fixes a regression that caused whitespace between elements in an expression to result invalid code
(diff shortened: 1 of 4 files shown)


natemoo-re:
Whoops, sorry about that! I probably should have merged these in any other order.

MoustaphaDev:
No worries! I'm working on this now (and getting weird issues where tests work locally and not on ci), but I'll finish the day soon so I'll probably only be able to get a fix by tomorrow.

natemoo-re (approved):

Issue #852, opened 2023-08-15
Meaningful whitespace collapsed with `compressHTML: true`


delucis:
### What version of `@astrojs/compiler` are you using?

1.8.1

### What package manager are you using?

npm

### What operating system are you using?

macOS / StackBlitz

### Describe the Bug

Certain instances of inline whitespace appear to be collapsed with an impact on the visual result when using Astro’s new `compressHTML: true` config option.

Specifically this seems to be the case with expressions (e.g. `{'string'}`) and inline elements (e.g. `<span>` or `<time>`):

#### Input
```astro
<p>{'text'} <span>text</span></p>
<p><span>text</span> <span>text</span></p>
```
#### Actual output
```html
<p>text<span>text</span></p>
<p><span>text</span><span>text</span></p>
```
#### Expected output
```html
<p>text <span>text</span></p>
<p><span>text</span> <span>text</span></p>
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-obrxxu?file=src%2Fpages%2Findex.astro

Princesseuh:
Hello, the Go compiler for Astro is no longer maintained. All future development will be focused on [our new Rust compiler](https://github.com/withastro/compiler-rs). If this issue is still relevant and you can reproduce it on the Rust compiler, please open a new issue there. Thank you!

PR #916, opened 2023-12-20
Update compressHTML logic

natemoo-re:
## Changes

- Account for elements that are known to be whitespace sensitive
- Closes https://github.com/withastro/compiler/issues/852
- This will hopefully fix our whitespace problems.

## Testing

I'd really love more comprehensive tests for this before we merge

## Docs

Bug fix only

PR #937, opened 2024-01-09
Fix regression in expressions

MoustaphaDev:
## Changes

Found the issue on [discord](https://discord.com/channels/830184174198718474/830184175176122389/1194085209369477130)
Fixes a regression introduced in https://github.com/withastro/compiler/pull/930. We were only looking for `nil` or an empty text node in the first child of the expression, instead making sure that all children match that criteria before assuming it's an empty expression

## Testing

Added a printer test

## Docs

N/A bug fix

### DIFF
--- a/.changeset/strong-chairs-walk.md
+++ b/.changeset/strong-chairs-walk.md
@@ -2,4 +2,4 @@
 '@astrojs/compiler': patch
 ---
 
-Fix expression regression when the first child of an expression is whitespace and is followed by any other node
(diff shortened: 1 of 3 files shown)


florian-lefebvre:
Does this fix #852?

MoustaphaDev:
No it's different (I found the issue on discord), it's https://github.com/withastro/compiler/pull/916 that's dealing with it apparently

ematipico (approved):

natemoo-re (approved):
Thanks!

Issue #713, opened 2023-01-18
Parser issue with a specific tag within a curly brace JSX "group" (`Expected ")" but found "}"`)


JulianCataldo:
### What version of `astro` are you using?

1.9.0

### Are you using an SSR adapter? If so, which one?

no

### What package manager are you using?

pnpm

### What operating system are you using?

mac

### Describe the Bug

The exact manoeuvre for correcting what it seems to be a parser bug is contained in this commit:

https://github.com/JulianCataldo/web-garden/commit/36704f9856772024abeb1fdf32f0d059af281948

Moving the `<title>{props.title}</title>` tag from outside a conditional JSX brace group fixed it.
Note that moving it inside the group, up and down, doesn't work.

This component ceased to work suddenly on a fresh project.

PR #942, opened 2024-01-16
Add regression test for #713

MoustaphaDev:
## Changes

- Closes https://github.com/withastro/compiler/issues/713 which seems to have been fixed in 2.4.0!

## Testing

Added a regression test to close the issue

## Docs

N/A bug fix


natemoo-re (approved):
❤️

PR #933, opened 2024-01-03
Fix various issues with slots

MoustaphaDev:
## Changes

- Closes https://github.com/withastro/compiler/issues/724
- Closes https://github.com/withastro/compiler/issues/868
- Partially addresses https://github.com/withastro/astro/issues/8687, the `undefined` slot name issue will be tackled in another PR 
[...]

PR #952, opened 2024-01-29
Fix slot regression when there are multiple expressions

### LINKED ISSUE #950: Issue with slots containing multiple items / conditions
### Astro Info

```block
Astro                    v4.2.3
Node                     v20.10.0
System                   macOS (arm64)
Package Manager          pnpm
Output                   server
Adapter                  none
Integrations             astro-dynamic-import
                         adds-to-head
                         @astrojs/vue
                         astro-purgecss
                         astro-compress
```

_No response_

### Describe the Bug

as of v4.2.2, slots are not longer working correctly. When a slot has a loops, only the last item is being rendered. From what I can tell, its seems to be something to do with wrapping a slot in a condition (see reproduction)

This does not work (only last item in items is rendered)
```
---
import Layout from '../layouts/Layout.astro';
import Test from '../components/Test.vue';

const items = [1,2,3,4,5];
const test = false;

---
<Layout title="Welcome to Astro.">
    Outside
    <Test>
        {test && (
            <p slot={test ? 'test' : undefined}>
                Test name slot
            </p>
        )}
        {items.map(item => (
            <p>Item {item}</p>
        ))}
    </Test>
</Layout>
```
This does work (all items rendered)
```
---
import Layout from '../layouts/Layout.astro';
import Test from '../components/Test.vue';

const items = [1,2,3,4,5];
const test = false;
---

<Layout title="Welcome to Astro.">
    Outside
    <Test>
      <p slot={test ? 'test' : undefined}>
        Test name slot
      </p>
      {items.map(item => (
        <p>Item {item}</p>
      ))}
    </Test>
</Layout>
```

all items to be rendered in correct slots

https://stackblitz.com/edit/github-vzmcxd?file=src%2Fpages%2Findex.astro,src%2Flayouts%2FLayout.astro,src%2Fcomponents%2FTest.vue

### Participation

MoustaphaDev:
## Changes

- Closes #950
- More granular printing of `mergeSlots`

A regression in how we print slot was introduced in my PR in https://github.com/withastro/compiler/pull/933.

Normally, if there is more than one slotted element or any slotted element with dynamic `slot` prop in an expression, we consider it a nested slot. However, instead of doing the count per expression, we aggregated it for all expressions.

Because of that, if there were any slotted element in an expression, any subsequent slotted element of the same kind would be considered a nested slot

To fix that, we localized each of those states per expression

## Testing

- Adapted and added a few printer tests to reflect the changes and ensure a correct output
- Added tests to Astro core to ensure a correct result

## Docs

N/A bug fix only

### DIFF
--- a/internal/printer/print-to-js.go
+++ b/internal/printer/print-to-js.go
@@ -432,7 +432,7 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 		p.printAttributesToObject(n)
 	} else if isSlot {
 		if len(n.Attr) == 0 {
-			p.print(`"default"`)
+			p.print(DEFAULT_SLOT_PROP)
 		} else {
 			slotted := false
 			for _, a := range n.Attr {
@@ -454,7 +454,7 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 				}
 			}
 			if !slotted {
-				p.print(`"default"`)
+				p.print(DEFAULT_SLOT_PROP)
 			}
 		}
 		p.print(`]`)
@@ -559,7 +559,7 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 			switch true {
 			case n.CustomElement:
 				p.print(`,({`)
-				p.print(fmt.Sprintf(`"%s": () => `, "default"))
+				p.print(fmt.Sprintf(`%s: () => `, DEFAULT_SLOT_PROP))
 				p.printTemplateLiteralOpen()
 				for c := n.FirstChild; c != nil; c = c.NextSibling {
 					render1(p, c, RenderOptions{
@@ -638,6 +638,8 @@ func render1(p *printer, n *Node, opts RenderOptions) {
 	}
 }
 
+const DEFAULT_SLOT_PROP = `"default"`
+
 // Section 12.1.2, "Elements", gives this list of void elements. Void elements
 // are those that can't have any contents.
 // nolint
@@ -667,7 +669,7 @@ func handleSlots(p *printer, n *Node, opts RenderOptions, depth int) {
 	numberOfNestedSlots := 0
 
 	for c := n.FirstChild; c != nil; c = c.NextSibling {
-		slotProp := `"default"`
+		slotProp := DEFAULT_SLOT_PROP
 		for _, a := range c.Attr {
 			if a.Key == "slot" {
 				if a.Type == QuotedAttribute {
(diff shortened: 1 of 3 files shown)


MoustaphaDev:
!preview mk-slot-expr-fix

MoustaphaDev:
!preview mk-slot-expr-fix

internal/printer/print-to-js.go
```diff
@@ -686,82 +690,79 @@ func handleSlots(p *printer, n *Node, opts RenderOptions, depth int) {
...
 		if c.Expression {
-			nestedSlotsCount := 0
-			var firstNestedSlotProp string
+			// Only slot ElementNodes (except expressions containing only comments) or non-empty TextNodes!
+			// CommentNode, JSX comments and others should not be slotted
+			if expressionOnlyHasComment(c) {
+				continue
+			}
+			nestedSlotsInExprCount := 0
+			hasAnyDynamicSlotsInExpr := false
```
MoustaphaDev, line 699: This is what the fix is all about: localizing these states per expression. Previously we used a global variables to track them, which caused leaks in other expressions.

The rest is just polishing and refactoring

internal/printer/print-to-js.go
```diff
@@ -907,25 +915,45 @@ func generateEndSlotIndexes(nestedSlotChildren []*NestedSlotChild) map[int]bool
...
+			if len(defaultSlot.Children) > 0 {
+				// if the last element in a slot chain is a default slot
+				// let's add it to the updated nested slot children
+				updatedEndSlotIndexes[len(updatedNestedSlotChildren)] = true
+				updatedNestedSlotChildren = append(updatedNestedSlotChildren, defaultSlot)
+				defaultSlot = &NestedSlotChild{SlotProp: DEFAULT_SLOT_PROP, Children: []*Node{}}
+			} else if !isDefault {
+				// if it's not a default slot, we just need to set the updated end slot indexes
+				// we already added it in the previous iteration
+				updatedEndSlotIndexes[len(updatedNestedSlotChildren)-1] = true
```
ematipico, line 947: Do you know if this operation is safe? Are we sure that `updatedNestedSlotChildren` has at least one element at this point? Is the lenght of `updatedEndSlotIndexes` high enough to store this new information? 

Sorry, I have minimal context so my questions my sound dumb

internal/printer/print-to-js.go
ematipico: Should we add a test with comments too?

MoustaphaDev: Ah, I should've clarified that we already have tests for that. I just moved the code above to do the check earlier

natemoo-re (approved):

ematipico (approved):

PR #954, opened 2024-01-30
Fix expression in `th` causing infinite loop

MoustaphaDev:
## Changes

Fix [an issue reported on discord](https://discord.com/channels/830184174198718474/1199343177446998116/1199343177446998116) in which if there were an expression in a `th` element, that would cause an infinite loop
This regression was introduced in 2.4.0, but I'm not sure what exactly caused it

## Testing

Added printer test to make sure that it doesn't hang and it's printed correctly

## Docs

N/A bug fix only

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -2730,6 +2730,10 @@ func inLiteralIM(p *parser) bool {
 }
 
 func inExpressionIM(p *parser) bool {
+	// if p.oe.contains(a.Table) {
+	// 	p.clearActiveFormattingElements()
+	// 	return inLiteralIM(p)
+	// }
 	switch p.tok.Type {
 	case ErrorToken:
 		p.oe.pop()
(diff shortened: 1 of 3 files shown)


internal/parser.go
```diff
@@ -2730,6 +2730,10 @@ func inLiteralIM(p *parser) bool {
 }
 
 func inExpressionIM(p *parser) bool {
+	// if p.oe.contains(a.Table) {
```
matthewp, line 2733: let's remove these lines if no longer relevant

natemoo-re (approved):
D’oh. Thanks!

internal/parser.go
MoustaphaDev: Ah yeah, forgot to remove that after debugging the issue

MoustaphaDev: Done in [`7f82f6b`](https://github.com/withastro/compiler/pull/954/commits/7f82f6be742414f534d575074863dcf8082403d3) !

PR #965, opened 2024-02-07
Adds lint rules for `data-astro-reload` and `data-astro-rerun` on script tags

martrapp:
## Changes

Depends on https://github.com/withastro/astro/pull/9977

~~Supports `transition:reload` as a nice Astro way to set `data-astro-reload`.~~
Warns if `data-astro-*`attributes are used the wrong way during view transitions.

For some time now, you can set `data-astro-reload` to a <a> tag to skip view transitions in favor of full page loads. https://github.com/withastro/astro/pull/9977 will also use ~~data-astro-reload~~ `data-astro-rerun` to mark `<scripts>` for re-execution during view transitions. 

### Update:

Since we have now separated `data-astro-reload` and `data-astro-rerun` (for scripts), I would stay with the data attributes and prefer not to introduce any additional directives.

The original warnings about transition:reload ~~were unnecessary as they are better covered by packages/astro/astro-jsx.d.ts~~ are carried over.

## Testing

added test case

## Docs

~~I will open a PR if this one gets accepted, see https://github.com/withastro/docs/pull/6856~~


internal/printer/printer.go
```diff
@@ -512,6 +514,32 @@ func maybeConvertTransition(n *astro.Node) {
 			})
 		}
 	}
+	if transform.HasAttr(n, transform.TRANSITION_RELOAD) {
+		transitionReloadIndex := transform.AttrIndex(n, transform.TRANSITION_RELOAD)
+		attr := n.Attr[transitionReloadIndex]
+
+		if n.Type == astro.ElementNode &&
```
matthewp, line 521: could you add a comment here for which elements are allowed to have this directive?

MoustaphaDev:
I noticed that `attr` isn't declared as a pointer. Since we're modifying the key in line 533, you might want to receive a pointer to ensure the changes take effect.

```suggestion
		attr := &n.Attr[transitionReloadIndex]
```

Perhas adding a test to ensure its presence like this one would be a good idea?https://github.com/withastro/compiler/blob/c17734d1607d83cd357a5b6204f33fd37ebe1782/internal/printer/printer_test.go#L3412-L3419

internal/printer/printer.go
```diff
@@ -512,6 +514,32 @@ func maybeConvertTransition(n *astro.Node) {
 			})
 		}
 	}
+	if transform.HasAttr(n, transform.TRANSITION_RELOAD) {
+		transitionReloadIndex := transform.AttrIndex(n, transform.TRANSITION_RELOAD)
+		attr := n.Attr[transitionReloadIndex]
```
MoustaphaDev, line 519: I noticed that `attr` isn't declared as a pointer. Since we're modifying the key in line 533, you might want to receive a pointer to ensure the changes take effect.

```suggestion
		attr := &n.Attr[transitionReloadIndex]
```

Perhas adding a test to ensure its presence like this one would be a good idea?https://github.com/withastro/compiler/blob/c17734d1607d83cd357a5b6204f33fd37ebe1782/internal/printer/printer_test.go#L3412-L3419

martrapp: oh dear, thank you very much for this hint, good catch!

matthewp: Pointers are fun :)

internal/printer/printer.go
```diff
@@ -512,6 +514,50 @@ func maybeConvertTransition(n *astro.Node) {
 			})
 		}
 	}
+	if transform.HasAttr(n, transform.TRANSITION_RELOAD) {
+		attr := &n.Attr[transform.AttrIndex(n, transform.TRANSITION_RELOAD)]
+
+		/*
+			For supported elements, the transition:reload directive is translated 1:1 into a data-astro-reload attribute.
+			When set on <a>, <form> and <area>, it replaces the transitions between views with the full page load.
+			If set on a supported <script> element, the script will be executed again during the next view transition, even if it was already part of the previous page. All scripts are supported, except external ECMAScript modules.
```
martrapp, line 523:
```suggestion
			If set on a supported <script> element, the script will be executed again during the next view transition, 
			even if it was already part of the previous page. All scripts are supported, except external ECMAScript modules.
```

internal/printer/printer.go
```diff
@@ -512,6 +514,51 @@ func maybeConvertTransition(n *astro.Node) {
 			})
 		}
 	}
+	if transform.HasAttr(n, transform.TRANSITION_RELOAD) {
+		attr := &n.Attr[transform.AttrIndex(n, transform.TRANSITION_RELOAD)]
+
+		/*
+			For supported elements, the transition:reload directive is translated 1:1 into a data-astro-reload attribute.
+			When set on <a>, <form> and <area>, it replaces the transitions between views with the full page load.
```
martrapp, line 522:
```suggestion
			When set on <a>, <form> and <area>, it replaces the view transitions with a full page load.
```

internal/printer/printer_test.go
```diff
@@ -3425,6 +3425,18 @@ const items = ["Dog", "Cat", "Platipus"];
 				code: `${$$maybeRenderHead($$result)}<div data-astro-transition-persist="foo"${$$addAttribute($$renderTransition($$result, "peuy4xf7", "", "foo"), "data-astro-transition-scope")}></div>`,
 			},
 		},
+		{
+			name: "transition:reload is converted to a data attribute",
+			source: `
+			---
```
martrapp, line 3431: transition:reload does not create a transition scope.
But the test code below checks for "transition:" in the 'source' and insists on that import when present. 
Maybe that "if" should be changed to 'if tt.transitions {' (or at least be guarded by tt.transitions)

.changeset/orange-knives-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/compiler': minor
```
martrapp, line 2: Am I right that this a minor?

internal/printer/printer.go
martrapp: Done!

internal/printer/printer.go
martrapp: @MoustaphaDev test added to printer_test.go

.changeset/orange-knives-smile.md
matthewp: Yep!

matthewp (approved):

martrapp:
!preview transition-reload

internal/printer/printer_test.go
MoustaphaDev: I agree, guarding with `tt.transitions` makes more sense

internal/printer/printer.go
MoustaphaDev: Awesome, love the refactor to an early return pattern too 🤩!

internal/printer/printer.go
```diff
@@ -512,6 +514,51 @@ func maybeConvertTransition(n *astro.Node) {
...
+		if attr.Val != "" {
+			h.AppendWarning(&loc.ErrorWithRange{
+				Code:  loc.WARNING_UNSUPPORTED_EXPRESSION,
+				Text:  "transition:reload does not accept a value",
+				Hint:  fmt.Sprintf("Use transition:reload without \"%s\"", attr.Val),
+				Range: loc.Range{Loc: attr.ValLoc, Len: len(attr.Val)},
+			})
+			return
+		}
+		if n.Data == "script" && transform.HasAttr(n, "src") && transform.HasAttr(n, "type") {
```
MoustaphaDev, line 546: Seems like this won't catch inline module scripts, is that intended?

MoustaphaDev (approved):
Looks good to me!

matthewp (approved):

MoustaphaDev (approved):

martrapp:
This one depends on https://github.com/withastro/astro/pull/9977
So i will merge with/after v4.5

sarah11918 (approved):
These are super well formatted and read wonderfully, @martrapp ! 🙌 

The only thing I wonder is whether "external ECMAScript Modules" is commonly used/understood, or a tiny bit too jargony.

For example, if "`<script>` tags" (or "x kind of `<script>` tags") is accurate and would work in its place, you know me, I'd totally say that! If you are confident that this phrase is "how people would say it when working in their project", then I'm happy! If there's a more common phrase that people would say, e.g. talking about their code in Discord, then I'd probably tend to use that.

Otherwise, docs is happy!

martrapp:
Haha, thank you for pointing that out!
Yes, the "ECMAScript Module" is a bit bulky. I would typically use ESM in that context, but I do not like to use abbreviations when I can't properly introduce them. I guess we go for "module scripts" as it becomes clear from the context.

natemoo-re (approved):
Great work!

PR #969, opened 2024-02-13
Add renderScript option

bluwy:
## Changes

This PR allows rendering non-inline script tags using a runtime API from `internalURL`, e.g.

```html
<script>console.log("a")</script>
```

becomes:

```js
`${$$renderScript($$metadata, "/src/pages/index.astro?astro&type=script&index=0&lang.ts")}`
```

Personally I'm not a fan of the hardcoded script URL default, but the styles were also hardcoded this way, and it was easier to implement.

## Testing

Added go tests

## Docs

Only added jsdoc to the types explaining this feature. Options added in the past didn't have public docs either (?)

### DIFF
--- a/.changeset/gold-apples-shout.md
+++ b/.changeset/gold-apples-shout.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': minor
+---
+
+Adds a new `renderScript` option to render non-inline script tags using a `renderScript` function from `internalURL`, instead of stripping the script entirely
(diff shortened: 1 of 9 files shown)


bluwy:
!preview render-script

ematipico:
What's `internalURL`? I lack some context

bluwy:
`internalURL` is one of the compiler option names. We pass something like `astro/compiler-runtime`.

MoustaphaDev (approved):

ematipico:
Let's 🚢  it

bluwy:
Thanks for the reviews! I'll merge this in for now then and we could get the other PRs in to cut a new minor together.

PR #970, opened 2024-02-16
Add lint rule for implicit `is:inline` on script tags.

lilnasy:
## Changes

- https://github.com/withastro/roadmap/blob/main/proposals/0016-style-script-defaults.md?plain=1#L74

## Testing
Added `scripts/isinline-hint.ts`.

## Docs
Would like docs review on the message.

### DIFF
--- a/.changeset/cuddly-pumas-call.md
+++ b/.changeset/cuddly-pumas-call.md
@@ -2,4 +2,4 @@
 '@astrojs/compiler': patch
 ---
 
-Adds a lint rules when attributes are added to a script tag, causing the script to be treated as `is:inline`.
+Adds a lint rule to display a message when attributes are added to a script tag, explaining that the script will be treated as `is:inline`.
--- a/.changeset/cuddly-pumas-call.md
+++ b/.changeset/cuddly-pumas-call.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Adds a lint rules when attributes are added to a script tag, causing the script to be treated as `is:inline`.
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -412,7 +412,7 @@ func HintAboutImplicitInlineDirective(n *astro.Node, h *handler.Handler) {
 		}
 		h.AppendHint(&loc.ErrorWithRange{
 			Code:  loc.HINT,
-			Text:  "This script will be treated as if it has the `is:inline` directive, because it contains an attribute. Therefore features that require processing (e.g. using TypeScript or npm packages in the script) are unavailable.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.",
+			Text:  "This script will be treated as if it has the `is:inline` directive, because it contains an attribute. Therefore, features that require processing (e.g. using TypeScript or npm packages in the script) are unavailable.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.",
 			Range: loc.Range{Loc: n.Attr[0].KeyLoc, Len: len(n.Attr[0].Key)},
 		})
 	}
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -412,7 +412,7 @@ func HintAboutImplicitInlineDirective(n *astro.Node, h *handler.Handler) {
 		}
 		h.AppendHint(&loc.ErrorWithRange{
 			Code:  loc.HINT,
-			Text:  "This script will be treated as if it has the `is:inline` directive, because it contains an attribute. Therefore, features that require processing (e.g. using TypeScript or npm packages in the script) are unavailable.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.",
+			Text:  "This script will be treated as if it has the `is:inline` directive, because it contains an attribute. Therefore, features that require processing (e.g. using TypeScript or npm packages in the script) are unavailable.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.\n\nAdd the `is:inline` directive explicitly to silence this hint.",
 			Range: loc.Range{Loc: n.Attr[0].KeyLoc, Len: len(n.Attr[0].Key)},
 		})
 	}
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -40,6 +40,7 @@ func Transform(doc *astro.Node, opts TransformOptions, h *handler.Handler) *astr
 	i := 0
 	walk(doc, func(n *astro.Node) {
 		i++
+		HintAboutImplicitInlineDirective(doc, n, &opts, h)
 		ExtractScript(doc, n, &opts, h)
 		AddComponentProps(doc, n, &opts)
 		if shouldScope {
@@ -401,6 +402,17 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *ha
 				}
 			}
 		}
+	}
+}
+
+func HintAboutImplicitInlineDirective(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *handler.Handler) {
+	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && !HasInlineDirective(n) {
+		h.AppendHint(&loc.ErrorWithRange{
+			Code:  loc.HINT,
+			Text:  "Astro processes your script tags to allow using TypeScript and NPM packages, and to optimize browser performance.\n\nAttributes cannot be used on Astro-processed script tags. Therefore, this script tag will be treated as if it has the `is:inline` directive, opting it out of the processing steps and its features.\n\nFor clarity, you might want to add the `is:inline` directive explicitly.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.",
+			Range: loc.Range{Loc: n.Attr[0].KeyLoc, Len: len(n.Attr[0].Key)},
+		})
+
 	}
 }
 
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -406,13 +406,15 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *ha
 }
 
 func HintAboutImplicitInlineDirective(n *astro.Node, h *handler.Handler) {
-	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && (!HasInlineDirective(n) || len(n.Attr) == 1 && n.Attr[0].Key != "src") {
+	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && !HasInlineDirective(n) {
+		if len(n.Attr) == 1 && n.Attr[0].Key == "src" {
+			return
+		}
 		h.AppendHint(&loc.ErrorWithRange{
 			Code:  loc.HINT,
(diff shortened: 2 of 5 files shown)


internal/transform/transform.go
```diff
@@ -404,6 +405,17 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *ha
 	}
 }
 
+func HintAboutImplicitInlineDirective(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *handler.Handler) {
+	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && !HasInlineDirective(n) {
+		h.AppendHint(&loc.ErrorWithRange{
+			Code:  loc.HINT,
+			Text:  "Astro processes your script tags to allow using TypeScript and NPM packages, and to optimize browser performance.\n\nAttributes cannot be used on Astro-processed script tags. Therefore, this script tag will be treated as if it has the `is:inline` directive, opting it out of the processing steps and its features.\n\nFor clarity, you might want to add the `is:inline` directive explicitly.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.",
```
lilnasy, line 412: image
image

natemoo-re (commented):
Approving the code change! Will leave decisions on the exact language to others, but this looks pretty good. Maybe a bit long for a lint error.

natemoo-re (approved):

internal/transform/transform.go
```diff
@@ -404,6 +405,17 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *ha
 	}
 }
 
+func HintAboutImplicitInlineDirective(n *astro.Node, h *handler.Handler) {
+	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && !HasInlineDirective(n) {
+		h.AppendHint(&loc.ErrorWithRange{
+			Code:  loc.HINT,
+			Text:  "Astro processes your script tags to allow using TypeScript and NPM packages, and to optimize browser performance.\n\nAttributes cannot be used on Astro-processed script tags. Therefore, this script tag will be treated as if it has the `is:inline` directive, opting it out of the processing steps and its features.\n\nFor clarity, you might want to add the `is:inline` directive explicitly.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.",
```
ematipico, line 412:
```suggestion
			Text:  "Astro processes your script tags to allow using TypeScript and npm packages, and to optimize browser performance.\n\nAttributes cannot be used on Astro-processed script tags. Therefore, this script tag will be treated as if it has the `is:inline` directive, opting it out of the processing steps and its features.\n\nFor clarity, you might want to add the `is:inline` directive explicitly.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.",
```

The name is `npm`

ematipico (approved):

internal/transform/transform.go
```diff
@@ -404,6 +405,17 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *ha
 	}
 }
 
+func HintAboutImplicitInlineDirective(n *astro.Node, h *handler.Handler) {
+	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && !HasInlineDirective(n) {
```
MoustaphaDev, line 409: Script tags only containing `src` also get processed, so we should check for that too
```suggestion
	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && len(n.Attr) == 1 && n.Attr[0].Key != "src" && !HasInlineDirective(n) {
```

lilnasy: Oh yeah, that's right!

internal/transform/transform.go
```diff
@@ -404,6 +405,17 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *ha
 	}
 }
 
+func HintAboutImplicitInlineDirective(n *astro.Node, h *handler.Handler) {
+	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && len(n.Attr) == 1 && n.Attr[0].Key != "src" && !HasInlineDirective(n) {
```
MoustaphaDev, line 409: Whoops, just realized the logic isn't correct after posting the comment 😅
This should be correct, hopefully
```suggestion
	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && (!HasInlineDirective(n) || len(n.Attr) == 1 && n.Attr[0].Key != "src") {
```

lilnasy: Thanks, I'll add a test for external scripts soon as well.

MoustaphaDev: Awesome!

internal/transform/transform.go
```diff
@@ -404,6 +405,19 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *ha
...
 }
 
+func HintAboutImplicitInlineDirective(n *astro.Node, h *handler.Handler) {
+	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && !HasInlineDirective(n) {
+		if len(n.Attr) == 1 && n.Attr[0].Key == "src" {
+			return
+		}
+		h.AppendHint(&loc.ErrorWithRange{
+			Code:  loc.HINT,
+			Text:  "Astro processes your script tags to allow using TypeScript and npm packages, and to optimize browser performance.\n\nAttributes cannot be used on Astro-processed script tags. Therefore, this script tag will be treated as if it has the `is:inline` directive, opting it out of the processing steps and its features.\n\nFor clarity, you might want to add the `is:inline` directive explicitly.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.",
```
sarah11918, line 415:
```suggestion
			Text: "This script contains an attribute (ex: `type="module"`, `type="text/partytown"`, `async`) or a directive (`define:vars`) and will not be processed and bundled for optimized browser performance.\n\nThis script will be treated as if it has the `is:inline` directive and therefore features that require processing (e.g. using TypeScript or npm packages in the script) are unavailable.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.",
```
This felt maybe a little "life story" for a helpful hint message? If this is a warning message shown when an attribute exists (because people might not realize it's gonna be inline), then I'd frame the ideas in something like the order above: 
- why this message is being shown (we found an attribute/directive!)
- what that means (no processing/optimizing for you!)
- why it means that/how they can understand why this is happening (it's treated like inline!)
- any other important information to know (no other features, too. See the docs)

I'm also not sure why the recommendation to add the `is:inline` directive explicitly. What's the benefit? What kind of clarity are you going for? Just to remind yourself that this one will be treated as inline? Unless there's a strong reason to suggest that I'm not seeing, I'd recommend removing.

.changeset/cuddly-pumas-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Adds a lint rules when attributes are added to a script tag, causing the script to be treated as `is:inline`.
```
sarah11918, line 5:
```suggestion
Adds a lint rule to display a message when attributes are added to a script tag, explaining that the script will be treated as `is:inline`.
```
Something like this maybe?

sarah11918 (commented):
I think this is a great idea to help people understand the effect of attributes in their script tags! Some thoughts on the wordingt for you @lilnasy ! 🙌

internal/transform/transform.go
martrapp: From Sarah's suggestion:
> This script contains an attribute (ex: `type="module"`, `type="text/partytown"`, `async`) or a directive (`define:vars`) and ...

Instead of giving examples that might have nothing to do with the user's code and might confuse them, we could make this hint dynamic and include current attributes.

martrapp: > you might want to add the `is:inline` directive explicitly

I think this information is valuable as it shows you how to disable the hint (and remove the three dots in the editor view)

lilnasy: > shows you how to disable the hint

That's exactly right. It was also part of the RFC that introduced `is:inline`.

> I'm also not sure why the recommendation to add the is:inline directive explicitly. 

I guess we wouldn't want to encourage `is:inline`, because it is almost always an accident. Is that what you're getting at?

sarah11918: I was getting at, I have no idea why this is a recommendation for this case, if it's already treated as inline. Why would you bother adding it?

Martin's response that adding it will *silence the hint/message* is a valid reason, so if it's helpful then something like:

> Adding the `is:inline` explicitly will silence this hint.

BUT, that seems like maybe kind of terrible DX -- you have to go back and add this everywhere or you're going to get editor notation yelling at you for things?

lilnasy: I should clarify, the motivation behind this PR is to add friction to the cases where `is:inline` is added implicitly.

It is usually accidental, it makes  the scripts behave drastically differently, and is only required in very specific use-cases.

Moreover, this "discouragement" was part of the original RFC from years ago.

internal/transform/transform.go
```diff
@@ -404,6 +405,19 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *ha
...
 }
 
+func HintAboutImplicitInlineDirective(n *astro.Node, h *handler.Handler) {
+	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && !HasInlineDirective(n) {
+		if len(n.Attr) == 1 && n.Attr[0].Key == "src" {
+			return
+		}
+		h.AppendHint(&loc.ErrorWithRange{
+			Code:  loc.HINT,
+			Text:  "This script will be treated as if it has the `is:inline` directive, because it contains an attribute. Therefore, features that require processing (e.g. using TypeScript or npm packages in the script) are unavailable.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.",
```
lilnasy, line 415:
```suggestion
			Text:  "This script will be treated as if it has the `is:inline` directive, because it contains an attribute. Therefore, features that require processing (e.g. using TypeScript or npm packages in the script) are unavailable.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.\n\nAdd the `is:inline` directive explicitly to silence this hint.",
```

internal/transform/transform.go
```diff
@@ -404,6 +405,19 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *ha
...
 }
 
+func HintAboutImplicitInlineDirective(n *astro.Node, h *handler.Handler) {
+	if n.Type == astro.ElementNode && n.DataAtom == a.Script && len(n.Attr) > 0 && !HasInlineDirective(n) {
+		if len(n.Attr) == 1 && n.Attr[0].Key == "src" {
+			return
+		}
+		h.AppendHint(&loc.ErrorWithRange{
+			Code:  loc.HINT,
+			Text:  "This script will be treated as if it has the `is:inline` directive, because it contains an attribute. Therefore, features that require processing (e.g. using TypeScript or npm packages in the script) are unavailable.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.\n\nAdd the `is:inline` directive explicitly to silence this hint.",
```
sarah11918, line 415:
```suggestion
			Text:  "This script will be treated as if it has the `is:inline` directive because it contains an attribute. Therefore, features that require processing (e.g. using TypeScript or npm packages in the script) are unavailable.\n\nSee docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.\n\nAdd the `is:inline` directive explicitly to silence this hint.",
```

sarah11918 (approved):
Docs approves once you remove the extraneous comma! :rocket:

sarah11918:
Just checking, because I don't know the required syntax here, but in some of these there are two spaces, and in another, only one. Just in case that matters! 
image

sasoria:
I'll add that @lilnasy did this pull request live with other community members. I had a lot of fun watching him do it and learned a lot.

lilnasy:
@sarah11918 Good eye, I hadn't noticed!

Go has formatting built-in, it will take care of it on its own sooner or later. Doesn't affect behavior.

PR #975, opened 2024-03-05
Implement transition:persist-props

matthewp:
## Changes

- Implements `transition:persist-props` as explained in https://github.com/withastro/astro/pull/10136
- This is a simple transformation to `data-astro-transition-persist-props` with the same value.
  - Very similar to how `transition:persist` works.
- This means for dynamic use-cases users can use the `data-` attribute instead.

## Testing

- New test case added

## Docs

- https://github.com/withastro/docs/pull/6915

### DIFF
--- a/.changeset/real-otters-mix.md
+++ b/.changeset/real-otters-mix.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': minor
+---
+
+Implement the `transition:persist-props` transformation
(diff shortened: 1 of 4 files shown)


martrapp (commented):
Is this directive intentionally not included in `skippedAttributes` and `skippedAttributesToObject`?

matthewp:
@martrapp It's renamed in the transform step, so it doesn't exist when it gets to the printer and doesn't need to be skipped.

martrapp (approved):
Looks good to me!

PR #972, opened 2024-02-28
Updates deprecated Node.js 16 github actions

martrapp:
## Changes

The plan is to update all action that log deprecation warnings starting with pnpm/action-setup.
I will keep this PR open as long as I try to cover the most common workflows.

- Updated actions/setup-go@v2 to v5
- Updated actions/checkout@v2/v3 to V4
- Updated pnpm/action-setup@v2 to v3
- Update actions/setup-node@v2 to v4
- Update golangci/golangci-lint-action@v2 to v4
- Update actions/github-script@v6 to v7

> - No new version for lannonbr/repo-permission-check-action@2.0.0
> - No new version for eficode/resolve-pr-refs@main

Not addressed in this PR:
- The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files.
- The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. 

## Testing

scan github warnings, update versions, trigger github actions, repeat

## Docs

n.a.

PR #977, opened 2024-03-08
chore: use better maintained permission checker action 

MoustaphaDev:
## Changes

Drop `lannonbr/repo-permission-check-action` for [`actions-cool/check-user-permission`](https://github.com/actions-cool/maintain-one-comment)
The earlier wasn't working anymore [after we updated it](https://github.com/withastro/compiler/pull/972/files#diff-9dc5e53b8fc4bd9d74ca8c2d8d9d5bf7f94c6a8ccc49f2bcc142ec5c171686d1R19).
[Martin opened an issue](https://github.com/actions-cool/maintain-one-comment) on that repo to suggest updating to Node 20 (to remove the deprecation warnings) but I'm not sure it's gonna move forward as the repo looks abandoned.

I checked, and the main repo doesn't currently use the bugged version of the action, but to get rid of the warnings, we shall make the switch to the better maintained one.

## Testing

Not tested but the usage looked straightforward after looking at the docs and how other repos use it

## Docs

N/A just a chore, doesn't touch the code

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - name: Check if user has admin access (only admins can publish snapshot releases).
         id: checkAccess
-        uses: actions-cool/check-user-permission@v2.2.1
+        uses: actions-cool/check-user-permission@v2
         with:
           require: admin
           username: ${{ github.triggering_actor }}
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -15,12 +15,21 @@ jobs:
     if: ${{ github.repository_owner == 'withastro' && github.event.issue.pull_request && startsWith(github.event.comment.body, '!preview') }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


ematipico (approved):

.github/workflows/snapshot-release.yml
```diff
@@ -15,12 +15,21 @@ jobs:
     if: ${{ github.repository_owner == 'withastro' && github.event.issue.pull_request && startsWith(github.event.comment.body, '!preview') }}
     runs-on: ubuntu-latest
     steps:
-      - name: 'Check if user has admin access (only admins can publish snapshot releases).'
-        uses: 'lannonbr/repo-permission-check-action@2.0.2'
+      - name: Check if user has admin access (only admins can publish snapshot releases).
+        id: checkAccess
+        uses: actions-cool/check-user-permission@v2.2.1
```
bluwy, line 20: Should we use 
```suggestion
        uses: actions-cool/check-user-permission@v2
```

instead so we can get patch updates when available?

MoustaphaDev: Ah good catch!

PR #979, opened 2024-03-08
chore: fix exit code for failed permission check

MoustaphaDev:
## Changes

The snapshot workflow in https://github.com/withastro/compiler/pull/977 returned `0` instead of `1`, causing the workflow to not fail for non-admins.

## Testing

Not tested but the only issue was the wrong exit code which wouldn't stop the workflow

## Docs

N/A doesn't affect users


ematipico (approved):

MoustaphaDev:
!preview test-admin-snapshot

PR #991, opened 2024-03-29
Make snapshot releases build the wasm

matthewp:
## Changes

- Snapshot releases weren't building wasm. They should

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -71,9 +71,6 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
           cache: 'pnpm'
 
-      - name: Build WASM
-        run: make wasm
-
       - name: Install dependencies
         run: pnpm install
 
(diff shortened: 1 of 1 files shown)


MoustaphaDev (approved):
Ah that makes sense, let's hope that fixes it 🤞

.github/workflows/snapshot-release.yml
```diff
@@ -56,6 +56,11 @@ jobs:
         with:
           ref: ${{ steps.refs.outputs.head_ref }}
 
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.19
```
ematipico, line 62: Is there a way to tell the action to use the version of `go.mod`?

Fryuni: Yes, there is: https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

PR #993, opened 2024-04-02
chore: upgrade to go 1.22

lilnasy:
## Changes
- Updates the entrypoint to reflect renaming of the imports object in 1.20.

## Testing
- Existing tests should pass.

## Docs
- Does not affect usage.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/withastro/compiler
 
-go 1.19
+go 1.20
 
 require (
 	github.com/google/go-cmp v0.5.9
--- a/packages/compiler/src/node/wasm_exec.ts
+++ b/packages/compiler/src/node/wasm_exec.ts
@@ -165,7 +165,7 @@ export default class Go {
 
     const timeOrigin = Date.now() - performance.now();
     this.importObject = {
-      go: {
+      gojs: {
         // Go's SP does not change as long as no Go code is running. Some operations (e.g. calls, getters and setters)
         // may synchronously trigger a Go event handler. This makes Go code get executed in the middle of the imported
         // function. A goroutine can switch to a new stack if the current stack is too small (see morestack function).
--- a/.changeset/tender-rockets-agree.md
+++ b/.changeset/tender-rockets-agree.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': minor
+---
+
+The WASM binaries for the compiler are now built using Go 1.22.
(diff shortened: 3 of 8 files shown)


packages/compiler/src/node/wasm_exec.ts
```diff
@@ -165,7 +165,7 @@ export default class Go {
 
     const timeOrigin = Date.now() - performance.now();
     this.importObject = {
-      go: {
+      gojs: {
```
lilnasy, line 168: This is the fix.

ematipico (approved):

ematipico (commented):
Should we update `go.mod` to use `1.22`?

lilnasy:
Yes, we should.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/withastro/compiler
 
-go 1.19
+go 1.20
```
ematipico, line 3:
```suggestion
go 1.22
```

lilnasy: `go.mod` is just the minimum version

ematipico: Why do we require a minimum version, but then build the program in a more recent version? I don't understand

lilnasy: If a contributor has 1.20 installed, it would work the same as 1.22. It's not necessary to make them upgrade their toolchain.

lilnasy: What's the minimum go version needed by the testing library you wanted to use?

ematipico: This left me with more doubts and questions:
- we test and **compile** against a different target. What could work locally might not work on CI and vice versa
- we don't release go libs, why do we require different versions?
- At the moment we all point to 1.19, but now we are branching. Does the contribution guide explain this?

matthewp (changes requested):
If we're going to upgrade the go version we should test against core tests before doing so. It has broken in the past. I would suggest:

1. Creating a preview release here.
2. Creating a PR in core with the compiler version as the preview release version.
3. If tests pass we are golden and can proceed.

lilnasy:
!preview go122

lilnasy:
!preview go122

lilnasy:
All good

natemoo-re (approved):
Thank you! Preview release passed === LGTM 🫡

go.mod
ematipico: Can you answer my questions?

The lib that I need requires 1.21

PR #1001, opened 2024-04-12
fix: prevent printing `defineVars()` into an inline style

### LINKED ISSUE #1000: `define:vars` not working with `is:inline` directive in `<style>` tags
### Astro Info

```block
Astro                    v4.5.9
Node                     v18.18.0
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

According to the Astro documentation, the `define:vars` directive should allow passing server-side variables from the component frontmatter into the `<style>` tags. The documentation does not mention any specific limitations or incompatibilities between `define:vars` and `is:inline` for <style> tags.

When using the `define:vars` directive in a `<style>` tag that also has the `is:inline` directive, the variables passed through `define:vars` do not seem to be properly injected or accessible within the styles. The styles do not apply as expected, and there is an error message but goes away once the page is reloaded.

https://stackblitz.com/edit/github-zbqqwf?file=src%2Fpages%2Findex.astro

### Participation

lilnasy:
## Changes
- Closes #1000
- Removed dependence on make. I don't have it installed, I have always had to manually copy and run the command.

## Testing
- Added go test

## Docs
- Does not affect usage.


ematipico (approved):

package.json
```diff
@@ -7,7 +7,7 @@
     "url": "https://github.com/withastro/compiler.git"
   },
   "scripts": {
-    "build": "make wasm",
+    "build": "CGO_ENABLED=0 GOOS=js GOARCH=wasm go build \"-ldflags=-s -w\" -trimpath -o ./packages/compiler/wasm/astro.wasm ./cmd/astro-wasm/astro-wasm.go",
```
matthewp, line 10: what's this about?

lilnasy: So that it isn't necessary to install `make` to contribute to the project, just pnpm and go.

jacobdalamb:
Can this be merged in the next release please.

Princesseuh:

PR #1002, opened 2024-04-18
Skip printing `createAstro` code if unused

bluwy:
## Changes

Idea from https://github.com/withastro/astro/pull/10773#pullrequestreview-2005386326

If the `Astro` global is not used, we don't need to print code that calls `createAstro`. This saves on runtime execution as it's created but not used.

---

Perf:

- `createAstro()` calls dropped from 571ms o 551ms. -0.02s (not a lot, but it helps other cases instead like garbage collection and AST parsing)
- `garbage collection` dropped from  59.52s to 40.88s. -18.7s

## Testing

Updated tests to handle cases where the `createAstro` code is no longer present.

## Docs

n/a. internal improvement.

### DIFF
--- a/.changeset/khaki-ways-wait.md
+++ b/.changeset/khaki-ways-wait.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Skips printing `createAstro` code if the `Astro` global is not referenced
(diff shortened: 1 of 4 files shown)


matthewp (approved):
great idea!

natemoo-re (approved):
Very awesome idea, nice work!

ematipico (approved):

PR #1004, opened 2024-04-30
Skip printing async if await is not used

bluwy:
## Changes

Idea by @natemoo-re. Currently we always print the component function with async even if no top-level await is used:

```js
const $$stdin = $$createComponent(async ($$result, $$props, $$slots) => {
```

We can skip printing the `async` if no `await` is used so that rendering the component won't always require one tick.

---

Performance:

This should improve page rendering. With the Astro docs repo, it fluctuates between 35-42s a lot, so it's hard to make a concrete comparison. But compared to the old timings, the average timing dropped about 1-2s.

## Testing

Updated test

## Docs

n/a. internal improvement.

### DIFF
--- a/.changeset/warm-zebras-rest.md
+++ b/.changeset/warm-zebras-rest.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Skips printing `async` for component functions if `await` is not used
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):
LGTM! Glad to see that this idea does have some small gains.

PR #1008, opened 2024-06-10
Add server island metadata

matthewp:
## Changes

- Discovers `server:defer` directive usage, then links that back to the import statement like what is done with `client:` islands. This information is exported as part of the compile result, on the `serverComponents` field, to be used by Astro.

## Testing

- Since the only visible change is to the metadata, this is tested by adding a wasm test, to verify that the information is included and is correct.

## Docs

N/A

### DIFF
--- a/.changeset/nine-ants-bow.md
+++ b/.changeset/nine-ants-bow.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Add localName to serverComponents output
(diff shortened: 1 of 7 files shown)


matthewp:
!preview server-islands

matthewp:
!preview server-islands

matthewp:
!preview server-islands

ematipico (commented):
Can you also create some snapshot tests in the go testing suite? It would be nice to have some safety against regressions

matthewp:
@ematipico this doesn't change the compiled output JS at all. It just adds the metadata that's only retrieved from the JS compiler API. That's why the tests are there.

Princesseuh (approved):
Code looks good to me!

PR #1011, opened 2024-06-19
chore: cover all tests

ematipico:
## Changes

This PR adds snapshot testing to ALL tests.

I also changed a bit the way we create snapshot testing, I created an utility called `MakeSnapshot` inside `test_utils.go`, which accepts certain options to generate the snapshots.

Before this PR, the snapshots were stored inside a `__snapshots__` folder, but I preferred to configure that value, and pass a specific folder for each test suite we're testing. I found dividing snapshots by test case to be nice.

## Testing

CI should pass

## Docs

N/A


bluwy:
Do we need snapshot testing for all tests? Seems like these are rather short and can be asserted entirely with a single string.  The features are also well-defined and isolated (and unlikely to have internal implementation change).

I think it's fine to leave it as is, and perhaps leave snapshots for larger file outputs only that's hard to assert and susceptible to internal implementation change.

ematipico:
> Do we need snapshot testing for all tests? Seems like these are rather short and can be asserted entirely with a single string. The features are also well-defined and isolated (and unlikely to have internal implementation change).

That's my question from me to you, as you have more knowledge than me, background-wise. I hope you can tell me which tests don't need a snapshot and why.

From what I can tell, ALL tests use a transformation step, which is the whole point of the snapshot.

matthewp:
The printer tests are the ones that should be a snapshot. But from this PR it doesn't look like the old manual tests are removed. So I don't think I am following what is going on.

ematipico:
> The printer tests are the ones that should be a snapshot. But from this PR it doesn't look like the old manual tests are removed. So I don't think I am following what is going on.

1. I'll reduce the amount of snapshots to the printing 
2. Regarding the old tests, I explained here why I found them useful: https://github.com/withastro/compiler/pull/995#issuecomment-2034492048

matthewp:
Regarding (2), I think you're saying that you find the print infrastructure useful for creating new tests. But once you've created them, there's not value in keeping them around forever, is there? If the existing printer `want{}` tests are duplicates of the snapshot tests then I don't see what value we are getting from the snapshots at all. I would be fine with keeping the print test infrastructure around as a helper during development, but don't see the benefit of checking in 2 copies of a testing.

CONTRIBUTING.md
```diff
@@ -94,6 +90,11 @@ Instead, if there are some **obsolete snapshots**, you can `UPDATE_SNAPS=clean`:
 UPDATE_SNAPS=clean go test -v ./internal/...
 ```
 
+
+### Adding new tests
+
+Adding tests for the tokenizer, scanner, and printer can be found in `internal/token_test.go`, `internal/js_scanner_test.go`, and `internal/printer/printer_test.go`, respectively.
```
Princesseuh, line 96: I don't really get how this explain how to make a new snapshot test, perhaps an example would be great?

ematipico: The commit [`037482d` (#1011)](https://github.com/withastro/compiler/pull/1011/commits/037482d117c52db47a3e1c5882a2e564bfcf64d4) adds more explanation

Princesseuh (approved):
I didn't check all the snapshots individually or anything, but this looks fine to me!

Issue #1006, opened 2024-05-30
🐛 BUG: `data-astro-reload` should accept value


stefanprobst:
### What version of `@astrojs/compiler` are you using?

2.8.0

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### Describe the Bug

i want to be able to conditionally apply the `data-astro-reload` attribute, i.e. `<a data-astro-reload={shouldReload || undefined}>`, but i get a warning that "The data-astro-reload attribute does not accept a value".

```ts
import { parse } from '@astrojs/compiler';

const result = await parse('<a data-astro-reload={undefined} href="#">Click</a>')

console.log(result.diagnostics)
```

```
[
  {
    severity: 2,
    code: 2005,
    location: { file: '<stdin>', line: 1, column: 23, length: 9 },
    hint: '',
    text: 'The data-astro-reload attribute does not accept a value'
  }
]
```

the warning is surfaced via `eslint-plugin-astro`'s `valid-compile` rule.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/stackblitz-starters-c53khs?file=index.js

matthewp:
Closed by https://github.com/withastro/compiler/pull/1012

PR #1012, opened 2024-06-21
Allow data-astro-reload to take a value

matthewp:
## Changes

- Removes the diagnostic disallowing `data-astro-reload` from having a value.
- I don't know the context for why this was added, but it prevents people from making the property from being conditional, ala `data-astro-reload={false}`. As far as I know we don't care about the value on the front-end.
- https://github.com/withastro/compiler/issues/1006

## Testing

- Updated the existing tests around this

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/light-ligers-rhyme.md
+++ b/.changeset/light-ligers-rhyme.md
@@ -2,4 +2,4 @@
 "@astrojs/compiler": patch
 ---
 
-Allow data-astro-reload to take a value
+Allow `data-astro-reload` to take a value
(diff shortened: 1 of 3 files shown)


martrapp (approved):
The original trigger were support channel threads with `data-astro-reload="some-name"` and its use on elements where it didn't work. But I have to admit that we overshot the mark by completely disabling expressions. ;-)

.changeset/light-ligers-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+Allow data-astro-reload to take a value
```
ematipico, line 5:
```suggestion
Allow `data-astro-reload` to take a value
```

ematipico (approved):

PR #1014, opened 2024-07-04
Avoid stringifying `undefined` in scoped class attributes

### LINKED ISSUE #1005: Undefined class attribute is not filtered out with `scopedStyleStrategy: 'class'`
### Astro Info

```block
Astro                    v4.8.3
Node                     v18.18.0
Package Manager          npm
Adapter                  none
Integrations             none
```

n/a / All

### Describe the Bug

If a component sets the `class` attribute to `undefined`, this is stringified and included in build output when using `scopedStyleStrategy: 'class'`, unlike behaviour for other attributes where setting them to `undefinded` causes the attribute to be omitted in build output.

For example, a component that does this:

```astro
<div class={undefined}>
```

Will render something like the following:

```html
<div class="undefined astro-HASH">
```

The build output should not include `undefined` in the `class` attribute.

https://stackblitz.com/edit/github-sgqbuf?file=src%2FTest.astro

### Participation

delucis:
## Changes

- Closes #1005
- Updates scoped class expression attributes to use nullish coalescing and avoid stringifying `undefined` as recommended by @bluwy.

## Testing

- Updates tests for the new expected syntax

## Docs

n/a — bug fix

### DIFF
--- a/.changeset/red-crabs-cough.md
+++ b/.changeset/red-crabs-cough.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/compiler': patch
+---
+
+Avoids stringifying `undefined` in scoped class attributes
--- a/internal/transform/scope-html.go
+++ b/internal/transform/scope-html.go
@@ -147,7 +147,7 @@ func injectScopedClass(n *astro.Node, opts TransformOptions) {
 					return
 				case astro.ExpressionAttribute:
 					// as an expression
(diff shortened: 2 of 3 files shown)


dreyfus92 (approved):
Awesome job @delucis, LGTM ✅

ascorbic (approved):

matthewp:
I had no idea the code worked like that :D

PR #1017, opened 2024-07-08
Fix TSX output for components with newlines in the starting tag

Princesseuh:
## Changes

In this situation:
```ts
<Image
/>
```
(note the line return)

The compiler would previously generate the following TSX (where [] represent the parts that are sourcemapped)

```
[<Image] [/>]
```

This is bad because in the editor, it means that you cannot get completion on that newline, because TypeScript thinks you're still writing the opening of the tag / has no idea where you are. This PR fixes this by adding a mapped trailing space, much like it does when there's a space.

Additionally, this PR fixes an edge case with the escaping code for attribute identifier where it would unnecessarily escape valid identifiers.

Fixes https://github.com/withastro/language-tools/issues/683

## Testing

Added / updated tests for both fixes. Also tested manually

## Docs

N/A


internal/printer/print-to-tsx.go
```diff
@@ -67,8 +67,9 @@ func isValidTSXAttribute(a Attribute) bool {
 		if i != 0 && !(isValidFirstRune(ch) ||
 			unicode.In(ch, unicode.Mn, unicode.Mc, unicode.Nd, unicode.Pc)) &&
 			// : is allowed inside TSX attributes, for namespaces purpose
+			// - is allowed inside TSX attributes, for custom attributes
 			// See https://facebook.github.io/jsx/#prod-JSXNamespacedName
-			ch != ':' {
+			ch != ':' && ch != '-' {
```
Princesseuh, line 72: I misread the JSX spec when implementing this, I didn't realise that it was saying that `-` was a valid character, I thought it was part of the visual design of the page 🤦‍♀️

matthewp (approved):

PR #1019, opened 2024-07-11
feat(tsx): Extract script and styles from TSX for language-server consumption

Princesseuh:
## Changes

Required by https://github.com/withastro/language-tools/pull/900

Previously, the language server would extract scripts and styles manually using both the compiler AST and the HTML AST. Not only is this kinda inefficient, it also lead to sometimes incorrect results due to wrong AST positions, HTML not understanding expressions, etc. 

This PR makes it so the compiler instead provide scripts and styles in the TSXResult. Additionally, this PR adds the ability to remove the scripts and styles from the output, for multiple reasons:

- The script tags produce invalid code, as no processing (apart from wrapping in an arrow function) was done on the content
- The style tags were simply text, they provide no actual value to the TSX. Actually, they end up often causing issues as in the language server, TSX seeing text there would override the symbol / selection range information of the actual content

These two changes were put under flags, so there's no breaking changes (certain consumer might still need this, for instance, our TS plugin does want the script tags content. Albeit, fixed, but that's for another PR, ha)

## Testing

Updated tests

## Docs

N/A


internal/printer/printer.go
```diff
@@ -36,6 +36,9 @@ type printer struct {
 
 	// Optional, used only for TSX output
 	ranges TSXRanges
+	// Keep track of how many multi-byte characters we've printed so that they can be skipped whenever we need a character-based index
+	// This could be directly in the token / node information, however this would require a fairly large refactor
+	bytesToSkip int
```
Princesseuh, line 41: So, this not really optimal, because it confuse the printed bytes with the source bytes. It's fine in 99.999% of cases, because we're not printing any multi-bytes ourselves that were not part of the source. But it's something we *could* do and would have to take into account then.

The real fix, but it'd require a large amount of work would be to have this information directly on the tokens / nodes. I first tried to do that, but I quickly realised that I was changing *everything* and ran into the risk of breaking things everywhere. 

If this prove to be an issue, I'd definitely be willing to make the proper change, but in my testing this honestly proved to work just fine.

matthewp (approved):

PR #1020, opened 2024-07-11
chore: format everything

Princesseuh:
## Changes

In every PR since the CI linting was introduced, making a change to the JS package result in a CI failure because of formatting change. Instead of every PR failing individually and fixing files individually, I figured I'd just run it all in one Go (get it)

## Testing

N/A

## Docs

N/A


ematipico (approved):

ematipico:
Maybe we can add this commit to the list of ignored commits

Princesseuh:
> Maybe we can add this commit to the list of ignored commits

This was done already!

Issue #862, opened 2023-09-08
🐛 BUG: [TSX] Unclosed tag result in unnecessary escaping


Princesseuh:
### What version of `@astrojs/compiler` are you using?

2.1.0

### What package manager are you using?

pnpm

### What operating system are you using?

macOS

### Describe the Bug

The following code:

```ts
<div></div>

<div class="something"
```

results in the following TSX:

```ts
<Fragment>
<div></div>{`

<div class="fdsfs"
`}
</Fragment>
```

Not quite sure why it does that, but it breaks type checking and completions (which can work even on a malformed tag). Closing the tag `<div class="fdsfs">` will fix it, but since this is editor tooling, you shouldn't have to.

### Link to Minimal Reproducible Example

See description

Princesseuh:
Hello, the Go compiler for Astro is no longer maintained. All future development will be focused on [our new Rust compiler](https://github.com/withastro/compiler-rs). If this issue is still relevant and you can reproduce it on the Rust compiler, please open a new issue there. Thank you!

PR #1023, opened 2024-07-12
fix(tsx): rework how escaping work

Princesseuh:
## Changes

Change how escaping works in TSX. Previously, if a TextNode had any of the following characters: `<>{}'"`, we'd escape the entire thing. This is fine, however, it doesn't work for TSX that is being written, as many times you'll end up with nodes that look like this:

```
<div></div>
<div
```

Where `<div` not being a completed tag, means that it get escaped (becomes ```<div></div>{`
<div`}```) and thus considered a string in the editor.

This PR changes two things:

- Leave this style of full escaping to style tags, JSON script tags and `is:raw`
- Only escape the individual offending characters for JSX (`>` and `}`), ex: ```<div>></div>``` becomes ```<div>{`>`}</div>```

The result isn't perfect, as the compiler still fundamentally does not understand unclosed tags, however it at least doesn't escape it in the vast majority of cases that affect editor tooling.

Fixes https://github.com/withastro/compiler/issues/862

## Testing

Added tests

## Docs

N/A


packages/compiler/test/tsx/escape.ts
```diff
@@ -43,4 +43,41 @@ export default function __AstroComponent_(_props: Record<string, any>): any {}\n
...
+});
+
+test('does not escape tag opening unnecessarily II', async () => {
+  const input = `<div>
+<div
+</div>
+`;
+  const output = `${TSXPrefix}<Fragment>
+<div>
+<div div {...{"<":true}}>
```
Princesseuh, line 65: This is what I mean by it not being perfect. Ideally, it would just be outputted as written, but the compiler thinks that those are attributes. This is very wonky looking, but as far as I could test, in the editor it's not a problem, TSX just think there's some text further on and an expression.

It might be possible to fix this, because the JS output actually does support this strangely enough. But it's perhaps a bit more complicated in TSX due to all the escaping required, not sure.

matthewp (approved):

PR #1028, opened 2024-07-17
fix(tsx): Transform top level returns into throws

Princesseuh:
## Changes

This PR makes it so top level returns are transformed into throws in the generated TSX

Fixes https://github.com/withastro/language-tools/issues/476

## Testing

Added a test

## Docs

N/A


bluwy (approved):

matthewp (approved):
finally 🙌

ematipico (approved):

PR #1032, opened 2024-07-29
Adds detected language to extracted style tags in TSX

Princesseuh:
## Changes

A long time ago, our language tooling would provide helpful intellisense inside SCSS and LESS tags, this was lost without notice. A recent regression made me remember this, so this PR brings it back

Additionally, this PR changes the script tag extraction a bit to handle `is:raw` and other edge cases better.

## Testing

Added tests

## Docs

N/A

### DIFF
--- a/internal/printer/print-to-tsx.go
+++ b/internal/printer/print-to-tsx.go
@@ -137,21 +137,57 @@ var htmlEvents = map[string]bool{
 	"onwheel":                   true,
 }
 
-func getScriptTypeForNode(n Node) string {
-	if n.Attr == nil || len(n.Attr) == 0 {
+func getStyleLangFromAttrs(attrs []astro.Attribute) string {
+	if len(attrs) == 0 {
+		return "css"
+	}
+
+	for _, attr := range attrs {
+		if attr.Key == "lang" {
+			if attr.Type == astro.QuotedAttribute {
+				return strings.TrimSpace(strings.ToLower(attr.Val))
+			} else {
+				// If the lang attribute exists, but is not quoted, we can't tell what's inside of it
+				// So we'll just return "unknown" and let the downstream client decide what to do with it
+				return "unknown"
+			}
+		}
+	}
+
+	return "css"
+}
+
+func getScriptTypeFromAttrs(attrs []astro.Attribute) string {
+	if len(attrs) == 0 {
 		return "processed-module"
 	}
 
-	// If the script tag has `type="module"`, it's not processed, but it's still a module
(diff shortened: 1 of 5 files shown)


internal/printer/print-to-tsx.go
```diff
@@ -137,21 +137,57 @@ var htmlEvents = map[string]bool{
...
+	}
+
+	for _, attr := range attrs {
+		if attr.Key == "lang" {
+			if attr.Type == astro.QuotedAttribute {
+				return strings.TrimSpace(strings.ToLower(attr.Val))
+			} else {
+				// If the lang attribute exists, but is not quoted, we can't tell what's inside of it
+				// So we'll just return "unknown" and let the downstream client decide what to do with it
+				return "unknown"
```
bluwy, line 152: Does this mean if the user writes `<style lang=scss>`, we return `"unknown"`? Shouldn't it be possible to parse the value still per HTML behaviour?

packages/compiler/test/tsx/meta.ts
```diff
@@ -52,6 +52,7 @@ test('extract scripts', async () => {
 				},
 				type: 'module',
 				content: 'console.log({ test: `literal` })',
+				lang: '',
```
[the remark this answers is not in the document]
Princesseuh: Go make this super annoying, it serializes nil strings into an empty string instead of undefined/omitting the property (and there's no omitempty, unlike JSON). 

We'd need two separate structs, but then since there's no inheritance, we'd need to duplicate the definition.

In truth, I kinda wanted to use the same format for both so that if there's ever a future where the language of a script tag is decoupled from its hoisting we already have the shape for it and it wouldn't be breaking (ex: `is:inline` + `lang="ts"` would result in type inline, lang ts)

internal/printer/print-to-tsx.go
Princesseuh: Ah yeah, you're right. I always forget about unquoted values, will fix!

Princesseuh: It turns out that for our compiler, unquoted values are included in the `QuotedAttribute` type, so no code change needed here!

packages/compiler/test/tsx/meta.ts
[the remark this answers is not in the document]
bluwy: Thanks that's good to know. Not a big deal then as long as our public API hides it. Maybe we could further generalize it with a `props: { ... }` property too (if Go allows that).

bluwy (approved):

PR #1037, opened 2024-08-05
fix(tsx): Extract positions based on utf-16 encoding

Princesseuh:
## Changes

Previously when extracting script and style tags, we tried to somewhat make it work with multibytes characters by counting them and skipping, not only was this cumbersome, it kinda didn't work because our loop would run multiple times over kinda the same characters, anyway it was annoying.

In this PR, I changed it so that we use the lineoffsets table from the sourcemapping logic to get the offsets. This is somewhat slower, especially in some extreme cases, but in most cases there's no difference, and at least it's now correct.

I also updated the frontmatter and body ranges extraction to use this method, as they suffered from the same problem

In theory, this is a breaking change, but the truth is that the numbers it'd spit out would be unusable in JS unless you did a lot of conversion yourself, now the numbers can be used as-is. Also, I doubt anyone other than me is using them...

Fixes https://github.com/withastro/language-tools/issues/921

## Testing

Tests should pass + updated some + added more

## Docs

N/A. Though I did a JSdoc comment on the type to say that it's UTF-16 based.

### DIFF
--- a/internal/printer/print-to-tsx_test.go
+++ b/internal/printer/print-to-tsx_test.go
@@ -9,126 +9,17 @@ import (
 	"github.com/withastro/compiler/internal/transform"
 )
 
+// One of the more performance-sensitive parts of the compiler is the handling of multibytes characters, this benchmark is an extreme case of that.
 func BenchmarkPrintToTSX(b *testing.B) {
-	source := `---
-import MobileMenu from "$components/layout/MobileMenu.astro";
-import MobileMenuSide from "$components/layout/MobileMenuSide.astro";
-import Header from "$components/layout/Header.astro";
-import Socials from "$components/layout/Socials.astro";
-import { getBaseSiteURL } from "$utils";
-import { Head } from "astro-capo";
-import "src/assets/style/prin.css";
-import type { MenuItem } from "../data/sidebarMenu";
-import Spritesheet from "$components/Spritesheet.astro";
-import GoBackUp from "$components/layout/GoBackUp.astro";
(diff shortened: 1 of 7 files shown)


bluwy (approved):
It does feel a bit non-performant, but if there isn't any noticable slowdown in benchmarks, then it's probably fine then.

Also, this needs a changeset before merging.

Princesseuh:
Yeah, I was surprised to not find too much perf difference. To be clear there is one, but I could only get real differences in files with 10k+ characters, multiple script and style tags late into the file and obscene amount of emojis.

I'll take a quick look to see if there's a way to re-use the line offset table from the sourcemapping logic, it'd speed up things a bunch, otherwise I'm not too bothered, this still ends up being faster for the language server because the previous logic it did to get script and style tags was expensive

bluwy:
The power of Go and native binaries 😄

Princesseuh:
Refactored to use the sourcemapping line offsets instead, it's much faster! Sourcemapping is still the ultimate bottleneck, though.

internal/printer/print-to-tsx_test.go
```diff
@@ -0,0 +1,38 @@
...
...n 🏬👸🍉👑 🎠🌇👰💄 🌕🗼🏮💇🐗 📌🏤🌲👹📌 🎐🕓📎🏤💾 tellus 💆🍨👂🕟💚🎋🌠🐳 🍘🍫🎵📚📟🔛🐑🏭🔄👌📏👚🏄🐞 💏🍊👾🍘👰📄💯🔑 proin 🐡🔝🐳🎻🐶🍜 🍕🌵📯🔖💅🕔💘🏁🌾💨🔲🔼🍜🍘🕂 in suscipit 🍎🎄🎬🔙👪💣🍣💯🕧 lectus 🌾🍚🍺🐆💉🎡👷 📖🔅👼🕞 🌄🌃📦🔲💘🌶💁🍯💿 senectus 👻💈🗽🍈 📜🍉🍒🍐🔖👙🔀🐅👙 🐬📷🎨👹🎬 🐷🌐🔽💨🎿🌌🌒 🎎🍊🔕🍁. 🍈🌄📧🏊📛👗🕟 🍋💌🔁🐛💫🔰👃🕑 ridiculus mattis 🕓🍌👘🌹 👅💏🍨🔯🍂🕃🌝👠🏁🔔 pellentesque elit eu, 💝🍧💯🔄📈📛🐻📆🔱📴🔸🍻🐘🎀 viverra 🌆🍉🍉🌆 📪🍕🏊🌜📺🔆🐢🎃 mi sed tellus luctus 🍜🍃🔶🗽 laoreet dui tristique 🔆🌸🐛🔣🏤📘🕜🍃🐋 pretium ultrices 🔳🏇🍏🎭🍀 🍱🐠👬📮🌗🌆💺🔆👨🌱 et 👸📁📩🔌👯👏👫👳💸 🐎💉🔱💦👴 🏪🍶🐸🔤 💔🏤👺👻🏤🎥🐽🔦👌🔡📚💡🔁🎻 🎵👡🕀🎩 🍸🐒🍭🕠🔢🎧🍚💪 🌸🍰💏🏁🎥🐕🔬💶 💃🌽💕🐭 💯🍁📭🐚📛📓🌏🔯📦 🕝👾📒📳💵 🏠💺🕔🕁📃 adipiscing nulla congue 🔖💲🕚🎰🔋👬 sem 👽👵👜👇 vehicula 📑🌎🍁💈 consectetur nulla ullamcorper enim, 🏫🍃🐢🔄🍝🕢👖💨👺💰👎💳🏠🏤 vel fermentum porttitor lacus 📱💧🔜🔰👱🎰🐉🔓🕧🍌🕙 🎴🍫🐱🐟🌖🕞🏥🍝📜🔃🏈🔡🐁🔗 💊🍐📝👭🎢🐳🐯📷🐀🔃. 🐦🍚🔵🕜🏩 id 🌔🔤💿🌻🍹 🌎🎈📢🔬🐖💢👸🎭 🍤🔵🔓🕕📪💢🔁👽💴 🐛👶🔢🎹🏄👜📌🍭🔼🎫🍯🎦💎🐦 🐨💧🍴🌊🔁 consequat pretium 👍🏬🏰👖🍥📺👛🌕 🏄🔱🔩🐏🌞🌺🐌👑 🏰🕢💱👖🐶🐥📯🎶💧🍭🔀🎾🏩🍑 massa, est 👶🍭💅🕔🍳💗🍞💚🍩🐷🍘🌄🐇 🐼👀👀🎅 📄🎁🐞💼 placerat erat 💧🌉💻🔣🐥📠🍪📻🐃🔻👠🕘🍞🍈🍔📴 dolor 📁💹🎐🍂🍉 neque, 👗🍁🍪🌵🕛🍄 🕝👔🔼🎡🔺📬 sed 💨💧🏢🐼👘🌳🎼👹🐉🕕 enim 🔏🏁👊💀🔓 🏫🐈💀🌳🔒👵🔘🎺💴🍑 👭🔲🍰💿🔪🌊 🌅🕔🎑📛🔶🔘🎑🍯 🐲📺👬📊🍒 elit, 🌽🎱💇🎥 ultricies 📡💲🐦🌁🍚🌵🍵 🎮📧🔟🍴👕🔏🌴🔊👳🗼💒🌴👍📞🔳👜🔥🐝🐾🐧🍊 🌰🗾🌂🐄 🔛🐀🍏🍞🔔📖💉📼🐥🍱 🐜🎺🍵💿👂🌋🌸 🔞🐤🔟🍀📙🏩👑 porta id 🎄💺🍙👶👪🔪👪🐭💚📗🎅🍐👡🎭 🐦🎌🐆💫 quis nulla dictumst non 📫🍵🔫🕠 🔯🔃🐷📖💙👓💍 🎬🕃🐥🏫🍛🌆🐗👅📷 📚🍭🌞🌎🕐👜👳 🏧💬💴🎆🐄🔠📄🐰📝🌈 🎩🌇🌟📙 suspendisse 🔡👫🍐🏩🌉🕚📘🐮 👐🏁👮🎭🏣 👰📤🍙🏈👓🐰🍦 lacinia diam eu vestibulum donec faucibus 🔝🎴🌷💩🍡🍜💙 🎐🔉🌛📠🏢📄🍆🎆.
```
Princesseuh, line 14: I tried this file in my editor, and well, performance was disastrous, but it's always been, even before this PR, so, it's not too bad.

internal/sourcemap/sourcemap.go
```diff
@@ -639,6 +648,22 @@ func (b *ChunkBuilder) GetLineAndColumnForLocation(location loc.Loc) []int {
...
+	lineAndColumn := b.GetLineAndColumnForLocation(location)
+	line := lineAndColumn[0] - 1
+	column := lineAndColumn[1] - 1
+
+	// Collect the length of every line before this one
+	offset := 0
+	for i := 0; i < line; i++ {
+		currentLine := b.lineOffsetTables[i]
+		offset += currentLine.utf16LineLength
+	}
```
bluwy, line 661: Perhaps if we keep `utf16LineLength` as the cumulative line length from the previous lines would help perf here a bit? That way we only need to access the the utf16 length once instead of calculating it everytime `OffsetAt` is called.

bluwy (approved):
Still looks good to me 👍

internal/sourcemap/sourcemap.go
Princesseuh: `utf16LineLength` is stored in the line offset table, it doesn't get calculated when `OffsetAt` is called. It ends up being a pretty simple int + int operation

PR #1055, opened 2024-12-21
Reverse order of styles & scripts

TheOtterlord:
## Changes

- Fixes https://github.com/withastro/astro/issues/12264 (See https://github.com/withastro/astro/issues/12264#issuecomment-2544037010 for more info)

For example, with component like the follow:

```html
<style>
  body {
    background-color: red;
  }
</style>
<style>
  body {
    background-color: yellow;
  }
</style>
```

The current compiler output would render `red` last, even though `yellow` was the last style-tag defined.

```html
<style>
  body {
    background-color: yellow;
  }
  body {
    background-color: red;
  }
</style>
```

Enabling the `experimentalScriptOrder` flag will respect the order style & script tags are defined.

```html
<style>
  body {
    background-color: red;
  }
  body {
    background-color: yellow;
  }
</style>
```

## Testing

- Updated existing styles test
- Added test for scripts order

## Docs

- Will require a section for the new experimental flag once implemented in withastro/astro.

### DIFF
--- a/.changeset/olive-beds-happen.md
+++ b/.changeset/olive-beds-happen.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+Fix the order that styles & scripts are rendered
--- a/.changeset/olive-beds-happen.md
+++ b/.changeset/olive-beds-happen.md
@@ -2,4 +2,4 @@
 "@astrojs/compiler": patch
 ---
 
-Fix the order that styles & scripts are rendered
+Add an experimental flag `experimentalScriptOrder` that corrects the order styles & scripts are rendered within a component. When enabled, the order styles & scripts are rendered will be consistent with the order they are defined.
--- a/.changeset/olive-beds-happen.md
+++ b/.changeset/olive-beds-happen.md
@@ -1,5 +1,5 @@
 ---
-"@astrojs/compiler": patch
+"@astrojs/compiler": minor
 ---
 
 Add an experimental flag `experimentalScriptOrder` that corrects the order styles & scripts are rendered within a component. When enabled, the order styles & scripts are rendered will be consistent with the order they are defined.
--- a/internal/transform/transform.go
+++ b/internal/transform/transform.go
@@ -125,7 +125,7 @@ func ExtractStyles(doc *astro.Node) {
 			if !IsHoistable(n) {
 				return
 			}
-			// prepend node to maintain authored order
+			// append node to maintain authored order
 			doc.Styles = append(doc.Styles, n)
 		}
 	})
@@ -434,7 +434,7 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *ha
 				}
 			}
 
-			// prepend node to maintain authored order
+			// append node to maintain authored order
 			if shouldAdd {
 				doc.Scripts = append(doc.Scripts, n)
 				n.HandledScript = true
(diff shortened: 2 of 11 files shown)


ematipico (approved):

MoustaphaDev (commented):
Changes look good to me, I just have a suggestion on a test

packages/compiler/test/scripts/order.ts
```diff
@@ -0,0 +1,17 @@
+import { transform } from '@astrojs/compiler';
+import { test } from 'uvu';
+import * as assert from 'uvu/assert';
+
+test('outputs scripts in expected order', async () => {
+	const result = await transform(`
+    <script type="module">console.log(1)</script>
+    <script type="module">console.log(2)</script>`);
```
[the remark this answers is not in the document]
TheOtterlord: Thanks! Updated

MoustaphaDev (approved):

MoustaphaDev:
Not sure why the go lint job is failing, but for you the js one running `pnpm check:write`should be able to fix it.

TheOtterlord:
It's also failing on main, so probably worth a separate PR

TheOtterlord:
ohh, only lint fails on main. I'll fix the js one now

MoustaphaDev (approved):

internal/transform/transform.go
```diff
@@ -126,7 +126,7 @@ func ExtractStyles(doc *astro.Node) {
 				return
 			}
 			// prepend node to maintain authored order
```
bluwy, line 128: Maybe we should update the comment too
```suggestion
			// append node to maintain authored order
```

internal/transform/transform.go
```diff
@@ -436,7 +436,7 @@ func ExtractScript(doc *astro.Node, n *astro.Node, opts *TransformOptions, h *ha
 
 			// prepend node to maintain authored order
```
bluwy, line 437:
```suggestion
			// append node to maintain authored order
```

bluwy (approved):

MoustaphaDev:
Anything needed to get this merged?

TheOtterlord:
Nope. I think this can merge!

MoustaphaDev:
Just realized applying this as a patch might break users' sites that rely on the current behavior. Should we maybe treat this as a major change or wait for the next major version? 🤔

ematipico:
I would also argue that the changeset is very cryptic and doesn't explain the fix well. What was the bug? How are styles and scripts ordered after the fix?

denisavitski:
is there a chance that this will be merged soon? the issue with styles arises not only when there are two `<style>` tags in a component but also when you use the component in other components, and these components, in turn, are used on different pages. it seems to me that this problem has been around for a very long time, and I constantly have to check after the build to ensure everything is displayed as intended and adjust the specificity of the selectors.

.changeset/olive-beds-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/compiler": patch
```
MoustaphaDev, line 2: Should it be a minor now that an experimental flag was introduced?
```suggestion
"@astrojs/compiler": minor
```

packages/compiler/test/scripts/order.ts
```diff
@@ -0,0 +1,20 @@
+import { transform } from '@astrojs/compiler';
+import { test } from 'uvu';
+import * as assert from 'uvu/assert';
+
+test('outputs scripts in expected order', async () => {
+	const result = await transform(`
+    <script>console.log(1)</script>
+    <script>console.log(2)</script>`);
```
MoustaphaDev, line 8: The flag is missing here

MoustaphaDev (approved):

ematipico (approved):

Issue #1052, opened 2024-12-02
Conditional rendering doesn't work directly on scripts


mayank99:
### Astro Info

```block
Astro                    v5.0.0-beta.12
Node                     v18.20.3
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

When conditionally rendering a `<script>`, Astro does not process it (i.e. typescript and npm imports do not work). I believe this is supposed to work after withastro/astro#11791.

```jsx
{condition && <script src="../main.ts"></script>}
```

As a workaround, I moved the `<script>` into a separate Astro component, and conditionally rendered that component, after which it started working as intended (i.e. Astro processes it).

```jsx
{condition && <ComponentContainingScript />}
```

### What's the expected result?

The behavior should be the same whether the `<script>` is rendered conditionally or whether a component containing the `<script>` is rendered conditionally.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-qpjs5e?file=src%2FMain.astro,src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #1057, opened 2025-01-07
fix: allow conditional rendering when `RenderScript` is enabled

MoustaphaDev:
## Changes

- Closes https://github.com/withastro/compiler/issues/1052

Conditionally rendered processed scripts were inlined like normal scripts. This made sense before `renderScript` was introduced as processed scripts (and styles) couldn't be conditionally rendered. Now that processed scripts can be conditionally rendered, we shouldn't inline them like normal scripts.

This PR adds the ability to conditionally render processed scripts

## Testing

- Added a printer test to ensure conditional rendering of scripts is working as expected whether `RenderScript` is enabled or not

## Docs

- https://github.com/withastro/docs/pull/10656


ematipico (approved):

bluwy (approved):
Not sure of the lint CI fail, but the fix here looks great to me. Thanks!

About docs, I think we need to update this part: https://docs.astro.build/en/guides/client-side-scripts/#opting-out-of-processing

> Astro will not process your script tags in some situations. In particular, adding type="module" or any attribute other than src to a <script> tag will cause Astro to treat the tag as if it had an is:inline directive. ~~The same will be true when the script is written in a JSX expression.~~

MoustaphaDev:
Good call, I'll update docs. The lint action fail is happening in main so it's unrelated.

sarah11918:
No problem with the docs PR making this change to be accurate! 

BUT to check:

> Astro will not process your script tags in some situations. In particular, adding type="module" or any attribute other than src to a <script> tag will cause Astro to treat the tag as if it had an is:inline directive. ~~The same will be true when the script is written in a JSX expression.~~

- if we are changing stated behaviour that people relied on, does this mean that some people might have written their scripts with certain expectations that are no longer true, and their project now works differently? Should we maybe add this note about scripts written in JSX expressions changing in the [upgrade guide entry for renderScript](https://docs.astro.build/en/guides/upgrade-to/v5/#script-tags-are-rendered-directly-as-declared), given that tons of people haven't upgraded yet and may still be using this guide for a while? That would handle the breaking change at the time it was intended (upgrade to v5 and directRenderScript becoming the default)

> This PR adds the ability to conditionally render processed scripts
- the docs changed don't seem to reference anything about a new option to allow conditional rendering? So, should this changeset should be more marked as a "fix that conditional rendering processed scripts wasn't working as intended" or there is some new feature for the docs to describe?

MoustaphaDev:
> if we are changing stated behavior that people relied on, does this mean that some people might have written their scripts with certain expectations that are no longer true, and their project now works differently?

That's a great question! It's possible some folks are probably still relying on the old script behavior, and their sites will behave differently after upgrading if we keep this as a patch release. To make the transition smoother we could go with a minor release instead, then for the next Astro minor, we can make the "breaking" change to the script behavior, and add a note to the changelog so people are aware of the change.

> Should we maybe add this note about scripts written in JSX expressions changing in the upgrade guide entry for renderScript, given that tons of people haven't upgraded yet and may still be using this guide for a while? That would handle the breaking change at the time it was intended (upgrade to v5 and directRenderScript becoming the default)

I think adding a note to the upgrade guide is a good idea! Was actually looking for a place to add it, so this is perfect.

> the docs changed don't seem to reference anything about a new option to allow conditional rendering? So, should this changeset should be more marked as a "fix that conditional rendering processed scripts wasn't working as intended" because there is not really some new feature for the docs to describe?

That's a good point! This is technically a fix as since v5, it was intended that processed scripts could be rendered conditionally.
(If this was a new feature, it would be a breaking change and would require a major release if not under a flag, i.e Astro 6 😅)

I'll update the changeset to reflect that. Thanks for the feedback! 🙌

MoustaphaDev:
Agreed! I think it's worth fixing this bug now, and we can mention this in the minor version blog post if we push this as a minor release. That way, those upgrading to next minor can be aware of the change and update their scripts accordingly, and current users are shielded.

And yeah, we can always revert if it becomes too big of a problem. Will lookout in support to gauge it.

bluwy:
I'm not sure how we can only limit this to the Astro minor. We have `"@astrojs/compiler": "^2.10.3"` which would allow minor releases of the compiler. Even if we use `~` and publish that, users who've not upgraded Astro could've also only upgraded `@astrojs/compiler` still and hit the issue. If we want to limit this as an Astro minor, we likely have to expose a new option from the compiler which we can enable in the Astro minor. I personally am also fine with releasing this PR as is though and check how this impacts people.

MoustaphaDev:
Good catch! I wrongly assumed that we pinned the compiler with `~`. Agreed with merging this as is!

.changeset/grumpy-ads-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+Allows conditional rendering of scripts when `RenderScript` is enabled
```
sarah11918, line 5:
```suggestion
Fixes an issue with the conditional rendering of scripts. 

**This change updates a v5.0 breaking change when `experimental.directRenderScript` became the default script handling behavior.** If you have already successfully upgraded to Astro v5, you may need to review your script tags again and make sure they behave as desired. [See the v5 Upgrade Guide for more details.](https://docs.astro.build/en/guides/upgrade-to/v5/#script-tags-are-rendered-directly-as-declared)
```

So, just a suggestion, but something of this style maybe? (And, check the content at the upgrade guide link to see whether there's anything else here needed?)

MoustaphaDev: Yeah this is great!

sarah11918 (approved):
Approving for docs, and the Docs PR is good to merge when this is released, too!

MoustaphaDev:
Awesome, thanks for your reviews and suggestions!

sarah11918:
Just following up on this one, since the docs are also waiting! What still needs to happen here?

MoustaphaDev:
Merging!

PR #1062, opened 2025-02-02
Fix expression parsing in foreign context

Trombach:
## Changes

- closes https://github.com/withastro/compiler/issues/1063

This PR adds the current `Namespace` to the expression Node.

When writing an expression inside foreign content (i.e. SVG) the astro compiler did not persist the namespace and therefore parsed the contained nodes in normal HTML mode. This lead to `<image>` tags incorrectly being transformed to `` tags, but could also have caused other Problems.

## Testing

- added printer test to validate correct parsing of `{<image />}` inside svg

## Docs

bug fix only

### DIFF
--- a/.changeset/plenty-geese-laugh.md
+++ b/.changeset/plenty-geese-laugh.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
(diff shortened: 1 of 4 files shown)


MoustaphaDev (approved):
Looks good to me, thanks for the fix!

ematipico (approved):

PR #1067, opened 2025-03-04
fix: await issue with fragment

mertmit:
## Changes

- What does this change?
- Move adding async prefix to printer method
- Fixes https://github.com/withastro/astro/issues/13334

## Testing

Printing following payload:
```
<body><Fragment> { await Promise.resolve("Awaited") } </Fragment></body>
```

## Docs

bug fix

### DIFF
--- a/.changeset/small-lions-reflect.md
+++ b/.changeset/small-lions-reflect.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+Add async properly when await used inside fragment
--- a/internal/printer/__printer_js__/Component_with_await.snap
+++ b/internal/printer/__printer_js__/Component_with_await.snap
@@ -1,0 +1,41 @@
+
+[TestPrinter/Component_with_await - 1]
+## Input
+
+```
+<body><Component> { await Promise.resolve("Awaited") } </Component></body>
+```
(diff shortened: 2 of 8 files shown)


ematipico:
Thank you @mertmit 

There are a couple of things missing:
- a changeset that explains the entity of the fix; the contribution guide should explain how to add one 
- a test that matches the bug filed in the issue

mertmit:
Hi @ematipico 
I've added the changeset.
Test is already included and it covers exact same scenario (await expression inside fragment).

ematipico:
I know, but the code provided by the user is different from the test, hence my request

ascorbic:
Thanks. The bug affects all Astro component wrappers, not just Fragments. Will this work with others too? It would be good to have a test for shorthand Fragment syntax too.

mertmit:
@ascorbic it seems it should work for others as well (it should only effect formerly broken ones so I believe it is safe)
I just came across astro so I am not very familiar with concepts so please double confirm.
I also added a test scenario for Component just in case.

ematipico (approved):
Thank you

jlwalkerlg:
@ematipico any idea when this fix is likely to be released? Thanks! :)

PR #1070, opened 2025-04-24
fix: parse `select` correctly when it contains HTML element

### LINKED ISSUE #1069: Astro incorrectly closes `option` tag when it contains an HTML tag
### Astro Info

```block
Astro                    v5.7.4
Node                     v20.19.0
Package Manager          unknown
Adapter                  none
Integrations             none
```

Chrome

### Describe the Bug

Astro does not correctly close `option` tags when they contain a tag inside of them. When it just contains text, `option` is closed correctly. However if you to place a span inside for example:
```astro
<select>
  <option value="lemon"><span>Lemon</span></option>
  <option value="lime"><span>Lime</span></option>
</select>
```
Astro produces
```html
<select>
    <option value="lemon"><span>Lemon
    <option value="lime"><span>Lime
        </span></option></span></option>
</select>
```
So you end up with an `option` that wraps a `span` that wraps an `option` that wraps a `span` instead of the expected `option` wraps `span`, `option` wraps `span`.

Firefox parses what it thinks you probably meant and separates them into their own options. Chrome treats it as as a single option.

Astro should accurately reproduce the expected HTML. Traditionally rich HTML content would just be ignored by the browser so there was no need to include other tags, but as of Chrome 135 select can now be [customized](https://developer.chrome.com/blog/a-customizable-select) and can include [rich HTML content](https://developer.chrome.com/blog/a-customizable-select#a_select_can_now_include_rich_html_content) enabling more interactive components without relying on JS. I have not tested whether Chrome 134 behaves like 135 or like Firefox.

https://stackblitz.com/edit/github-ttndfuvp?file=src%2Fpages%2Findex.astro

### Participation

lbrooney:
Closes #1069. Select was the problem not options, but that issue was prior to me looking at the WHATWG documentation and the Astro parser.

## Changes

- This fixes parsing of `select` elements when they contain HTML elements.
- The issue is specifically with how End Tags were being handle when `inSelect` mode
  - Only `option`, `optgroup`, `select`, and `template` end tags were being handled.
  - Others were unhandled, this PR just changes so that if the end tag is not any of the ones from the previous bullet point, the parser should treat the tag as if it were `inBody`.
  - Start tag handles it this way already.
- Previously the compiler could correctly produce a `select` containing a single HTML element, which I'll refer to as *X*, or a `select` containing an `option` containing a single *X*.
- However if a HTML element, *Y*, is added as the next sibling to either the `select`, `option`, or *X*, *Y* will cannibalized by *X* becoming its child. Each subsequent element(child or sibling) will have whatever element was preceding it as its parent.

ie:
```astro
<select id="select-span-sibling">
  <span>Lemon</span>
  <option>Lime</option>
</select>
<div><span>Orange</span></div>
<div>Melon</div>
```
produces (formatted for readability):
```html
<select id="select-span-sibling">
  <span>
    Lemon
    <option>
      Lime
    </option>
    <div>
      <span>
        Orange
        <div>
          Melon
        </div>
      </span>
    </div>
  </span>
</select>
```
- This change fixes it so that the previous example now produces the expected output:
```html
<select id="select-span-sibling">
  <span>Lemon</span> 
  <option>Lime</option>
</select>
<div><span>Orange</span></div>
<div>Melon</div>
```
- I briefly tested the compiled WASM in a local Astro project and the generated HTML from the `.astro` was matching the expected/"correct" output.
- I don't know if there are additional changes needed to be made to handle special cases or better match the spec or if this somehow broke how people were previously using it, but I believe this PR produces a more correct output than before.

## Testing

I wasn't sure where best to do testing so I just added to `test_printer` where I could most immediately observe outputs. Additionally I wasn't sure what was the appropriate amount of tests I should add or how verbose to be with the naming, so I added only a few to cover some limited scenarios and it's far from exhaustive. I can add, remove, or change them if it is desired. 

## Docs

I did not change any documentation as I believe it's only a bugfix.

### DIFF
--- a/.changeset/large-mangos-look.md
+++ b/.changeset/large-mangos-look.md
@@ -2,4 +2,4 @@
 "@astrojs/compiler": patch
 ---
 
-fix: correctly parse select elements containing HTML, preventing elements from being incorrectly nested
+Fixes a bug where the compiler couldn't correctly parse `select` elements, preventing elements from being incorrectly nested
--- a/.changeset/large-mangos-look.md
+++ b/.changeset/large-mangos-look.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+fix: correctly parse select elements containing HTML, preventing elements from being incorrectly nested
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -2319,6 +2319,8 @@ func inSelectIM(p *parser) bool {
 			p.resetInsertionMode()
 		case a.Template:
 			return inHeadIM(p)
+		default:
+			return inBodyIM(p)
 		}
 	case CommentToken:
 		p.addChild(&Node{
--- a/internal/printer/__printer_js__/select_with_2_options_containing_element.snap
+++ b/internal/printer/__printer_js__/select_with_2_options_containing_element.snap
@@ -1,0 +1,41 @@
+
+[TestPrinter/select_with_2_options_containing_element - 1]
+## Input
+
+```
+<select><option><span>Lemon</span></option><option><span>Lime</span></option></select>
+```
+
+## Output
+
+```js
+import {
+  Fragment,
+  render as $$render,
+  createAstro as $$createAstro,
+  createComponent as $$createComponent,
+  renderComponent as $$renderComponent,
+  renderHead as $$renderHead,
+  maybeRenderHead as $$maybeRenderHead,
+  unescapeHTML as $$unescapeHTML,
+  renderSlot as $$renderSlot,
+  mergeSlots as $$mergeSlots,
+  addAttribute as $$addAttribute,
+  spreadAttributes as $$spreadAttributes,
+  defineStyleVars as $$defineStyleVars,
+  defineScriptVars as $$defineScriptVars,
+  renderTransition as $$renderTransition,
+  createTransitionScope as $$createTransitionScope,
+  renderScript as $$renderScript,
+  createMetadata as $$createMetadata
+} from "http://localhost:3000/";
+
+export const $$metadata = $$createMetadata(import.meta.url, { modules: [], hydratedComponents: [], clientOnlyComponents: [], hydrationDirectives: new Set([]), hoisted: [] });
+
+const $$Component = $$createComponent(($$result, $$props, $$slots) => {
+
+return $$render`${$$maybeRenderHead($$result)}<select><option><span>Lemon</span></option><option><span>Lime</span></option></select>`;
+}, undefined, undefined);
+export default $$Component;
+```
+---
--- a/internal/printer/__printer_js__/select_with_2_options_containing_element_with_div_sibling.snap
+++ b/internal/printer/__printer_js__/select_with_2_options_containing_element_with_div_sibling.snap
@@ -1,0 +1,41 @@
+
+[TestPrinter/select_with_2_options_containing_element_with_div_sibling - 1]
+## Input
+
+```
+<select><option><span>Lemon</span></option><option><span>Lime</span></option></select><div>Orange</div>
+```
+
+## Output
+
+```js
+import {
+  Fragment,
+  render as $$render,
+  createAstro as $$createAstro,
+  createComponent as $$createComponent,
+  renderComponent as $$renderComponent,
+  renderHead as $$renderHead,
+  maybeRenderHead as $$maybeRenderHead,
+  unescapeHTML as $$unescapeHTML,
+  renderSlot as $$renderSlot,
+  mergeSlots as $$mergeSlots,
+  addAttribute as $$addAttribute,
+  spreadAttributes as $$spreadAttributes,
+  defineStyleVars as $$defineStyleVars,
+  defineScriptVars as $$defineScriptVars,
+  renderTransition as $$renderTransition,
+  createTransitionScope as $$createTransitionScope,
+  renderScript as $$renderScript,
+  createMetadata as $$createMetadata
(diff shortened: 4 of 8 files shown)


.changeset/large-mangos-look.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+fix: correctly parse select elements containing HTML, preventing elements from being incorrectly nested
```
ematipico, line 5:
```suggestion
Fixes a bug where the compiler couldn't correctly parse `select` elements, preventing elements from being incorrectly nested
```

ematipico (approved):
Thank you!

ematipico:
@lbrooney are you sure about the validity of the bug? The w3c validator returns an error when `span` is a child of `select`

lbrooney:
> @lbrooney are you sure about the validity of the bug? The w3c validator returns an error when `span` is a child of `select`

So I believe you are correct in that a `span` is not a valid child of a `select`(if my understanding is correct the [current spec](https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inselect) says anything not explicity defined is an error).

To my knowledge, currently a `span` in an `option` is also not spec compliant either, but considering as of Chrome 135, with the introduction of customizable selects and the ability for them to contain [rich HTML content](https://developer.chrome.com/blog/a-customizable-select#a_select_can_now_include_rich_html_content) that is changing. There is an active spec pull request here: whatwg/html#9799. Additional details can also be found in the previously linked Chrome dev blog as well.

Admittedly my example in the initial PR is bad as I'm not sure that even within the new spec changes whether that would be valid, but as of right now Chrome doesn't raise any concerns about it, it exists within the DOM, but it not visible/presented(maybe with some CSS you can make it visible). Apologies, however, I do think it demonstrated how broken the output produced could be. 

I probably should have included another example in my opening issue, but for instance:
```astro
<select>
  <option><span>Lemon</span></option>
  <option><span>Lime</span></option>
</select>
```
currently produces(formatted):
```html
<select>
  <option>
    <span>Lemon
      <option>
        <span>Lime</span>
      </option>
    </span>
  </option>
</select>
```

Taking advantage of this new feature is currently impossible within an Astro component.

Blink currently has this behind a [runtime flag](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/forms/html_select_element.cc;drc=a6951b1a6d609ed109abcf25278eb1f74886223b;bpv=1;bpt=1;l=1899?gsn=SelectParserRelaxationEnabled&gs=KYTHE%3A%2F%2Fkythe%3A%2F%2Fchromium.googlesource.com%2Fcodesearch%2Fchromium%2Fsrc%2F%2Fmain%3Flang%3Dc%252B%252B%3Fpath%3Dthird_party%2Fblink%2Frenderer%2Fcore%2Fhtml%2Fforms%2Fhtml_select_element.cc%23ZoPo0Sk3LsrGv0LQDyW_u7dPDL-kNpvm2IodgS1UNV8) (the function above it does the actual check). Checking the references you can see how they've changed parsing, with `html_tree_builder.cc` I believe being most relevent in this case.

MoustaphaDev:
I posted a proposal to discuss that!

https://github.com/withastro/roadmap/discussions/1166

lbrooney:
> Thanks for contributing!

For why my solution is rather terse is that while, to my understanding, the new select spec is largely finalized, I didn't want to engineer a spec-for-spec implementation only for it to change and have it now be incorrect (long way to say I was lazy 😅). The current changes introduced by the PR solved my problems so I was satisfied with it. To your second point, if the produced HTML with this solution was not spec compliant, the browser would handle it, which while not ideal is well within their capabilities.

I can push some additional snapshot tests, but with the current PR if a `select` tag (properly closed/opened or not) is a child of an an element within a select, it is does not exist in the produced output.
```html
<!-- any of these inputs -->
<select><option><span><select>Lemon</select></span></option></select>
<select><option><span><select></select>Lemon</span></option></select>
<select><option><span><select>Lemon</span></option></select>
<select><option><span></select>Lemon</span></option></select>
<!-- same output by astro -->
<select><option><span>Lemon</span></option></select>`
```
The behaviour on handled cases within inSelect.
```html 
<!-- select tag in option any of these inputs -->
<select><option><select></select><span>Lemon</span></option></select>
<select><option><select><span>Lemon</span></option></select>
<select><option></select><span>Lemon</span></option></select>
<!-- same output by astro -->
<select><option></option></select><span>Lemon</span>
```

> I posted a proposal to discuss that!
> 
> [withastro/roadmap#1166](https://github.com/withastro/roadmap/discussions/1166)

In regards to your proposal, I was having similar thoughts when I initally came across this bug and found that I didn't encounter it with React TSX. I believe that's essentially how the current Astro TSX parser workers by just ignoring parse errors, and I may be incorrect in this, I believe React TSX also doesn't "correct" improper HTML.

FlorianRappl:
> I believe React TSX also doesn't "correct" improper HTML.

How / why should it? All the DOM knowledge is in `react-dom` and JSX is constructed via simple rules - which do not follow HTML(5) rules. The tree is constructed just by using methods such as `createElement` and `appendChild`, not via the HTML(5) parser (which would take a string to be parsed as a fragment, i.e., be much slower and not synced with the VDOM).

In short React / JSX does not deal at all with HTML - just with DOM nodes. The only time `innerHTML` is used is via the special `dangerouslySetInnerHTML` prop on "native" elements (i.e., handled at runtime via `react-dom`).

Princesseuh:

PR #1075, opened 2025-05-19
fix: export w/ curly brace on next line, extra curly brace before def

alex-snezhko:
## Changes
Closes https://github.com/withastro/astro/issues/13790, where a curly brace which is not the start of a function body is incorrectly classified as such.

This is just a patch for the case reported in this issue and a few other ones, such as:
- A curly brace appearing on the parameter type i.e. `export function getStaticPaths(input: { paginate: any })`
- A curly brace in the type definition of an arrow function i.e. `export const getStaticPaths: () => { params: any }[]`

There may be an additional case that I missed, please let me know if you can think of one!

### Note
Though I did not do this in this PR to minimize the surface area of the changes, in my opinion it would be worthwhile to pull in a full-fledged TS parser and operate on the AST for this export hoisting use case. Perhaps when the new TS Go compiler from Microsoft is stable the parser can be taken from there :shrug:

## Testing

Tests added to appropriate files

## Docs

Bug fix, no changes to public API

### DIFF
--- a/.changeset/fast-lemons-pay.md
+++ b/.changeset/fast-lemons-pay.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+Improves detection of function body opening curly brace for exported functions.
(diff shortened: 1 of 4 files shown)


ematipico (approved):

ematipico (commented):
Can you please add a changeset @alex-snezhko ?

PR #1098, opened 2025-08-22
Fix head content parsing issue

### LINKED ISSUE #957: Closing HTML tag getting removed when there's a component above
### Astro Info

```block
Astro                    v4.3.1
Node                     v18.19.0
Package Manager          unknown
Adapter                  none
Integrations             @astrojs/sitemap
```

_No response_

### Describe the Bug

Steps to reproduce:

1. Clone https://github.com/robertguss/Astro-Theme-Creek
2. Update Astro to v4.3.1
3. Start dev server (`npm run dev`)
4. Observe the issue: the `</title>` tag is missing

From what I can see, if you remove `<GoogleAnalytics />` inside `BaseHead.astro`, then it fixes the issue, which suggests this issue only crops up when a component is added. This issue still exists even if `GoogleAnalytics.astro` is empty.

The `</title>` tag should be present. Because it's missing, it's causing some side effects, such as the Google Fonts not loading. The H1 font should look like this:

https://astro-theme-creek.netlify.app/

https://stackblitz.com/edit/github-2bvud4-crhr4g

### Participation

HiDeoo:
## Changes

- Closes #957

This PR fixes an issue with head content parsing when such content is defined in a component that first renders another component before other elements like `<title>` when using an expression for the title.

When first encoutering a start tag token for a component before any HTML, we [switch to the "in body" insertion mode](https://github.com/withastro/compiler/blob/main/internal/parser.go#L692). After that, when we encounter a start tag token for some head related tags, we [use the "in head" insertion mode](https://github.com/withastro/compiler/blob/main/internal/parser.go#L1163-L1164) to parse them correctly. Altho, when we later encounter the associated end tag token, we do not switch back to the "in head" insertion mode, which causes invalid markup to be generated.

For example, considering the following Astro component supposed to be rendered in the head:

```astro
---
import Analytics from '../components/Analytics.astro';
---
<Analytics />
<title>{title}</title>
<meta name="description" content="a description" />
```

The generated code before this fix was invalid (notice the `meta` tag is inside the `title` tag):

```js
return $$render`${$$renderComponent($$result, "Analytics", Analytics, {})}
<title>${title}
<meta name="description" content="a description"></title>`;
```

## Testing

I added a new print test case for a similar scenario as the one described in the issue which generates a snapshot file with the expected output after the fix.

## Docs

This is a bug fix only, so no docs were added or updated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,9 @@ jobs:
         uses: golangci/golangci-lint-action@v4
         with:
           version: latest
+        env:
+          GOOS: js
+          GOARCH: wasm
 
   lint-js:
     timeout-minutes: 3
@@ -81,7 +84,5 @@ jobs:
       - uses: actions/checkout@v4
       - name: Setup Biome
         uses: biomejs/setup-biome@v2
-        with:
-          version: latest
       - name: Run linting
         run: biome ci --diagnostic-level=warn
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,10 +1,8 @@
-run:
-  skip-dirs:
-    - lib
-  skip-files:
-    - xxhash.go
-
 issues:
+  exclude-dirs:
+    - lib
+  exclude-files:
+    - xxhash.go
   exclude-rules:
     - path: token.go
       linters:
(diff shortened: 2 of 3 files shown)


ematipico (approved):
Great fix!

PR #1101, opened 2025-08-23
Fix: Support HTML `<selectedcontent>` element

### LINKED ISSUE #1093: Unable to render base-select in .astro pages
### Astro Info

```block
Astro                    v5.7.10
Node                     v22.14.0
System                   macOS (arm64)
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

When rendering a native customizable select widget, the HTML output on `.astro` pages will be messed up. See the reproduction for more details.

A fully functional select widget should be rendered. See the reproduction for more details.

https://stackblitz.com/edit/github-wt73r64g?file=src%2Fpages%2Ftest.html,src%2Fpages%2Findex.astro&on=stackblitz

### Participation

jp-knj:
Closes #1093

### Background

The `<selectedcontent>` element is used inside customizable `<select>` elements to display the currently selected option within a custom button. It's part of the new web standard for styleable select elements that's been years in the making.

```html
<select>
  <button>
    <selectedcontent></selectedcontent>
  </button>
  <option>Option 1</option>
  <option>Option 2</option>
</select>
```

### Problem

Without this fix, the Astro compiler treats <selectedcontent> as a regular container element, causing it to incorrectly consume all following elements until the parent's closing tag. This breaks the structure of customizable select elements.

Before: Options incorrectly nested inside selectedcontent

```
<selectedcontent>
  <option>Option 1</option>
  <option>Option 2</option>
</button>
```

After: Correct structure with selectedcontent as void element

```
<selectedcontent>
</button>
<option>Option 1</option>
<option>Option 2</option>
```

### Changes

- Handling for the `<selectedcontent>` element:
  - Only treat `selectedcontent` as self-closing when it has the `/>` syntax 
  - Allow normal start/end tag pairs for `<selectedcontent></selectedcontent>`
  - Handle `selectedcontent` as a void element
  - Ignore `</selectedcontent>` end tags, Add support for closing `</button>` tags within select context
  - Keep `selectedcontent` in the voidElements map   

### Testing

- Added tests for both `<selectedcontent></selectedcontent>` and `<selectedcontent />` syntax
- Updated test snapshots with correct expected output

### References

  - https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/selectedcontent
  - https://developer.chrome.com/blog/a-customizable-select
  - https://open-ui.org/components/customizableselect/

  ---

### DIFF
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -1418,6 +1418,14 @@ func inBodyIM(p *parser) bool {
 				return true
 			}
 		default:
+			// Special handling for selectedcontent as a void element
+			if p.tok.Data == "selectedcontent" {
+				p.reconstructActiveFormattingElements()
+				p.addElement()
+				p.oe.pop()
+				p.acknowledgeSelfClosingTag()
+				return true
+			}
 			p.reconstructActiveFormattingElements()
 			p.addElement()
 			if p.hasSelfClosingToken {
@@ -1433,6 +1441,12 @@ func inBodyIM(p *parser) bool {
 		if isComponent(p.tok.Data) {
 			p.addLoc()
 			p.oe.pop()
+			return true
+		}
+
+		// Special handling for selectedcontent end tag - just ignore it
+		// since it's treated as a void element
+		if p.tok.Data == "selectedcontent" {
 			return true
 		}
 
@@ -2319,6 +2333,19 @@ func inSelectIM(p *parser) bool {
 			p.resetInsertionMode()
 		case a.Template:
 			return inHeadIM(p)
+		default:
+			// Handle closing tags for elements that are allowed in customizable select
+			// (like button for the new HTML select element)
+			if p.tok.Data == "button" {
+				// Close the button if it's open
+				for i := len(p.oe) - 1; i >= 0; i-- {
+					if p.oe[i].Data == "button" {
+						p.oe = p.oe[:i]
+						break
+					}
+				}
+				return true
+			}
 		}
 	case CommentToken:
 		p.addChild(&Node{
--- a/.changeset/happy-ways-collect.md
+++ b/.changeset/happy-ways-collect.md
@@ -1,0 +1,8 @@
+---
+"@astrojs/compiler": minor
+---
+
+Support HTML <selectedcontent> element
+
+Based on the recent commit history, this change appears to be related to fixing issue #1093 regarding selectedcontent parsing in customizable selects. The <selectedcontent> element is part of the new Customizable Select Element API
+in HTML, used within <selectlist> elements to display the currently selected option(s).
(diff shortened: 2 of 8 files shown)


internal/parser.go
```diff
@@ -1436,6 +1444,12 @@ func inBodyIM(p *parser) bool {
 			return true
 		}
 
+		// Special handling for selectedcontent end tag - just ignore it
+		// since it's treated as a void element
+		if p.tok.Data == "selectedcontent" {
+			return true
+		}
+
```
jp-knj, line 1452: ### Why Ignore </selectedcontent> End Tags

The Problem Flow:
 When HTML contains `<selectedcontent></selectedcontent>`:

1. Tokenizer generates these tokens:
  - `<selectedcontent>` → StartTagToken
  - `</selectedcontent>` → EndTagToken
2. Parser processing:
  - Treats <selectedcontent> as a void element (can't have children)
  - Adds it to DOM, then immediately pops it from the stack
  - At this point, selectedcontent is no longer on the stack
3. When </selectedcontent> arrives:
  - There's no matching opening tag on the stack
  - If we don't ignore it: Parser gets confused and misplaces subsequent elements

  What Actually Happened:
```html
  <!-- Input -->
  <select><button><selectedcontent></selectedcontent></button><option>A</option></select>

  <!-- Wrong output (without ignoring the end tag) -->
  <select><button><selectedcontent><option>A</option></button></select>
  <!-- ↑ option becomes a child of selectedcontent (WRONG!) -->
  ```

File: internal/parser.go [hunk 2]
```diff
@@ -2319,6 +2333,19 @@ func inSelectIM(p *parser) bool {
...
+			if p.tok.Data == "button" {
+				// Close the button if it's open
+				for i := len(p.oe) - 1; i >= 0; i-- {
+					if p.oe[i].Data == "button" {
+						p.oe = p.oe[:i]
+						break
+					}
+				}
+				return true
+			}
```
jp-knj, line 2348:
```suggestion
		default:
			// Handle closing tags for elements that are allowed in customizable select
			// (like button for the new HTML select element)
			if p.tok.Data == "button" {
				// Close the button if it's open
				for i := len(p.oe) - 1; i >= 0; i-- {
					if p.oe[i].Data == "button" {
						p.oe = p.oe[:i]
						break
					}
				}
				return true
			}
```

jp-knj, line 2348: ### Why support `</button>` tags in select context

- In standard HTML, a `<button>` is not allowed inside a `<select>`.
- The Customizable Select feature, however,  allow interactive content (including `<button>`) inside `<select>`.
  Supporting `</button>` here is forward-compatible with that model.

#### before

```
EndTag </button>
Stack before: <html> <select> <button>
Stack after:  <html> <select> <button>   ← button NOT closed!

StartTag <option>
Stack after:  <html> <select> <button> <option>   ← option becomes child of button (WRONG!)
```

#### After

```
EndTag </button>
Stack before: <html> <select> <button>
Stack after:  <html> <select>              ← button correctly closed

StartTag <option>
Stack after:  <html> <select> <option>     ← option is direct child of select (CORRECT!)
```

PR #1111, opened 2025-09-25
feat: Astro v6

florian-lefebvre:
See https://github.com/withastro/astro/pull/14383 for more details


florian-lefebvre:
I have admin privileges, let me try (we should probably migrate to a similar workflow than the monorepo)

delucis:
Nice try 😄 
image

PR #1114, opened 2025-10-02
feat: remove renderScript option

florian-lefebvre:
## Changes

- It became the default in v5 so it makes no sense to keep it now
- Lot of trial and error, maybe what I made is total nonsense. REVIEW CAREFULLY

## Testing

Should pass

## Docs

Changeset


ematipico (approved):
Looks good! You should prune the old snapshots. Since you changed the title of the current snapshots, new ones were created, but the old ones are still there

PR #1115, opened 2025-10-02
feat: remove experimentalScriptOrder option

florian-lefebvre:
## Changes

- It is now the default

## Testing

Should pass

## Docs

Changeset


louisescher (approved):
Happy to see this land!

PR #1127, opened 2025-11-18
Pin action versions

delucis:
## Changes

- Pins the versions of actions in use in CI workflows

## Testing

Existing CI should run

## Docs

n/a


HiDeoo (approved):
Great idea 👏 👍

PR #1130, opened 2025-12-23
Fix: semantic props order override

samuelhulla:
## Changes

Adjusts astro compiler to adhere to JSX semantic prop overrides.

```astro
<!-- user html pre-compilation  -->
<div a="1" a="2" /> 

<!-- Currently results in ❌ -->
<div a="1" />

<!-- After this PR ✅ -->
<div a="2" />
```

Manually overriding like that is a rare occurance, but consider this far more common issue

```astro
---
const props = { class: "foo" } // obv. normally we'd use Astro.props but for keeping example concise
---
<!-- user html pre-compilation -->
<div {...props} class="bar" >

<!-- What you end up getting ❌ -->
<div class="foo" />

<!-- What most users expect they will get -- and will get after this PR ✅ -->
<div class="bar" />
```

## Testing

Added both ts and go tests for testing the compiler (parser/printer)  for deduping various example and cornercases like svg namespaces, multiple spreads, etc.

Tested via standard test workflow: 

```sh
pnpm build:all
UPDATE_SNAPS=true go test -v ./internal
pnpm test
```

## Docs

Doesn't require any direct documentation changes imho, but does require a breaking change notice in the changelog / announcement post of the version *(even though majority of users should be unaffected by this change)*

## Closing remarks

Full disclosure, I have very limited `golang` experience. I did my best to address the necessary changes in a meaningful and performant way, but I wanted to make it explicitly clear -- My expertise is TypeScript with a hobbylang of Rust, so just so we're on the same page -- I did use AI to help me get this PR over the finish line to make up for my lack of go knowledge.

I did my best to verify everything and write a comprehensive test suite for it, but  it is more than likely that there may be something suboptimal given the scope of the compiler, the fact that it's my first change and that I have very limited go experience from a single sideproject ~2 years ago. 

Either way hopefully this should serve as a solid starting point and wanted to deliver you guys'n'gals bit of an early christmas present for all the hard work you've been putting in all this time. Have wonderful holidays 🎄

- [ ] Closes https://github.com/withastro/astro/issues/11920

### DIFF
--- a/.changeset/lucky-cloths-own.md
+++ b/.changeset/lucky-cloths-own.md
@@ -1,0 +1,10 @@
+---
+"@astrojs/compiler": patch
+---
+
+Use JSX/TSX style attribute deduplication
+
+Changed compiler to use the last occurrence of duplicate props instead of first, matching standard JSX semantics where later values override earlier ones. This ensures correct prop precedence for patterns like `<div {...props} className="override" />`.
+
+Before: <div a="1" a="2" /> → <div a="1" />
+After: <div a="1" a="2" /> → <div a="2" />
--- a/internal/parser.go
+++ b/internal/parser.go
@@ -432,7 +432,7 @@ func (p *parser) addElement() {
 		Type:          ElementNode,
 		DataAtom:      p.tok.DataAtom,
 		Data:          p.tok.Data,
-		Attr:          p.tok.Attr,
+		Attr:          deduplicateAttributes(p.tok.Attr),
 		Fragment:      isFragment(p.tok.Data),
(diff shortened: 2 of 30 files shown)


packages/compiler/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @astrojs/compiler
 
+## 2.13.1
+
+### Patch Changes
```
matthewp: also remove this change, it's done through the changeset.

samuelhulla:
@matthewp  no problem :) Done

Princesseuh:

PR #1131, opened 2026-01-15
chore: update devcontainer configuration

openscript:
## Changes

It seems like the dev container configuration has not been touched for a long time. It is not compatible with the current environment. A more modern configuration representing the same environment (Go 1.22, pnpm, node 24) like in the ci has been set.

## Testing

No tests needed for setting up dev containers.

## Docs

No documentation needed, it's for Astro compiler development only and for contributors there is no change to the previous state.


openscript:
We don't need to bump the version for this changes, hence no changeset is needed.

Princesseuh:

PR #1136, opened 2026-02-02
Fix trailing whitespace before extracted style tags

### LINKED ISSUE #1065: Slots remove whitespace where it shouldn't when `<style>` is used
### Astro Info

```block
Astro                    v5.2.5
Node                     v22.13.1
Package Manager          npm
Adapter                  none
Integrations             @astrojs/mdx
                         @astrojs/sitemap
```

_No response_

### Describe the Bug

Whitespaces are being stripped between multiple `<span>` elements when using Layout components (`<slot>`), if and only if a `<style>` is present.
This leads to inconsistent rendering between using slots and not using slots.

Weirdly, this is something that was discussed in https://github.com/withastro/astro/issues/6893, and it was mentioned that astro not removing whitespaces was a feature !

Astro should have the same rendering results whether `<slot/>`/`<style>` are being used or not.

https://stackblitz.com/edit/github-wwx73rkk?file=src%2Fpages%2FindexWithLayout.astro

### Participation

matthewp:
## Changes

- When a style tag is extracted from the document, any preceding whitespace-only text node is now also removed if it would become trailing whitespace
- Fixes #1065

## Testing

Added test case in `TestTransformTrailingSpace` to verify trailing whitespace before style tags is removed.

## Docs

Bug fix only, no docs needed.

### DIFF
--- a/.changeset/fix-trailing-whitespace.md
+++ b/.changeset/fix-trailing-whitespace.md
@@ -2,4 +2,4 @@
 "@astrojs/compiler": patch
 ---
 
-Fixes trailing whitespace being preserved before style tags when they are extracted
+Fixes a bug where trailing whitespaces were preserved before `<style>` tags after transformation, in certain cases. Now trailing whitespaces are correctly removed.


.changeset/fix-trailing-whitespace.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+Fixes trailing whitespace being preserved before style tags when they are extracted
```
ematipico:
```suggestion
Fixes a bug where trailing whitespaces were preserved before `<style>` tags after transformation, in certain cases. Now trailing whitespaces are correctly removed.
```

Different wording, it took me a while to understand the actual bug and fix

PR #1138, opened 2026-02-03
feat(spec): Minimal syntax specification

Princesseuh:
## Changes

As we look ahead, the JS ecosystem is oxidizing (or go-ing, for Microsoft), AI is arriving etc. And all those efforts are about 10 times harder when we don't know what we're doing.

In this PR, I'd like for us to make a syntax specification for the Astro format. As Astro is inherently mostly a format made up of other formats, there's not much technical details as to *how* one would make a parser for it, just the set of rules and how they interact with the specification (if there's one) of those other formats.

## Testing

N/A

## Docs

Well... It is docs


SYNTAX_SPEC.md
```diff
@@ -0,0 +1,232 @@
...
+</style>
+```
+
+---
+
+## 2. Component Script (Frontmatter)
+
+The region between the two `---` fences.
+
+- Both the opening and closing `---` must be on their own line.
```
Fryuni: This is not a requirement of the current compiler (tho maybe it should). This works:
```astro
---const greeting = "hello";---
<div>{greeting}</div>
```

natemoo-re:
@Princesseuh good points, HTML comments and slots are normal as far as parsing goes. The global `<Fragment>` and the fact that it accepts `slot` is a bit novel, though.

SYNTAX_SPEC.md
```diff
@@ -0,0 +1,343 @@
...
+---
+
+## 2. Component Script (Frontmatter)
+
+The region between the two `---` fences.
+
+- The opening and closing fences are not required on their own line. Code may appear on the same line as both fences.
+- Only one component script block is allowed per file.
+- Any amount of whitespace may appear before the opening fence or after the closing fence.
+- Any content may appear before before the opening fence, but is customarily ignored.
```
ArmandPhilippot:
```suggestion
- Any content may appear before the opening fence, but is customarily ignored.
```

SYNTAX_SPEC.md
```diff
@@ -0,0 +1,343 @@
...
+## 2. Component Script (Frontmatter)
+
+The region between the two `---` fences.
+
+- The opening and closing fences are not required on their own line. Code may appear on the same line as both fences.
+- Only one component script block is allowed per file.
+- Any amount of whitespace may appear before the opening fence or after the closing fence.
+- Any content may appear before the opening fence, but is customarily ignored.
+
+The component script is TypeScript. All standard TypeScript syntax is valid, apart from the exceptions and additions outlined in §2.1.
```
matthewp: I think this section should talk about the semantics, that it's both a module scope and a function body.

jp-knj (commented):
Just asking because I was curious 🙏🏿

SYNTAX_SPEC.md
```diff
@@ -0,0 +1,224 @@
...
+<h1>{greeting}, World!</h1>
+
+<style>
+  h1 { color: royalblue; }
+</style>
+```
+
+---
+
+## 2. Component Script (Frontmatter)
```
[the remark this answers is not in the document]
Princesseuh: Component script is most definitely the official term, frontmatter is only included for ease of reading. I should write it differently, perhaps! I'll note that in code, it is always referred as the frontmatter in Astro, the language-tools etc.

SYNTAX_SPEC.md
```diff
@@ -0,0 +1,367 @@
...
+
+The region between the two `---` fences.
+
+- The opening and closing fences are not required on their own line. Code may appear on the same line as both fences.
+- Only one component script block is allowed per file.
```
morinokami: This is a small nitpick, but to avoid confusion with the "script blocks" mentioned later, I think the word "block" in this line may be unnecessary.


```suggestion
- Only one component script is allowed per file.
```

SYNTAX_SPEC.md
```diff
@@ -0,0 +1,386 @@
...
+
+#### Attribute names
+
+Attribute names [follow the HTML conventions](https://html.spec.whatwg.org/multipage/syntax.html#syntax-attribute-name) and are not required to be valid JavaScript identifiers. For example, characters like hyphens and colons are allowed in attribute names:
+
+```astro
+<div @click="handler" x.data="value" :class="my-class" />
+```
+
+#### Namespace in component names are not supported
```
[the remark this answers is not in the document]
Princesseuh: There can only be one namespace I think!

PR #1151, opened 2026-02-28
Expand `compact` option to support JSX-style whitespace stripping

meyer:
## Changes

In this PR I’ve expanded the `compact` option to support a new string value, `"jsx"`. This allows users to enable JSX style whitespace stripping in Astro. Handy for folks like me who often pair Astro and React. No more mental shift when jumping between components… all my JSX code looks and behaves the same way. If you all are more curious about rationale, please let me know, I’ve got more to say about the topic but didn’t want to clog up the PR description.

The [JSX spec](https://facebook.github.io/jsx/) unfortunately doesn’t specify [how whitespace is dealt with in JSX](https://github.com/facebook/jsx/issues/6), but JS tooling has landed on [Babel’s whitespace handling](https://github.com/babel/babel/blob/8ddfe9c/packages/babel-types/src/utils/react/cleanJSXElementLiteralChild.ts#L5-L51) as a kind of standard.

The implementation is based on esbuild’s JSX whitespace handling, which is an exact implementation of TypeScript’s JSX whitespace handling, which in turn is an implementation of Babel’s whitespace handling.

I did this work with some serious assistance from Claude Code + Opus, but I have read each line of code and gone through the tests as well.

While working on this I (well, claude 🫣) noticed that minify tests weren’t running due to a missing `test.run()`. That has been fixed in this PR, as well as the missing `test.run()` calls from a few other test files (second commit).

## Testing

New tests added

## Docs

https://github.com/withastro/docs/pull/13326

### DIFF
--- a/packages/compiler/test/compact/minify.ts
+++ b/packages/compiler/test/compact/minify.ts
@@ -245,3 +245,5 @@ test('separated by newlines (#815)', async () => {
 
 	assert.match(result, output);
 });
+
+test.run();
--- a/.changeset/fluffy-eels-flash.md
+++ b/.changeset/fluffy-eels-flash.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": minor
+---
+
+Expand `compact` option to support JSX-style whitespace stripping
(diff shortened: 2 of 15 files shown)


packages/compiler/test/compact/minify.ts
```diff
@@ -245,3 +245,5 @@ test('separated by newlines (#815)', async () => {
 
 	assert.match(result, output);
 });
+
+test.run();
```
meyer: unintentional omissions? lmk if this is intentional and i’ll remove the commit that adds them

Princesseuh (approved):
Looks great, thank you!

PR #1156, opened 2026-03-19
feat: emit template depth tracking instructions

seroperson:
This PR adds a special handling for `<template>` tags, so they call special `templateEnter` and `templateExit` functions. It's necessary for https://github.com/withastro/astro/pull/15980 PR.

## Changes

- `print-to-js.go` now detects `template` nodes and adds `templateEnter` / `templateExit` calls inside. 
- `transform/transform.go` now fills `ContainsTemplateElement` flag to be able to dynamically include imports later.

## Testing

- Added two snap tests to confirm the behavior: simple `<template>` case and nested case

## Docs

I'm unsure whether it should be documented. I can add the documentation somewhere if needed.

### DIFF
--- a/.changeset/forty-cars-lick.md
+++ b/.changeset/forty-cars-lick.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": minor
+---
+
+Calling `templateEnter` / `templateExit` when emitting `<template>`
--- a/.changeset/forty-cars-lick.md
+++ b/.changeset/forty-cars-lick.md
@@ -1,5 +1,5 @@
 ---
-"@astrojs/compiler": minor
+"@astrojs/compiler": major
 ---
 
 Calling `templateEnter` / `templateExit` when emitting `<template>`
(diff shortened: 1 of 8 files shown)


seroperson:
@matthewp @Princesseuh Hello! Sorry for ping, but would you mind reviewing this?

seroperson:
@matthewp @Princesseuh Sorry for pinging again =) So, after Rust PR is approved, what are next the next steps? We also have this one implementation, and the PR in Astro itself ..

seroperson:
@matthewp @Princesseuh weekly ping =)

Princesseuh:
Hello, yes, sorry! We'll get to it. Just needs to be careful here because Astro uses the compiler as `^3.0.1`, so it would match a minor. If the compiler outputs a call that isn't implemented in Astro, it'd break everything etc. This is fine on the Rust counter-part, though, since it's experimental and what not.

.changeset/forty-cars-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/compiler": minor
```
Princesseuh:
```suggestion
"@astrojs/compiler": major
```

Princesseuh (approved):
Awesome work, apologies for the delay!

PR #1158, opened 2026-03-23
Fix(js_scanner): import hoisting corruption when import.meta.env …

bosmievoll:
…precedes import statements (#944)

## Changes

  - Fix position tracking in the DotToken handler sub-loop within NextImportStatement (js_scanner.go)
  - The sub-loop consumed tokens (meta, ., etc.) via l.Next() without updating the position counter i, causing subsequent import statement spans to be calculated with an offset
  - This corrupted hoisted output when import.meta.env.* appeared before import statements in Astro frontmatter
  - Add continue after the ImportToken block to prevent double-counting i += len(value) for the import keyword

## Testing

  Added 2 test cases to js_scanner_test.go:
  - import.meta.env before import — single import after import.meta.env expression
  - import.meta.env before multiple imports — two imports after bare import.meta.env reference

##  Docs

  Bug fix only — no public API changes.

### DIFF
--- a/.changeset/mighty-moments-begin.md
+++ b/.changeset/mighty-moments-begin.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+Fix (js_scanner): import hoisting corruption when import.meta.env precedes import statements (#944)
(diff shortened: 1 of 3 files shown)


Princesseuh:

PR #1162, opened 2026-05-07
fix: strip dead imports for client:only components

kimjune01:
## Changes

Strips dead imports for `client:only` components from the compiled `.astro` output. The compiler emits `import Foo from './Foo'` for `client:only` components but passes `null` to `$$renderComponent` — the import value is dead code in the prerender environment. However, Rollup still resolves the import and traverses the component's dependency tree (e.g. React, CodeMirror), inflating prerender build time for island-heavy sites.

### What this PR does

In `print-to-js.go`, before printing hoisted imports, builds a set of specifiers that appear exclusively in `ClientOnlyComponents` (not also in `HydratedComponents` or `ServerComponents`). For each hoisted import matching that set, skips it if:

1. The import has **exactly one binding** (not mixed like `import Repl, { getMeta } from './Repl'`)
2. The import is **not** a type-only import
3. The import is **not** a bare side-effect import (`import './Repl'`)

The `clientOnlyComponents` metadata in the compiled output is preserved — only the dead import statement is removed.

### Why this is safe

- The compiler already passes `null` to `$$renderComponent` for `client:only`: the import value is provably unused in the render path
- Mixed imports are preserved (guard: `len(stmt.Imports) == 1`)
- Side-effect imports are preserved (guard: `len(stmt.Imports) == 0` → not matched)
- Type imports are preserved (guard: `!stmt.IsType`)
- Dual-use components (same specifier in both `ClientOnlyComponents` and `HydratedComponents`) are preserved (specifier removed from the skip set)

### Known limitation

If frontmatter code references the imported binding for non-component purposes (e.g. `const Alias = Repl`), the import would be incorrectly removed. This is an unlikely pattern for `client:only` components (which by definition have no server-side use), but could be addressed with binding usage analysis if needed.

### Measured impact

906-page static site (Astro 6.1.2), 472 pages with 2,055 `client:only` React islands:

| Experiment | Pages | Prerender (ms) | Delta |
|---|---|---|---|
| Baseline | 906 | 4,338 | — |
| Remove reading pages entirely | 433 | 1,012 | -3,326 (-77%) |
| **Trivial reading pages (no React imports)** | **906** | **2,203** | **-2,135 (-49%)** |
| Switch to `client:only="react"` | 906 | 4,390 | +52 (noise) |

The trivial-pages experiment (same page count, no React imports) confirms ~2,135ms of prerender time is attributable to dead `client:only` import graph traversal.

### Compiled output before/after

**Before:**
```javascript
import Repl from './Repl';
$$renderComponent($$result, 'Repl', null, {"client:only":"react", ...})
```

**After:**
```javascript
$$renderComponent($$result, 'Repl', null, {"client:only":"react", ...})
```

The `clientOnlyComponents` metadata array is unchanged.

## Testing

All 6 client:only snapshot tests updated and passing. Full internal test suite passes.

Previously opened (and closed) [withastro/astro#16634](https://github.com/withastro/astro/pull/16634) with a Vite plugin approach, but moved to the compiler as the cleaner fix point.

### DIFF
--- a/.changeset/long-swans-smoke.md
+++ b/.changeset/long-swans-smoke.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+fix: strip dead imports for client:only components
--- a/internal/printer/__printer_js__/client_only_component__default_.snap
+++ b/internal/printer/__printer_js__/client_only_component__default_.snap
@@ -39,7 +39,6 @@ import {
   renderScript as $$renderScript,
   createMetadata as $$createMetadata
 } from "http://localhost:3000/";
-import Component from '../components';
 
 export const $$metadata = $$createMetadata(import.meta.url, { modules: [], hydratedComponents: [], clientOnlyComponents: ['../components'], hydrationDirectives: new Set(['only']), hoisted: [] });
 
--- a/internal/printer/__printer_js__/client_only_component__multiple_.snap
+++ b/internal/printer/__printer_js__/client_only_component__multiple_.snap
@@ -41,7 +41,6 @@ import {
   renderScript as $$renderScript,
   createMetadata as $$createMetadata
 } from "http://localhost:3000/";
-import Component from '../components';
 
 export const $$metadata = $$createMetadata(import.meta.url, { modules: [], hydratedComponents: [], clientOnlyComponents: ['../components'], hydrationDirectives: new Set(['only']), hoisted: [] });
 
--- a/internal/printer/__printer_js__/client_only_component__named_.snap
+++ b/internal/printer/__printer_js__/client_only_component__named_.snap
@@ -39,7 +39,6 @@ import {
   renderScript as $$renderScript,
   createMetadata as $$createMetadata
 } from "http://localhost:3000/";
-import { Component } from '../components';
 
 export const $$metadata = $$createMetadata(import.meta.url, { modules: [], hydratedComponents: [], clientOnlyComponents: ['../components'], hydrationDirectives: new Set(['only']), hoisted: [] });
 
--- a/internal/printer/__printer_js__/client_only_component__namespace_.snap
+++ b/internal/printer/__printer_js__/client_only_component__namespace_.snap
@@ -39,7 +39,6 @@ import {
   renderScript as $$renderScript,
   createMetadata as $$createMetadata
 } from "http://localhost:3000/";
-import * as components from '../components';
 
 export const $$metadata = $$createMetadata(import.meta.url, { modules: [], hydratedComponents: [], clientOnlyComponents: ['../components'], hydrationDirectives: new Set(['only']), hoisted: [] });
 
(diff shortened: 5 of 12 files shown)


kimjune01:
Posting the hypothesis graph for this fix to make the reasoning auditable.

## H0: the strip is correct (single-use case)

For a `client:only` component, the compiler emits:

```js
$$renderComponent($$result, 'Repl', null, {"client:only":"react", ...})
```

The component reference is `null` — the import value is provably unread in the prerender environment. Rollup still resolves the import statement and walks the component's dependency tree (React, CodeMirror, etc.), which is what inflates prerender time on island-heavy sites.

Citation: `internal/printer/print-to-js.go` already passes `null` as the third argument for `client:only`. The new code (lines added in this PR) only suppresses the printed `import` statement; it does not touch the `$$renderComponent` call or the `clientOnlyComponents` metadata array.

## H1: dual-use guard is necessary

A specifier used as `client:only` *and* anywhere else must keep its import. The guard chains four exclusions before stripping:

1. Remove from skip set if also in `HydratedComponentNodes` (any other `client:*` directive)
2. Remove if also in `ServerComponents`
3. Walk the AST and remove if any plain (no-directive) component node uses the binding
4. Word-boundary regex over the frontmatter body — catches `const Alias = Repl`, `if (Repl) {…}`, etc.

Plus the per-import gates: `len(stmt.Imports) == 1`, `!stmt.IsType`, side-effect imports skipped (zero bindings → not matched).

Verified locally with four perturbations against this branch:

| Perturbation | Expected | Result |
|---|---|---|
| `<Foo client:only />` (single use) | import stripped | stripped |
| `<Foo client:only /> <Foo />` (server-rendered too) | import preserved | preserved |
| `<Foo client:only /> <Foo client:load />` (hydrated too) | import preserved | preserved |
| `const Alias = Foo` in frontmatter | import preserved | preserved |

The corresponding snapshot tests (`client_only_preserves_*`) cover the same shapes in-tree.

## H2 (devil's advocate): is there a hidden consumer of the import?

Considered:

- **HMR**: Astro dev HMR is driven by Vite watching the `.astro` source file, not the printed JS import graph. Removing a dead `import` from the compiled output doesn't change which source files are watched.
- **Source maps**: `printTextWithSourcemap` maps printed JS back to original source positions. Suppressing a printed import means there's nothing to map for that line; it doesn't break maps for surrounding code.
- **Runtime introspection**: the `clientOnlyComponents` metadata array (consumed by the runtime to mount the island) is preserved unchanged. Only the `import` statement is suppressed.

Couldn't find a codepath where the printed binding name is read by anything other than Rollup's resolver. Open to counterexamples.

## What this PR doesn't do

- Doesn't change the runtime renderer or the metadata emitted for islands
- Doesn't touch HMR, source maps, or the scanner
- Doesn't strip mixed bindings (`import Foo, { helper } from './Foo'`) — the whole import is preserved
- Doesn't strip type-only or side-effect imports
- Doesn't do binding usage analysis beyond a word-boundary scan of the frontmatter body (noted as a known limitation in the PR description)

Happy to extend the guard, add more snapshot cases, or address review if anyone has an additional shape in mind.

PR #1164, opened 2026-06-05
escape backslashes in JS string literal output

digi-scrypt:
1. escapeDoubleQuote and escapeSingleQuote escape the closing quote but not the backslash, so a value ending in a backslash (or one before a quote) lands a `\` right against the closing quote the printer adds and breaks out of the JS string literal.
2. these feed the renderComponent props object, slot names and the hoisted script `src` metadata, all built from attribute values in the source.

A quoted attribute is the trigger. `<Component path='C:\dir\' />` currently emits `{"path":"C:\dir\"}` which node refuses to parse; a crafted trailing backslash terminates the string early and the following bytes are no longer inside it. Same story for `src` on a hoisted `<script>`.

Escaped the backslash first in both helpers. The one place escapeNewlines wraps the value I flipped the order so the `\n` it inserts isn't doubled. Added a snapshot covering a backslash attribute.

### DIFF
--- a/internal/printer/__printer_js__/component_attribute_with_backslash.snap
+++ b/internal/printer/__printer_js__/component_attribute_with_backslash.snap
@@ -1,0 +1,41 @@
+
+[TestPrinter/component_attribute_with_backslash - 1]
+## Input
+
+```
+<Component path='C:\dir\' name={x} />
+```
+
+## Output
+
+```js
+import {
+  Fragment,
+  render as $$render,
+  createAstro as $$createAstro,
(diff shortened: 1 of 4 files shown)


ematipico:
> which node refuses to parse;

Can you file a bug for this? Or can you be more precise about the operation? Because your `JSON.parse()` for the emitted about is a runtime error

ematipico:
Closing. New bugs should be reported to https://github.com/withastro/compiler-rs

PR #1165, opened 2026-06-15
Fix crash on Props generic containing a shift or comparison operator

### LINKED ISSUE #78: Transformer can’t reliably parse your generic polymorphic props type
### Astro Info

```block
Astro                    v6.1.3
Node                     v24.14.1
Package Manager          pnpm
Output                   server
Adapter                  @astrojs/node
Integrations             none
```

Firefox

### Describe the Bug

In our standard code rules, we use to interface strictly for components.
We are creating some polymorphic components and defining them as follows:
```ts
interface Props<Tag extends HTMLTag = "button" | "a"> extends Polymorphic<
  { as: Tag } & ButtonVariants
> {
  disabled?: boolean;
  fullWidth?: boolean;
  ariaLabel?: string;
}

const {
  as: Tag = "button",
  variant = "solid",
  size = "md",
  color = "primary",
  fullWidth = false,
  disabled = false,
  ariaLabel,
  href,
  class: className,
  ...rest
} = Astro.props as Props<typeof Tag>;

const classes = cn(
  buttonVariants({
    variant,
    size,
    color,
    fullWidth,
  }),
  disabled && "pointer-events-none",
  className,
  Tag === "button" && "cursor-pointer",
);
---

<Tag
  {...rest}
  href={Tag === "a" ? (disabled ? undefined : href) : undefined}
  disabled={Tag === "button" ? disabled || undefined : undefined}
  aria-disabled={Tag === "a" ? disabled || undefined : undefined}
  aria-label={ariaLabel}
  class={classes}
>
  <slot />
</Tag>
```

When we save this component, the `Astro` language server crashes with the following error:
```sh
There was an error transforming /home/LuisAlaguna/Projects/central-mosque-bogota/frontend/src/components/actions/Button/Button.astro to TSX. An empty file will be returned instead. Please create an issue: https://github.com/withastro/astro/issues
Error: TypeError: Cannot read properties of undefined (reading 'map').
panic: runtime error: slice bounds out of range [:15370] with capacity 1792

goroutine 6 [running]:
github.com/withastro/compiler/internal/js_scanner.GetPropsType({0x1622a00, 0x6a4, 0x700})
        github.com/withastro/compiler/internal/js_scanner/js_scanner.go:430 +0x11f
github.com/withastro/compiler/internal/printer.renderTsx(0x1413838, 0x150d760, 0x1413ab0)
        github.com/withastro/compiler/internal/printer/print-to-tsx.go:336 +0xc7
github.com/withastro/compiler/internal/printer.PrintToTSX({0x1622300, 0x6a4}, 0x150d760, {0x0, 0x0}, {{0x3c714, 0x6}, {0x17ea850, 0x6c}, {0x17ea850, ...}, ...}, ...)
        github.com/withastro/compiler/internal/printer/print-to-tsx.go:35 +0x12
main.main.ConvertToTSX.func3({{}, 0x7ff8000100000089, 0x17f5698}, {0x1434260, 0x2, 0x2})
        ./astro-wasm.go:293 +0xf
syscall/js.handleEvent()
        syscall/js/func.go:100 +0x23
exit code: 2
```

Then, the extension does not work.

It should be able to transform the component even using polymorphic generic types and interfaces in components.

We are following this documentation: https://docs.astro.build/en/guides/typescript/#polymorphic-type

https://github.com/SalahAdDin/central-mosque-bogota/tree/main/frontend

### Participation

sarathfrancis90:
While converting a component to TSX, `GetPropsType` panics with a slice-bounds error on valid TypeScript like:

```ts
interface Props<T extends number = (1 << 2)> {
  count: T;
}
```

The scanner walks the generic parameter list tracking angle-bracket depth, but it treated *any* token containing `<` or `>` as a generic bracket. The lexer emits operators such as `<<`, `<=` and `>>=` as a single token, so each one bumped the bracket depth and the byte offset the wrong way. The offset then ran past the end of the source and `source[start:end]` panicked, which surfaces as "There was an error transforming … to TSX. An empty file will be returned" in the language server.

The fix narrows the bracket handling: only a bare `<` opens a generic, and only a run of `>` (the `>>` / `>>>` tokens used to close nested generics like `Foo<Bar<T>>`) closes one. Real operators are left untouched, so the depth and offset stay in sync.

Tested by adding cases to `TestGetPropsType` for a left-shift default, a comparison default and a nested-generic close; the first two panic on `main` and pass with the fix, and the nested-generic case guards against regressing the `>>` handling. `go test ./internal/...`, `go vet` and `gofmt` are clean, and the wasm build succeeds.

Fixes withastro/compiler-rs#78

### DIFF
--- a/.changeset/fix-generic-props-operator-crash.md
+++ b/.changeset/fix-generic-props-operator-crash.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+Fix a crash when converting a component to TSX whose `Props` generic contains an operator that the lexer emits as a single `<`/`>` token, such as `interface Props<T extends number = (1 << 2)>`. These operators were mistaken for generic angle brackets, desyncing the bracket-depth tracking and panicking with a slice bounds error.
--- a/internal/js_scanner/js_scanner.go
+++ b/internal/js_scanner/js_scanner.go
@@ -479,6 +479,22 @@ type Props struct {
 	Generics  string
 }
 
+// isClosingAngleRun reports whether value is a non-empty run of `>` characters,
+// i.e. one of the `>`, `>>` or `>>>` tokens the lexer emits when several
+// generics close at once (`Foo<Bar<T>>`). It deliberately excludes operators
+// like `>=` and `>>=` that merely contain a `>`.
+func isClosingAngleRun(value []byte) bool {
+	if len(value) == 0 {
+		return false
+	}
+	for _, c := range value {
+		if c != '>' {
+			return false
+		}
+	}
+	return true
+}
+
 func GetPropsType(source []byte) Props {
 	defaultPropType := "Record<string, any>"
 	ident := defaultPropType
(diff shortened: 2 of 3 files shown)


Princesseuh:

sarathfrancis90:
Thanks for letting me know, and no problem. I'll check whether the panic still reproduces on compiler-rs and re-open there if so.

PR #1167, opened 2026-06-15
fix: escape quoted transition attribute value in convertAttributeValue

rootvector2:
## Changes

- root cause: `convertAttributeValue` wraps a quoted `transition:name` / `transition:animate` value in a JS string literal but never escapes it, unlike every other quoted-value path in the printer which goes through `escapeDoubleQuote`.
- a single-quoted `.astro` attribute value can legally contain a `"`, so `transition:name='x"+(expr)+"y'` breaks out of the string in the emitted `$$renderTransition(...)` call and injects arbitrary js into the generated module.
- escape the value with `escapeDoubleQuote(escapeNewlines(...))`, matching `printAttributesToObject` and the slot paths.

## Testing

- added a snapshot case `transition:name with a quote in the value`; it emits the value as `"one\"two"`, and reverting the one-line change fails it. found while auditing the printer's quoted-value emission sites.

## Docs

- bug fix only.

### DIFF
--- a/.changeset/escape-transition-attr-value.md
+++ b/.changeset/escape-transition-attr-value.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/compiler": patch
+---
+
+Escape quoted `transition:name` and `transition:animate` attribute values when emitting the `renderTransition` call, so a value containing a double quote no longer breaks out of the generated JS string literal.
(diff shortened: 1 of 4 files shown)


Princesseuh:

